PR #131, opened 2015-01-30
Proposal for pagination strategy to be applied to endpoints and provider...

vibe13:
...s

### DIFF
--- a/pnc-rest/src/main/java/org/jboss/pnc/rest/provider/ProductProvider.java
+++ b/pnc-rest/src/main/java/org/jboss/pnc/rest/provider/ProductProvider.java
@@ -24,20 +24,21 @@ public class ProductProvider extends BasePaginationProvider<ProductRest, Product
     public ProductProvider() {
     }
 
+    // Needed to map the Entity into the proper REST object
     @Override
-    public Function<? super Product, ? extends ProductRest> mapper() {
+    public Function<? super Product, ? extends ProductRest> toRestModel() {
         return product -> new ProductRest(product);
     }
 
     @Override
     public String getDefaultSortingField() {
-        return "name";
+        return Product.DEFAULT_SORTING_FIELD;
     }
 
     public Object getAll(Integer pageIndex, Integer pageSize, String field, String sorting) {
 
         if (noPaginationRequired(pageIndex, pageSize, field, sorting)) {
-            return productRepository.findAll().stream().map(mapper()).collect(Collectors.toList());
+            return productRepository.findAll().stream().map(toRestModel()).collect(Collectors.toList());
         } else {
             return transform(productRepository.findAll(buildPageRequest(pageIndex, pageSize, field, sorting)));
         }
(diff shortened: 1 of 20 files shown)


vibe13:
@slaskawi @alexcreasy 
With this proposal, you would get i.e. all the products still using 
http://[host]:[port]/pnc-web/rest/product
while you could paginate and sort using i.e.
http://[host]:[port]/pnc-web/rest/product?pageIndex=1&pageSize=5&sorted_by=name&sorting=desc, getting a Pagination wrapper object with all the metadata of pagination taken from the Pageable object from Spring Data. These metadata in the wrapper object could be consumed on the client for the pagination behaviour.

Waiting for feedback about the implementation.

pnc-rest/src/main/java/org/jboss/pnc/rest/provider/ProductProvider.java
```diff
@@ -24,8 +24,23 @@ public ProductProvider(ProductRepository productRepository) {
     public ProductProvider() {
     }
 
-    public List<ProductRest> getAll() {
-        return productRepository.findAll().stream().map(product -> new ProductRest(product)).collect(Collectors.toList());
+    @Override
+    public Function<? super Product, ? extends ProductRest> mapper() {
```
slaskawi: Maybe toRestModel would be a slightly better name

slaskawi:
Looks great :)

pnc-rest/src/main/java/org/jboss/pnc/rest/provider/ProductProvider.java
```diff
@@ -24,8 +24,23 @@ public ProductProvider(ProductRepository productRepository) {
...
+    public String getDefaultSortingField() {
+        return "name";
+    }
+
+    public Object getAll(Integer pageIndex, Integer pageSize, String field, String sorting) {
+
+        if (noPaginationRequired(pageIndex, pageSize, field, sorting)) {
+            return productRepository.findAll().stream().map(mapper()).collect(Collectors.toList());
+        } else {
+            return transform(productRepository.findAll(buildPageRequest(pageIndex, pageSize, field, sorting)));
```
slaskawi: Hmmm on the second thought, I see we are extracting everything from DB and apply pagination on those results. I think we should modify Datastore layer to accept `Pageable` objects. Just like it is done here: <http://docs.spring.io/spring-data/commons/docs/current/reference/html/#repositories.special-parameters>

pnc-rest/src/main/java/org/jboss/pnc/rest/provider/ProductProvider.java
vibe13: :+1:

pnc-rest/src/main/java/org/jboss/pnc/rest/provider/ProductProvider.java
vibe13: ok as clarified on IRC, a PageRequest is already fetched into the findAll() method, so Hibernate is doing the pagination work for us already.

PR #133, opened 2015-02-03
Store repository manager results to datastore.

matejonnet:


jdcasey:
It seems like it might make more sense to contain the logic for extracting artifacts within the driver itself, and pass it back to the result (via repository configuration, right?) using a Supplier. That way, at least all the logic is contained in a single class instead of being spread over classes that are drivers and classes that are meant to hold configuration/session tokens?

jdcasey:
retest this please

PR #160, opened 2015-02-12
[NCL-444] Add abbreviation, prodCode, and pgmSysName to product

pgier:


slaskawi:
OS1/Private cloud is restarting, so we won't have CI server today...
Checking manually

slaskawi:
Integrated thanks!

PR #172, opened 2015-02-20
[NCL-434] Remove product_version_project and add build_record_set

pgier:
This changes the db model and rest interface to remove the existing
product_version_project entity and add the new entity build_record_set

The UI portion of the changes will be done in NCL-435


pnc-rest/src/main/java/org/jboss/pnc/rest/endpoint/ProjectEndpoint.java
```diff
@@ -75,12 +75,12 @@ public Response update(@ApiParam(value = "Project id", required = true) @PathPar
         return Response.ok().build();
     }
 
-    @ApiOperation(value = "Gets all Projects of a Product Version")
+    /*@ApiOperation(value = "Gets all Projects of a Product Version")
```
vibe13: This will break actual UI

vibe13:
@pgier @slaskawi @alexcreasy This story was moved to Sprint 16. Unless we mimic the removed endpoint above with the new model (ProductVersion->BuildConfigurationSet->BuildConfiguration->Project), the UI will be broken for Sprint 15, and we may have to revert this merge and reapply in in Sprint 16. wdyt?

slaskawi:
+1 for revert

alexcreasy:
also +1 for revert and re-merge when we have all aspects of this story marked for development.

rnc:
+1 to postpone

vibe13:
@alexcreasy @slaskawi @rnc @pgier opened https://github.com/project-ncl/pnc/pull/178

pnc-rest/src/main/java/org/jboss/pnc/rest/endpoint/ProjectEndpoint.java
pgier: Yes, it purposely breaks that part of the UI because the direct relationship between product versions and projects doesn't make sense, IMO, and should be removed.
The UI will be updated in NCL-480

PR #177, opened 2015-02-25
Docker env service

jbartece:
Docker environment driver implementation
Definition of SPI for EnvironmentService


vibe13:
retest this please

docker-environment-driver/src/test/java/org/jboss/pnc/environment/docker/DockerEnvironmentDriverRemoteTest.java
```diff
@@ -0,0 +1,278 @@
...
+
+    private static final Logger log = Logger.getLogger(DockerEnvironmentDriverRemoteTest.class.getName());
+
+    @Inject
+    private DockerEnvironmentDriver dockerEnvDriver;
+
+    @Inject
+    private Generator generator;
+
+    private final String DOCKER_IP = "10.3.8.102";
```
matejonnet: Use configuration properties instead of fixed addresses.

docker-environment-driver/src/main/java/org/jboss/pnc/environment/docker/ConfigurationBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+     * 
+     * @param systemPropertyName Name of system property with filename
+     * @param fileName Name of default template file name
+     * @return Loaded template
+     * @throws EnvironmentDriverException Thrown if template cannot be loaded
+     */
+    private String loadConfigTemplate(String systemPropertyName, String fileName)
+            throws EnvironmentDriverException {
+        try {
+            return IoUtils.readFileOrResource(systemPropertyName, fileName,
```
slaskawi: Shouldn't we use context class loader here? It may become tricky when deploying on app server.

docker-environment-driver/src/main/java/org/jboss/pnc/environment/docker/DockerEnvironmentDriver.java
```diff
@@ -1,17 +1,297 @@
...
+                    configBuilder.createMavenConfig(dependencyUrl, deployUrl), null);
+        } catch (Exception e) {
+            // Creating container failed => clean up
+            if (buildContainerState != BuildContainerState.NOT_BUILT) {
+                if (buildContainerState == BuildContainerState.BUILT)
+                    destroyContainer(containerId, false);
+                else
+                    destroyContainer(containerId, true);
+            }
+            e.printStackTrace();
```
slaskawi: To be removed

docker-environment-driver/src/main/java/org/jboss/pnc/environment/docker/DockerEnvironmentDriver.java
```diff
@@ -1,17 +1,297 @@
...
+     * @param pathOnHost Path in target container, where the data are passed
+     * @param stringData Data, which will be transfered to the target container (may be null if streamData are set)
+     * @param streamData Data, which will be transfered to the target container (may be null if stringData are set)
+     * @throws EnvironmentDriverException Thrown if both stringData and streamData are null
+     */
+    public void copyFileToContainer(int sshPort, String pathOnHost, String stringData,
+            InputStream streamData) throws EnvironmentDriverException {
+        SshClient sshClient = new SshjSshClient(new BackoffLimitedRetryHandler() {
+        }, // TODO check retryHandler configuration
+                HostAndPort.fromParts("10.3.8.102", sshPort),
```
slaskawi: This address should be configurable (ideally via environment variables)

docker-environment-driver/src/main/java/org/jboss/pnc/environment/docker/DockerEnvironmentDriver.java
```diff
@@ -1,17 +1,297 @@
...
+ *
+ */
+@ApplicationScoped
 public class DockerEnvironmentDriver implements EnvironmentDriver {
 
-    private static final Logger logger = Logger.getLogger(MethodHandles.lookup().lookupClass().getName());
+    private static final Logger logger =
+            Logger.getLogger(MethodHandles.lookup().lookupClass().getName());
+
+    /** User in running environment to which we can connect with SSH */
```
slaskawi: Maybe I'm picky - but why do we need comments on private fields? Those are implementation details...

jbartece:
It is not final version, I'm just working on configuration via configuration files

docker-environment-driver/src/main/java/org/jboss/pnc/environment/docker/ConfigurationBuilder.java
jbartece: I took an inspiration in JenkinsBuildDriver, but I can change it if you think it is needed

docker-environment-driver/src/main/java/org/jboss/pnc/environment/docker/DockerEnvironmentDriver.java
jbartece: Forgot to remove it :-)

docker-environment-driver/src/main/java/org/jboss/pnc/environment/docker/DockerEnvironmentDriver.java
jbartece: I'm just working on it

pom.xml
```diff
@@ -362,6 +362,12 @@
         <artifactId>jclouds-slf4j</artifactId>
         <version>${version.jclouds}</version>
       </dependency>
+      
+      <dependency>
+        <groupId>org.jboss.resteasy</groupId>
+        <artifactId>resteasy-jaxrs</artifactId>
```
slaskawi: Do we need an implementation here? Can't we use JAX-RS api?

docker-environment-driver/src/main/java/org/jboss/pnc/environment/docker/DockerEnvironmentDriver.java
jbartece: I think that comments on private fields are not required, but it can be sometimes easily readable for other people, who read the code

pom.xml
jbartece: There is a bug in JClouds and I cannot get needed data from Docker. So I have to get the data from Docker REST. T tried to use "jaxrs-api", but there is no "ClientRequest" class. So I used implementation

docker-environment-driver/src/main/java/org/jboss/pnc/environment/docker/ConfigurationBuilder.java
slaskawi: Yes, please. In general, when implementing things against JEE spec, we should always use Context CL.

docker-environment-driver/src/main/java/org/jboss/pnc/environment/docker/DockerEnvironmentDriver.java
slaskawi: Yeah, this is just my preference... Instead of using comments I would prefer more expressive variable names (even though they are much longer in that case). Comments also tend to survive refactorings and after some time they don't reflect the reality.

pom.xml
slaskawi: Ok, understood. Using a JAX-RS 2.0 Client is a perfect use case when you need an implementation. Thanks for clarification.

jbartece:
retest this please

jbartece:
retest this please

jbartece:
retest this please

docker-environment-driver/pom.xml
```diff
@@ -1,85 +1,249 @@
...
+		<java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+		<jboss.version>8.2.0.Final</jboss.version>
+		<app.server>wildfly</app.server>
+		<test.server.unpack.dir>${project.build.directory}</test.server.unpack.dir>
+		<test.server.build.dir>${test.server.unpack.dir}/${app.server}-${jboss.version}</test.server.build.dir>
+	</properties>
+
+	<repositories>
+		<repository>
+			<id>jboss-public-repository-group</id>
```
slaskawi: Do we need this repo? If so, perhaps it would be better to add it in parent pom?

docker-environment-driver/pom.xml
```diff
@@ -1,85 +1,249 @@
...
+				<groupId>org.wildfly</groupId>
+				<artifactId>wildfly-arquillian-container-embedded</artifactId>
+				<version>${jboss.version}</version>
+				<scope>test</scope>
+			</dependency>
+			<dependency>
+				<groupId>org.wildfly</groupId>
+				<artifactId>wildfly-embedded</artifactId>
+				<version>${jboss.version}</version>
+				<scope>test</scope>
```
slaskawi: Shouldn't this also go to parent pom?

docker-environment-driver/pom.xml
```diff
@@ -1,85 +1,249 @@
...
+			</dependency>
+			<dependency>
+				<groupId>org.jboss.shrinkwrap.resolver</groupId>
+				<artifactId>shrinkwrap-resolver-bom</artifactId>
```
slaskawi: I think in general resolver bom should be above Arquillian bom (this is one of the criteria to set dependency versions).

pnc-core/src/test/java/org/jboss/pnc/core/test/mock/EnvironmentDriverMock.java
```diff
@@ -1,19 +1,31 @@
...
 @ApplicationScoped
 public class EnvironmentDriverMock implements EnvironmentDriver {
 
     private static final Logger logger = Logger.getLogger(MethodHandles.lookup().lookupClass().getName());
 
     @Override
-    public void buildEnvironment(Environment buildEnvironment) {
+    public RunningEnvironment buildEnvironment(Environment environment, String dependencyUrl, String deployUrl)
+            throws EnvironmentDriverException {
+        // TODO Auto-generated method stub
```
slaskawi: //TODO to be removed :D

pnc-core/src/test/java/org/jboss/pnc/core/test/mock/EnvironmentDriverMock.java
```diff
@@ -1,19 +1,31 @@
...
+        return null;
+    }
 
+    @Override
+    public boolean canBuildEnvironment(Environment environment) {
+        // TODO Auto-generated method stub
```
slaskawi: The same here

docker-environment-driver/pom.xml
jbartece: I left it there by mistake. Removing it

docker-environment-driver/pom.xml
jbartece: OK

common/src/main/java/org/jboss/pnc/common/json/moduleconfig/DockerEnvironmentDriverModuleConfig.java
```diff
@@ -0,0 +1,62 @@
...
+        return inContainerUserPassword;
+    }
+
+    public void setInContainerUserPassword(String inContainerUserPassword) {
+        this.inContainerUserPassword = inContainerUserPassword;
+    }
+
+    @Override
+    public String toString() {
+        return "DockerEnvironmentDriverModuleConfig ["
```
slaskawi: Since we have Guava in classpath - you may consider using ToStringHelper: http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Objects.ToStringHelper.html

docker-environment-driver/pom.xml
jbartece: OK

docker-environment-driver/pom.xml
```diff
@@ -1,85 +1,249 @@
...
+		<!-- Remote dependencies -->
+		<dependency>
+			<groupId>org.jboss.spec</groupId>
+			<artifactId>jboss-javaee-6.0</artifactId>
+			<type>pom</type>
+			<scope>provided</scope>
+		</dependency>
+
+		<dependency>
+			<groupId>javax.enterprise</groupId>
```
slaskawi: Since we import JEE6 api - do we need those APIs? They all should be included in JEE6 spec.

pnc-core/src/test/java/org/jboss/pnc/core/test/mock/EnvironmentDriverMock.java
jbartece: That are mocks, which was there a longer time ago. I didn't want to make this patch too big. I'll remove it after integrating this driver with JenkinsBuildDriver

docker-environment-driver/pom.xml
```diff
@@ -1,85 +1,249 @@
...
+									<overWrite>true</overWrite>
+									<outputDirectory>${test.server.unpack.dir}</outputDirectory>
+								</artifactItem>
+							</artifactItems>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
+			<plugin>
+				<groupId>org.wildfly.plugins</groupId>
```
slaskawi: Just to be clear here - with this POM all Arquillian tests will run on Wildfly only. We use some more complicated setup in integration-test module and CI runs there twice - with WF and EAP. I would like to confirm that this is what you want to do - run it only on WF (or maybe another approach - test as much as you can using unit tests and perform full integration tests in integration-test module).

docker-environment-driver/src/test/java/org/jboss/pnc/environment/docker/DockerEnvironmentDriverRemoteTest.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /**
+     * Checks if container was started and the services are on.
+     * 
+     * @param runningEnv Connection data about environment to test
+     * @param shouldBeRunning Indicates, if the environment should be running or should be not available
+     * @param baseErrorMsg Prefix of error message, which is printed if the container is not in expected state
+     */
+    private void testRunningContainer(DockerRunningEnvironment runningEnv, boolean shouldBeRunning,
```
slaskawi: Maybe using custom assertions would be a better idea (maybe not now, but in the future?): they look much nicer: http://joel-costigliola.github.io/assertj/assertj-core-custom-assertions.html

common/src/main/java/org/jboss/pnc/common/json/moduleconfig/DockerEnvironmentDriverModuleConfig.java
jbartece: I generated it using IDE. What is the advantage of ToStringHelper?

slaskawi:
Great job! I have fond some really small things. Please take a look at my comments in pom file - those things might be quite important.

docker-environment-driver/src/test/java/org/jboss/pnc/environment/docker/DockerEnvironmentDriverRemoteTest.java
jbartece: I'll look at later

docker-environment-driver/pom.xml
jbartece: I know it is tested on both EAP and Wildfly, but I think it is not needed for Unit tests

common/src/main/java/org/jboss/pnc/common/json/moduleconfig/DockerEnvironmentDriverModuleConfig.java
slaskawi: It's shorter, but generating from IDE is also OK.

pnc-core/src/test/java/org/jboss/pnc/core/test/mock/EnvironmentDriverMock.java
slaskawi: Ok, understood

PR #184, opened 2015-03-04
Network isolation for docker with jenkins/maven

mareknovotny:
https://projects.engineering.redhat.com/browse/NCL-528 and partially https://projects.engineering.redhat.com/browse/NCL-529 for maven preparation

### DIFF
--- a/docker-environment-driver/src/main/resources/dockerConfiguration/Dockerfile
+++ b/docker-environment-driver/src/main/resources/dockerConfiguration/Dockerfile
@@ -3,6 +3,7 @@ MAINTAINER Jakub Bartecek <jbartece@redhat.com>
 USER root
 
 #Install SSHD
+# why? see http://blog.docker.com/2014/06/why-you-dont-need-to-run-sshd-in-docker/
 RUN apt-get update && apt-get install -y openssh-server
 RUN mkdir /var/run/sshd
 
@@ -19,12 +20,22 @@ RUN echo "export VISIBLE=now" >> /etc/profile
 #RUN chown jenkins /usr/sbin/sshd
 
 #Install maven
-ADD http://mirror.hosting90.cz/apache/maven/maven-3/3.2.5/binaries/apache-maven-3.2.5-bin.tar.gz /root/maven.tar.gz
-RUN mkdir /usr/apache-maven/
-RUN tar -xzvf /root/maven.tar.gz --directory=/usr/apache-maven
-RUN echo "export M2_HOME=/usr/apache-maven/apache-maven-3.2.5" >> /etc/profile
-RUN echo "export PATH=$PATH:/usr/apache-maven/apache-maven-3.2.5/bin" >> /etc/profile
+# set installed Maven version you can easily change it later
+ENV MAVEN_VERSION 3.2.5
+RUN curl -sSL http://archive.apache.org/dist/maven/maven-3/$MAVEN_VERSION/binaries/apache-maven-$MAVEN_VERSION-bin.tar.gz | tar xzf - -C /usr/share \
+&& mv /usr/share/apache-maven-$MAVEN_VERSION /usr/share/maven \
+&& ln -s /usr/share/maven/bin/mvn /usr/bin/mvn
+ 
+RUN echo "export M2_HOME=/usr/share/maven" >> /etc/profile
+# isn't better to set it like? ^^^
+# ENV M2_HOME /usr/share/maven 
 RUN mkdir /root/.m2
+
+# Could be used for images based on this Docker image to specify default mirrored repository if different from Maven Central below
+ENV mavenRepositoryUrl http://repo.maven.apache.org/maven2
+ADD settings.xml /usr/share/maven/conf/settings.xml
+
+ADD isolate-with-iptables.sh /tmp/isolate-with-iptables.sh
 
 #Prepare script to start Jenkins and SSH daemon
 RUN echo "/usr/sbin/sshd -D & \n /usr/local/bin/jenkins.sh " > /usr/bin/startPrograms.sh
--- a/docker-environment-driver/src/main/resources/isolatedDockerConfiguration/Dockerfile
+++ b/docker-environment-driver/src/main/resources/isolatedDockerConfiguration/Dockerfile
@@ -1,0 +1,80 @@
+FROM fedora
+MAINTAINER Marek Novotny <mnovotny@redhat.com>
+
+# install java and zip
+RUN yum -y update && yum clean all
+RUN yum -y install tar gzip java java-devel && yum clean all
+
+# set JAVA_HOME
+ENV JAVA_HOME /usr/lib/jvm/java-openjdk
+
+## jenkins instalation
+ENV JENKINS_HOME /var/jenkins_home
+
+# Jenkins is ran with user `jenkins`, uid = 1000
+# If you bind mount a volume from host/vloume from a data container,
+# ensure you use same uid
+RUN useradd -d "$JENKINS_HOME" -u 1000 -m -s /bin/bash jenkins
+# Jenkins home directoy is a volume, so configuration and build history
+# can be persisted and survive image upgrades
+VOLUME /var/jenkins_home
+# `/usr/share/jenkins/ref/` contains all reference configuration we want
+# to set on a fresh new installation. Use it to bundle additional plugins
+# or config file with your custom jenkins Docker image.
+RUN mkdir -p /usr/share/jenkins/ref/init.groovy.d
+COPY init.groovy /usr/share/jenkins/ref/init.groovy.d/tcp-slave-angent-port.groovy
+ENV JENKINS_VERSION 1.554.1
+# could use ADD but this one does not check Last-Modified header
+# see https://github.com/docker/docker/issues/8331
+RUN curl -L http://mirrors.jenkins-ci.org/war-stable/$JENKINS_VERSION/jenkins.war -o /usr/share/jenkins/jenkins.war
+ENV JENKINS_UC https://updates.jenkins-ci.org
+RUN chown -R jenkins "$JENKINS_HOME" /usr/share/jenkins/ref
+# for main web interface:
+EXPOSE 8080
+# will be used by attached slave agents:
+EXPOSE 50000
+USER jenkins
+COPY jenkins.sh /usr/local/bin/jenkins.sh
+ENTRYPOINT ["/usr/local/bin/jenkins.sh"]
+# from a derived Dockerfile, can use `RUN plugin.sh active.txt` to setup /usr/share/jenkins/ref/plugins from a support bundle
+
+## end of jenkins installation
+
+USER root
+#Install SSHD
+# why? see http://blog.docker.com/2014/06/why-you-dont-need-to-run-sshd-in-docker/
+RUN yum -y install openssh-server  && yum clean all
+RUN mkdir /var/run/sshd
+
+#!!!! CHANGE ROOT PASSWORD BEFORE REAL USE !!!!
+RUN echo 'root:changeme' | chpasswd
+RUN sed -i 's/PermitRootLogin without-password/PermitRootLogin yes/' /etc/ssh/sshd_config
+
+# SSH login fix. Otherwise user is kicked off after login
+RUN sed 's@session\s*required\s*pam_loginuid.so@session optional pam_loginuid.so@g' -i /etc/pam.d/sshd
+ENV NOTVISIBLE "in users profile"
+RUN echo "export VISIBLE=now" >> /etc/profile
+
+#Install maven
+# set installed Maven version you can easily change it later
+ENV MAVEN_VERSION 3.2.5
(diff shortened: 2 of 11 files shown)


slaskawi:
test this please

jbartece:
@mareknovotny Rebase to upstream/master. It could help

docker-environment-driver/src/main/resources/dockerConfiguration/Dockerfile
```diff
@@ -19,13 +20,23 @@ RUN echo "export VISIBLE=now" >> /etc/profile
...
-RUN tar -xzvf /root/maven.tar.gz --directory=/usr/apache-maven
-RUN echo "export M2_HOME=/usr/apache-maven/apache-maven-3.2.5" >> /etc/profile
-RUN echo "export PATH=$PATH:/usr/apache-maven/apache-maven-3.2.5/bin" >> /etc/profile
+# set installed Maven version you can easily change it later
+ENV MAVEN_VERSION 3.2.5
+RUN curl -sSL http://archive.apache.org/dist/maven/maven-3/$MAVEN_VERSION/binaries/apache-maven-$MAVEN_VERSION-bin.tar.gz | tar xzf - -C /usr/share \
+&& mv /usr/share/apache-maven-$MAVEN_VERSION /usr/share/maven \
+&& ln -s /usr/share/maven/bin/mvn /usr/bin/mvn
+ 
+RUN echo "export M2_HOME=/usr/share/maven" >> /etc/profile
```
slaskawi: Just wondering - what's the argument for keeping all artifact in custom directory?

docker-environment-driver/src/main/resources/dockerConfiguration/Dockerfile
```diff
@@ -19,13 +20,23 @@ RUN echo "export VISIBLE=now" >> /etc/profile
...
+# isn't better to set it like? ^^^
+# ENV M2_HOME /usr/share/maven 
 RUN mkdir /root/.m2
 
+# Could be used for images based on this Docker image to specify default mirrored repository if different from Maven Central below
+ENV mavenRepositoryUrl http://repo.maven.apache.org/maven2
+ADD settings.xml /usr/share/maven/conf/settings.xml
```
slaskawi: Why are we replacing global config instead of using local one (`~/.m2/settings.xml`)?

docker-environment-driver/src/main/resources/isolatedDockerConfiguration/Dockerfile
```diff
@@ -0,0 +1,80 @@
+FROM fedora
+MAINTAINER Marek Novotny <mnovotny@redhat.com>
+
+# install java and zip
+RUN yum -y update && yum clean all
+RUN yum -y install tar gzip java java-devel && yum clean all
```
slaskawi: Shouldn't we use some specific java version here?

docker-environment-driver/src/main/resources/dockerConfiguration/Dockerfile
mareknovotny: because later we want to limit by setting mirror to access only for our proxy server or our Maven repository (nexus or whatever else)

docker-environment-driver/src/main/resources/isolatedDockerConfiguration/Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+RUN mkdir -p /usr/share/jenkins/ref/init.groovy.d
+COPY init.groovy /usr/share/jenkins/ref/init.groovy.d/tcp-slave-angent-port.groovy
+ENV JENKINS_VERSION 1.554.1
+# could use ADD but this one does not check Last-Modified header
+# see https://github.com/docker/docker/issues/8331
+RUN curl -L http://mirrors.jenkins-ci.org/war-stable/$JENKINS_VERSION/jenkins.war -o /usr/share/jenkins/jenkins.war
+ENV JENKINS_UC https://updates.jenkins-ci.org
+RUN chown -R jenkins "$JENKINS_HOME" /usr/share/jenkins/ref
+# for main web interface:
+EXPOSE 8080
```
slaskawi: expose might be squashed into 1 liner

docker-environment-driver/src/main/resources/isolatedDockerConfiguration/Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+
+# install java and zip
+RUN yum -y update && yum clean all
+RUN yum -y install tar gzip java java-devel && yum clean all
+
+# set JAVA_HOME
+ENV JAVA_HOME /usr/lib/jvm/java-openjdk
+
+## jenkins instalation
+ENV JENKINS_HOME /var/jenkins_home
```
slaskawi: Why we are not using standard jenkins installation directory `/var/lib/jenkins` ?

docker-environment-driver/src/main/resources/isolatedDockerConfiguration/Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+VOLUME /var/jenkins_home
+# `/usr/share/jenkins/ref/` contains all reference configuration we want
+# to set on a fresh new installation. Use it to bundle additional plugins
+# or config file with your custom jenkins Docker image.
+RUN mkdir -p /usr/share/jenkins/ref/init.groovy.d
+COPY init.groovy /usr/share/jenkins/ref/init.groovy.d/tcp-slave-angent-port.groovy
+ENV JENKINS_VERSION 1.554.1
+# could use ADD but this one does not check Last-Modified header
+# see https://github.com/docker/docker/issues/8331
+RUN curl -L http://mirrors.jenkins-ci.org/war-stable/$JENKINS_VERSION/jenkins.war -o /usr/share/jenkins/jenkins.war
```
slaskawi: Just a thought - wouldn't that be better to install Jenkins using yum?

docker-environment-driver/src/main/resources/dockerConfiguration/Dockerfile
mareknovotny: don't get your question, artifacts? maven is just installed from binary, installing from rpms is not prefered way due other depencies it could bring and grow docker image size.

docker-environment-driver/src/main/resources/isolatedDockerConfiguration/Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+
+# SSH login fix. Otherwise user is kicked off after login
+RUN sed 's@session\s*required\s*pam_loginuid.so@session optional pam_loginuid.so@g' -i /etc/pam.d/sshd
+ENV NOTVISIBLE "in users profile"
+RUN echo "export VISIBLE=now" >> /etc/profile
+
+#Install maven
+# set installed Maven version you can easily change it later
+ENV MAVEN_VERSION 3.2.5
+RUN curl -sSL http://archive.apache.org/dist/maven/maven-3/$MAVEN_VERSION/binaries/apache-maven-$MAVEN_VERSION-bin.tar.gz | tar xzf - -C /usr/share \
```
slaskawi: And here we have the same situation - why can't we use yum?

docker-environment-driver/src/main/resources/isolatedDockerConfiguration/init.groovy
```diff
@@ -0,0 +1,9 @@
+import hudson.model.*;
+import jenkins.model.*;
+
+
+Thread.start {
+      sleep 10000
+      println "--> setting agent port for jnlp"
+      Jenkins.instance.setSlaveAgentPort(50000)
```
slaskawi: Perhaps it would be a bit simpler to use Jenkins slave.jar file. It should do it automatically without involving custom groovy script here.

docker-environment-driver/src/main/resources/dockerConfiguration/Dockerfile
slaskawi: Sorry I didn't make myself clear - I was referring to `export M2_HOME=/usr/share/maven`. Why shouldn't we use simply `~/.m2/repository`?

docker-environment-driver/src/main/resources/isolatedDockerConfiguration/Dockerfile
mareknovotny: we could, even it is much better to use default installed java from base image - in this case it is java 8. We can fairly easily replace this later if for instance we need to use Oracle jdk or other vendor installations.

docker-environment-driver/src/main/resources/isolatedDockerConfiguration/Dockerfile
mareknovotny: we need to tidy up jenkins installation anyway, I am not sold to have jenkins master in a default build image. we should have there only useful things in that image.

docker-environment-driver/src/main/resources/isolatedDockerConfiguration/Dockerfile
mareknovotny: not, I just copied what is in @jbartece first docker image - he used jenkins image based on ubuntu from cloudbees repository. so I didn't want to change that for the initial commit.

docker-environment-driver/src/main/resources/isolatedDockerConfiguration/Dockerfile
mareknovotny: that is the question, even in Docker best practices is recommended to use binaries not installing  rpms for other than strictly distro system dependencies

docker-environment-driver/src/main/resources/isolatedDockerConfiguration/Dockerfile
mareknovotny: the same answer, rpm based installation for maven or jenkins is different from original distributed binaries.

docker-environment-driver/src/main/resources/isolatedDockerConfiguration/init.groovy
mareknovotny: this and 2 more scripts is just copy for @jbartece's jenkins which I don't want to change now as I didn't know what else is dependent on its exact setup and installation.

docker-environment-driver/src/main/resources/dockerConfiguration/Dockerfile
mareknovotny: you probably mismatch that with default repository path, *M2_HOME=/usr/share/maven* is env variable for installed Maven binaries not for local maven repository path ;)

docker-environment-driver/src/main/resources/dockerConfiguration/Dockerfile
jbartece: What's the order/priority of loading Maven configs? I just want to make sure, it won't hide settings.xml, which will DockerEnvironmentDriver copy to ~/-m2/settings.xml.

docker-environment-driver/src/main/resources/isolatedDockerConfiguration/Dockerfile
slaskawi: We agreed to leave it as is.

docker-environment-driver/src/main/resources/dockerConfiguration/Dockerfile
slaskawi: I'll just paste a quote from: http://maven.apache.org/settings.html:

>There are two locations where a settings.xml file may live:
>The Maven install: $M2_HOME/conf/settings.xml
>A user’s install: ${user.home}/.m2/settings.xml
>The former settings.xml are also called global settings, the latter settings.xml are referred to as user settings. If both files exists, their contents gets merged, with the user-specific settings.xml being dominant.

The most important part - they are actually merged. We need to have this in mind when playing with mirrors in both of those files.

mareknovotny: well, we should eliminate the Maven user settings in product building like it is now for MEAD.
just one preinstalled one.
What is flexible is that URL in global settings.xml you can set it up for every running container differently by environment variable

jbartece: @matejonnet wanted to set up AProx URLs via DockerEnvironmentDriver in runtime. So it has to be set up using user settings.

mareknovotny: don't understand that, you can still set up this URL through environment variable --env mavenRepositoryUrl=<someUrl>. So why is there "has to be in user settings.xml" ?

jbartece: That's the config: https://github.com/project-ncl/pnc/blob/master/docker-environment-driver/src/main/resources/jenkins-maven-config.xml

mareknovotny: @jbartece I don't care so much about the content of settings.xml even the profile repo is a little weird. It is still better to have the configuration in docker place as a global one not user's settings.xml. So we could move it to docker image IMHO.

jbartece: I didn't create it so I cannot provide satisfactory answer :-). Maybe @matejonnet could provide you better answer

PR #193, opened 2015-03-06
Integrate environment driver into BuildOrchestrator workflow

jbartece:


common/src/main/java/org/jboss/pnc/common/json/moduleconfig/JenkinsBuildDriverModuleConfig.java
```diff
@@ -39,6 +29,6 @@ public void setPassword(String password) {
     }
     @Override
     public String toString() {
-        return "JenkinsBuildDriverModuleConfig [url=" + url + ", username=" + username + ", password=" + password + "]";
+        return "JenkinsBuildDriverModuleConfig [username=" + username + ", password=" + password + "]";
     }
```
matejonnet: I would remove password from toString to avoid printing it in logs.

jbartece: Done

PR #199, opened 2015-03-06
Make Docker driver image configurable

jbartece:


jbartece:
Tested locally - all tests passed. Jenkins and Docker host is down, so it cannot be tested on CI.

PR #202, opened 2015-03-08
0.3 integration bug fixes

matejonnet:


pnc-core/src/main/java/org/jboss/pnc/core/BuildDriverFactory.java
```diff
@@ -25,7 +25,7 @@ public BuildDriver getBuildDriver(BuildType buildType) throws CoreException {
             }
         }
 
-        throw new CoreException("No build driver available for " + buildType + " build type.");
+        throw new CoreException("No build driver available for " + buildType + " build type."); //TODO create test to make sure exception is properly handled
```
slaskawi: Todo to be removed

pnc-core/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
```diff
@@ -1,23 +1,5 @@
 package org.jboss.pnc.core.builder;
```
slaskawi: This change shouldn't be there

slaskawi:
retest this please

pnc-core/src/main/java/org/jboss/pnc/core/BuildDriverFactory.java
matejonnet: Why do you think it's wrong if TODO is left there, we are not GA yet? Before removing it we need to make sure that it is done. Do you think we should create jira for it? I think Jira ticket has to much overhead for such task.

slaskawi: I think we should either implement this test straight away (my preference) or if we can't do that for some reason - we should create a JIRA for that - not a TODO in the code. Without JIRA we won't be able to track it.

pnc-core/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
matejonnet: That's an issue because IDEs are automatically arranging imports. We need need to share the same formatter settings to avoid it.

slaskawi: I don't think that commiting a file without meaningful changes is a good practice. It adds only a noise.

pnc-core/src/main/java/org/jboss/pnc/core/BuildDriverFactory.java
matejonnet: This is opening wider discussion that is out of scope of this PR, do we allow TODOs in commited code? Maybe we need a Jira to track the discussion ;)

matejonnet:
Retest this please.

matejonnet:
Retest this please.

slaskawi:
retest this please

PR #210, opened 2015-03-10
[NCL-588]: Destroy Docker container in case of system exception

jbartece:


jbartece:
retest this please

jbartece:
retest this please

jbartece:
retest this please

pnc-core/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
```diff
@@ -317,6 +330,24 @@ public BuildTask getBuild(String identifier) {
...
+        private RunningEnvironment runningEnvironment;
+        
+        public RunningEnvironment getRunningEnvironment() {
+            return runningEnvironment;
+        }
+        
+        public void setRunningEnvironment(RunningEnvironment runningEnvironment) {
+            this.runningEnvironment = runningEnvironment;
+        }
+    }
```
matejonnet: You can use generic org.jboss.pnc.common.util.ObjectWrapper instead of specific RunningEnvironmentWrapper.

jbartece: OK. I'll make a note and fix it later

pnc-core/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
```diff
@@ -153,15 +154,17 @@ void startBuilding(BuildTask buildTask) throws CoreException {
...
         configureRepository(buildTask, repositoryManager)
-                .thenCompose(repositoryConfiguration -> setUpEnvironment(buildTask, envDriver, repositoryConfiguration))
+                .thenCompose(repositoryConfiguration -> setUpEnvironment(buildTask, envDriver, repositoryConfiguration, backupRunningEnvironment))
                 .thenCompose(runningEnvironment -> buildSetUp(buildTask, buildDriver, runningEnvironment))
                 .thenCompose(runningBuild -> waitBuildToComplete(buildTask, runningBuild))
                 .thenCompose(completedBuild -> retrieveBuildDriverResults(buildTask, completedBuild))
                 .thenCompose(buildDriverResult -> retrieveRepositoryManagerResults(buildTask, buildDriverResult))
                 .thenCompose(buildResults -> destroyEnvironment(buildTask, buildResults))
-                .handle((buildResults, e) -> storeResults(buildTask, buildResults, e));
+                .handle((buildResults, e) -> storeResults(buildTask, buildResults, backupRunningEnvironment, e));
```
matejonnet: No the best passing of reference `backupRunningEnvironment`, can't it be accessed through "chain", each has link to the previous result.

jbartece: I think it cannot be done by passing it through the chain, because if some part fails than the method storeResults is called with buildResults  == null. So it has to be somehow backuped out of the chain.

matejonnet: Can we add information required to destroy environment to the exception, by adding new field to our exception type ?

jbartece: Good idea. I'll do it

PR #215, opened 2015-03-11
NCL-286 follow-up

pslegr:


pnc-web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -1,6 +1,32 @@
...
+  </listener>
+  <security-constraint>
+    <web-resource-collection>
+      <web-resource-name>pnc-servlet</web-resource-name>
+      <url-pattern>/*</url-pattern>
+      <http-method>GET</http-method>
+      <http-method>POST</http-method>
+    </web-resource-collection>
+    <auth-constraint>
+      <role-name>admin</role-name>
```
slaskawi: Hey! I'm definitely not an expert in case of authentication/authorization but shouldn't we have a Keycloak server in our environment? We will probably need LDAP integration, OAuth for REST and probably Kerberos integration or SSO integration. Seems to me like a perfect scenario for Keycloak.

pnc-core/src/test/java/org/jboss/pnc/core/test/buildCoordinator/ProjectBuilder.java
```diff
@@ -49,6 +50,7 @@ public static JavaArchive createDeployment() {
         JavaArchive jar = ShrinkWrap.create(JavaArchive.class)
                 .addClass(Configuration.class)
                 .addClass(Environment.Builder.class)
+                .addClass(AuthenticationProvider.class)
```
slaskawi: I would like to have a separate integration test for authentication/authorization.

slaskawi:
Great work but I would like to learn more about how do we want to implement authorization/authentication in NCL ecosystem.

pnc-web/src/main/webapp/WEB-INF/web.xml
pslegr: Hey Sebastian,

well I am learning as well ;)
However, this is just a start... EAP security domains can be configured 
for either Kerberos or LDAP or for any desired Authentication Provider 
we will choose in the future.
OAuth for Rest is on of the possibilities, it depends on what the 
requirements are
Rest endpoints can be also secured via @RoleAllowed annotations, once we 
define the
roles of endpoints

I don't know much in detail about Keycloak, but I will have a look into it.

SSO, would come in the future if there is need for it, once we do have 
more web sites ?
In case there is plan to... currently we have pnc-web only.

I suggest go step-by-step and not introduce a pollute NCL with things, 
we do not necessarily be using now.

I am going to dig deeper and elaborate on this

One more thing... we need to split this into more issues, and well 
define, what and when to implement.
P.

jenkins-build-driver/pom.xml
```diff
@@ -127,6 +127,11 @@
       <artifactId>org.jacoco.core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jboss.spec.javax.ejb</groupId>
+      <artifactId>jboss-ejb-api_3.1_spec</artifactId>
+      <scope>test</scope>
+    </dependency>
     <!-- /Test coverage -->
```
matejonnet: Do we really need a dependency to EJB in a driver tests ?

slaskawi:
@pslegr I think this can be closed right?

pslegr:
yes, closing

PR #218, opened 2015-03-11
NCL-608 Removed persisting cascades

slaskawi:
* Now everything needs to be saved separately
* This will hopefully prevent us from accidentally removing data, just like it happened in NCL-608


slaskawi:
retest this please

PR #220, opened 2015-03-12
NCL-592 added jenkins system configuration for git/java/maven binaries a...

mareknovotny:
...nd customized Dockerfile with these files

### DIFF
--- a/jenkins-build-driver/src/main/java/org/jboss/pnc/jenkinsbuilddriver/BuildJob.java
+++ b/jenkins-build-driver/src/main/java/org/jboss/pnc/jenkinsbuilddriver/BuildJob.java
@@ -20,6 +20,8 @@ class BuildJob {
     private JobWithDetails job;
     private int buildNumber;
     
+    private boolean crumbflag;
+    
     private static final Logger log = Logger.getLogger(BuildJob.class.getName());
 
     public BuildJob(JenkinsServer jenkinsServer, BuildConfiguration buildConfiguration) {
@@ -27,9 +29,12 @@ class BuildJob {
         this.buildConfiguration = buildConfiguration;
     }
 
-    public boolean configure(RunningEnvironment runningEnvironment, boolean override) throws BuildDriverException {
+    public boolean configure(RunningEnvironment runningEnvironment, boolean override, boolean crumbFlag) throws BuildDriverException {
         String jobName = getJobName();
-
+        
+        // is jenkins using Default Crumb Issuer - Prevent Cross Site Request Forgery exploits
+        this.crumbflag = crumbFlag;
+        
         this.buildJobConfig = new BuildJobConfig(
                 jobName,
                 buildConfiguration.getScmRepoURL(),
@@ -39,20 +44,19 @@ class BuildJob {
         try {
             job = jenkinsServer.getJob(jobName);
         } catch (IOException e) {
-            // mnovotny: no job exist so throwing exception is effectively stopping configuration 
-            //throw new BuildDriverException("Cannot check for existing job.", e);           
+            throw new BuildDriverException("Cannot check for existing job.", e);
         }
 
         try {
             if (job != null) {
                 if (override) {
-                    jenkinsServer.updateJob(jobName, buildJobConfig.getXml(), false);
+                    jenkinsServer.updateJob(jobName, buildJobConfig.getXml(), crumbflag);
                 } else {
                     throw new BuildDriverException("Cannot update existing job without 'override=true'."); 
                 }
             } else {
                 log.info("job config:\n" + buildJobConfig.getXml());
-                jenkinsServer.createJob(jobName, buildJobConfig.getXml(), false);
+                jenkinsServer.createJob(jobName, buildJobConfig.getXml(), crumbflag);
             }
         } catch (IOException e) {
             throw new BuildDriverException("Cannot create/update job.", e);
@@ -74,7 +78,7 @@ class BuildJob {
         buildNumber = -1;
         try {
             buildNumber = job.getNextBuildNumber();
-            job.build(false);
+            job.build(crumbflag);
         } catch (IOException e) {
             throw new BuildDriverException("Cannot start project build.", e);
         }
--- a/jenkins-build-driver/src/main/java/org/jboss/pnc/jenkinsbuilddriver/BuildJob.java
+++ b/jenkins-build-driver/src/main/java/org/jboss/pnc/jenkinsbuilddriver/BuildJob.java
@@ -1,12 +1,14 @@
 package org.jboss.pnc.jenkinsbuilddriver;
 
-import com.offbytwo.jenkins.JenkinsServer;
-import com.offbytwo.jenkins.model.JobWithDetails;
+import java.io.IOException;
+import java.util.logging.Logger;
+
 import org.jboss.pnc.model.BuildConfiguration;
 import org.jboss.pnc.spi.builddriver.exception.BuildDriverException;
 import org.jboss.pnc.spi.environment.RunningEnvironment;
 
-import java.io.IOException;
+import com.offbytwo.jenkins.JenkinsServer;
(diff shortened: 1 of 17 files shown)


slaskawi:
:+1:

jbartece:
+1

mareknovotny:
I have fixed JenkinsDriverRemoteTest, the only issue here is the hardcoded jenkins port in test

mareknovotny:
retest this please

jbartece:
Using which server do you want to run this test?

mareknovotny:
the test should connect to docker host with running container AFAIK and submit the jenkins build job. CI is setup to run on
```
export PNC_JENKINS_URL=http://10.3.8.18
export PNC_JENKINS_PORT=8081
```

mareknovotny:
I have found that our jenkins server is setup with option 'Prevent Cross Site Request Forgery exploits' which prevents pure rest-api calls without a token to call jenkins API and the test fails due that. Changing of that option will cause passing the test. So we need to add that feature to jenkins Build Driver to pass the Jenkins CI with such config too.

mareknovotny:
retest this please

mareknovotny:
retest this please

matejonnet:
What about using the client lib that is used in BuildDriver it should deal with tokens.

mareknovotny:
Not sure what you mean by "the client lib that is used in BuildDriver"

matejonnet:
JenkinsBuildDriver uses com.offbytwo.jenkins:jenkins-client instead of direct REST calls, the libs can deal with tokens. See JenkinsServerFactory and its usage.
Take a look if you can use this lib instead of direct calls.

mareknovotny:
I have used JenkinsHttpClient for that Scrumb issuer header ;)

jenkins-build-driver/src/main/java/org/jboss/pnc/jenkinsbuilddriver/BuildJob.java
```diff
@@ -35,19 +44,19 @@ public boolean configure(RunningEnvironment runningEnvironment, boolean override
         try {
             job = jenkinsServer.getJob(jobName);
         } catch (IOException e) {
-            throw new BuildDriverException("Cannot check for existing job.", e);
+            log.info("No job exists, so there will be created new one.");
```
matejonnet: I would expect IOException when something went wrong not when the job does not exists. This exception should be re-thrown.

jenkins-build-driver/src/main/java/org/jboss/pnc/jenkinsbuilddriver/JenkinsBuildDriver.java
```diff
@@ -24,13 +24,15 @@
 
     private JenkinsServerFactory jenkinsServerFactory;
     private JenkinsBuildMonitor jenkinsBuildMonitor;
+    
+    private boolean isCrumbUsed;
 
     JenkinsBuildDriver() {}
```
matejonnet: Add @Deprecated to indicate that this constructor is not meant to use.

jenkins-build-driver/src/main/java/org/jboss/pnc/jenkinsbuilddriver/JenkinsBuildDriver.java
```diff
@@ -42,11 +44,22 @@ public String getDriverId() {
     public boolean canBuild(BuildType buildType) {
         return BuildType.JAVA.equals(buildType);
     }
+    
+    public void init(RunningEnvironment runningEnvironment) {
+        try {
+            this.isCrumbUsed = jenkinsServerFactory.isJenkinsServerSecuredWithCSRF(runningEnvironment.getJenkinsUrl());    
+        } catch ( BuildDriverException bde) {
```
matejonnet: If it safe to swallow this exception, can we continue when init fails ? Mybe we shold set the default value to isCrumbUsed ?

jenkins-build-driver/src/test/java/org/jboss/pnc/jenkinsbuilddriver/test/JenkinsDriverRemoteTest.java
```diff
@@ -143,12 +142,12 @@ public RepositorySession getRepositorySession() {
             
             @Override
             public String getJenkinsUrl() {
-                return System.getProperty("PNC_JENKINS_URL");
+                return System.getenv("PNC_JENKINS_URL") + ":" + getJenkinsPort();
```
matejonnet: Try to avoid using env properties it requires that user set them in order to run tests. Env variables we are using is just one option to pass values to the configuration. Can you use org.jboss.pnc.common.Configuration ?

jenkins-build-driver/src/main/java/org/jboss/pnc/jenkinsbuilddriver/BuildJob.java
mareknovotny: But if you rethrow it you never create a new job, the method jenkinsServer.getJob() can have two exceptions one as HttpResponseException which is still in IOException inheritance but there is other which are not caught in  jenkinsServer.getJob() method and that could be blocker for the case there is not job or there is something with Jenkins connection.

jenkins-build-driver/src/main/java/org/jboss/pnc/jenkinsbuilddriver/JenkinsBuildDriver.java
mareknovotny: Yes, if something is wrong with Jenkins server it would be reported later, that is just checking for flag, there is not really reason to assume that this is significant for following process.

jenkins-build-driver/src/main/java/org/jboss/pnc/jenkinsbuilddriver/BuildJob.java
matejonnet: Each exception you catch must be handled, in case of IOException, I see two options, you can retry or fail.
I think in our case it is safe to fail and cancel the build by trowing en exception. The error will be logged in BuildRecord which is better than system log that won't be available to users.

jenkins-build-driver/src/main/java/org/jboss/pnc/jenkinsbuilddriver/JenkinsServerFactory.java
```diff
@@ -38,4 +42,36 @@ JenkinsServer getJenkinsServer(String url) throws BuildDriverException {
...
             throw new BuildDriverException("Cannot read configuration for " + JenkinsBuildDriver.DRIVER_ID + ".", e);        }
     }
+    
+    /**
+     * This checks if jenkins does not use option Prevent Cross Site Request Forgery exploits
+     * 
+     * @param url Jenkins instance URL with port
+     * @return
+     * @throws BuildDriverException
+     */
```
matejonnet: XFactory class is responsible for instantiating X classes, this method does not fit well here.

jenkins-build-driver/src/test/java/org/jboss/pnc/jenkinsbuilddriver/test/JenkinsDriverRemoteTest.java
mareknovotny: Matej, this is only a test usage, in real build you will get it from Docker Build driver. Test environment needs to be flexible and common configuration doesn't fit to this need IMO.

matejonnet: Will developers be able to run this test without setting the System env variable ?

mareknovotny: No, but I added them into README.md file where are they described.

PR #232, opened 2015-03-21
[NCL-651] Add NotNull contstraints to JPA model

pgier:
Make the DB more strict regarding data integrity by not allowing null
values in several fields.


pgier:
retest this please

pnc-model/src/main/java/org/jboss/pnc/model/BuildRecord.java
```diff
@@ -57,10 +59,13 @@
...
 
+    @NotNull
     private Timestamp startTime;
 
+    @NotNull
     private Timestamp endTime;
 
+    @NotNull
     @ManyToOne
     private User user;
```
vibe13: Since we are going to use Envers for the BuildConfiguration parameters, ok for now not to mark as NotNull the "name" field

pnc-model/src/main/java/org/jboss/pnc/model/BuildConfiguration.java
```diff
@@ -54,9 +56,11 @@
     @ManyToOne(cascade = { CascadeType.MERGE, CascadeType.PERSIST, CascadeType.REFRESH })
     private ProductVersion productVersion;
```
vibe13: @pgier  Just looking at the model, didn't we removed ProductVersion from BuildConfiguration? And have it only inside BuildConfiguratioSet?

pgier: We replaced productversionsprojects with buildconfigurationset, but I don't think we specifically said we would remove this field.  However, I do agree that it should be removed, and there is a jira for it: https://projects.engineering.redhat.com/browse/NCL-406

pnc-model/src/main/java/org/jboss/pnc/model/BuildRecord.java
pgier: Did you mean the "user" field?  I think even with auditing of changes to BuildConfiguration, we'd still need to record the "user" in the build record to know who triggered the build.

PR #241, opened 2015-03-26
[NCL-477] Return HTTP 400 when an id is sent for a new build config

pgier:
If the client sends the id in the POST request for a new config we
should return an HTTP 400 (bad request), because we require that
the server generates the correct next id in the sequence.


pnc-rest/src/main/java/org/jboss/pnc/rest/endpoint/BuildConfigurationEndpoint.java
```diff
@@ -56,6 +56,9 @@ public BuildConfigurationEndpoint(BuildConfigurationProvider buildConfigurationP
     @ApiOperation(value = "Creates a new Build Configuration")
     @POST
     public Response createNew(@NotNull @Valid BuildConfigurationRest buildConfigurationRest, @Context UriInfo uriInfo) {
+        if (buildConfigurationRest.getId() != null) {
```
slaskawi: Consider using Guava's Predications in buildConfigurationProvider#store. This way the provider could be reused and IllegalArgumentException will be automatically translated into HTTP 400

PR #242, opened 2015-03-26
Pnc keycloak

pslegr:


common/pom.xml
```diff
@@ -56,6 +56,23 @@
...
+        <artifactId>jboss-servlet-api_3.0_spec</artifactId>
+        <scope>provided</scope>
+    </dependency>
+    <dependency>
+        <groupId>org.keycloak</groupId>
+        <artifactId>keycloak-core</artifactId>
+        <scope>provided</scope>
+    </dependency>
+    <!-- dependency>
+        <groupId>org.keycloak</groupId>
```
slaskawi: Please remove commented code

common/src/main/java/org/jboss/pnc/common/auth/AuthenticationProvider.java
```diff
@@ -0,0 +1,52 @@
+package org.jboss.pnc.common.auth;
```
slaskawi: Since this requires Servet spec, wouldn't pnc-rest be a better module for that?

common/src/main/java/org/jboss/pnc/common/auth/AuthenticationProvider.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.Set;
+
+import javax.servlet.http.HttpServletRequest;
+
+import org.keycloak.KeycloakSecurityContext;
+import org.keycloak.representations.AccessToken;
+
+
+
```
slaskawi: Formatting doesn't look right in this class, lots of empty lines etc.

pnc-rest/src/main/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,68 @@
...
+        <web-resource-collection>
+            <web-resource-name>configuration-rest-endpoint_all</web-resource-name>
+            <url-pattern>/rest/configuration/*</url-pattern>
+        </web-resource-collection>
+        <auth-constraint>
+            <role-name>config-manager</role-name>
+        </auth-constraint>
+    </security-constraint>
+    
+
```
slaskawi: Formatting

pnc-ui/Gruntfile.js
```diff
@@ -415,6 +415,7 @@ module.exports = function (grunt) {
             '!index.html',
             'images/{,*/}*.{webp}',
             'fonts/{,*/}*.*',
+            'keycloak.json'
```
slaskawi: So the keycloak file is also added to pnc-web? Do we need to secure frontend also?

pnc-ui/app/keycloak.json
```diff
@@ -0,0 +1,21 @@
+{
+  "realm" : "PNC.REDHAT.COM",
+  "resource" : "pnc-web-local",
+  "realm-public-key" : "MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQCF6WP9SC6i9ulVqAItdu1qhsN9LoIywlJ3Qw7RUIMmPjKToCnDoq8glkJAkS4I3PbpWaFSBRvSObmAm4XvGv69kl1UPUW76Wgyqvr09fNSNds34MV4NuULE1muHHj3x49RqH4J/UPQvICtE77t+lvjd526uqHPyn+mZlN4B0K6hQIDAQAB",
+  "auth-server-url" : "https://keycloak-pncauth.rhcloud.com/auth",
+  "ssl-required" : "none",
+  "use-resource-role-mappings" : false,
+  "enable-cors" : true,
+  "cors-max-age" : 1000,
```
slaskawi: TBH - I think we need security only in REST module. Securing Web App doesn't make much sense to me. On the other hand if we must secure both of them - maybe we should merge them into single WAR?

pnc-web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -1,6 +1,5 @@
...
-<web-app xmlns="http://xmlns.jcp.org/xml/ns/javaee"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/web-app_3_1.xsd"
-         version="3.1">
-</web-app>
\ No newline at end of file
+<web-app xmlns="http://java.sun.com/xml/ns/javaee"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd"
+      version="3.0">
```
slaskawi: Empty web.xml is not needed in JEE6

common/pom.xml
```diff
@@ -56,6 +56,23 @@
...
     </dependency>
+    
+    <dependency>
+        <groupId>org.jboss.spec.javax.servlet</groupId>
+        <artifactId>jboss-servlet-api_3.0_spec</artifactId>
+        <scope>provided</scope>
+    </dependency>
+    <dependency>
+        <groupId>org.keycloak</groupId>
+        <artifactId>keycloak-core</artifactId>
```
slaskawi: It seems we use Keycloak from the server right? If so, do we need to change anything in standalone.xml to turn it on/configure?

slaskawi:
Hey @pslegr !

Great work :) 
My biggest concern is whether we need to add security to UI module (or just adding it to REST is sufficient)? Could you please also squash your commits and add JIRA to commit message?

Thanks!

pnc-ui/app/keycloak.json
matejonnet: We should keep REST and WEB in different WARs so that you can use (also build) REST without a WEB UI.

pslegr: +1
Yes, we do have them in separate wars
we need to have 2 contexts
/pnc-web
/pnc-rest

To be able external apps to access PNC REST backend

pslegr:
retest this please

slaskawi:
How about extracting auth from commons to separate module? It seems we have at least 2 votes for that on mailing list. Of course it may be done with another PR if you wish...

pslegr:
yes, I will extract that into auth module

integration-test/src/test/resources/keycloak-standalone.xml
```diff
@@ -0,0 +1,380 @@
...
+        <extension module="org.jboss.as.pojo"/>
+        <extension module="org.jboss.as.remoting"/>
+        <extension module="org.jboss.as.sar"/>
+        <extension module="org.jboss.as.security"/>
+        <extension module="org.jboss.as.threads"/>
+        <extension module="org.jboss.as.transactions"/>
+        <extension module="org.jboss.as.web"/>
+        <extension module="org.jboss.as.webservices"/>
+        <extension module="org.jboss.as.weld"/>
+        <extension module="org.keycloak.keycloak-subsystem"/>
```
rnc: If we are storing a modified standalone xml could we use CLI scripts and modify the server via the admin commands instead?

pslegr: yes I will change it to that approach

PR #243, opened 2015-03-26
Test profiles refactoring

slaskawi:
Hey!

Please take a look at first part of test refactoring.

Highlights:
* There are 2 profiles for running tests - container-tests and remote-tests
* Each container or remote test should have proper c@Category
* Each test type is triggered in it's own phase
* Server is unpacked only once

More comments inlined into the code pieces.


common/pom.xml
```diff
@@ -57,6 +57,12 @@
       <artifactId>resteasy-jaxrs</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.jboss.pnc</groupId>
```
slaskawi: This is the only dependency needed for creating standard tests.

docker-environment-driver/pom.xml
```diff
@@ -1,213 +1,138 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.jboss.pnc</groupId>
+      <artifactId>test-arquillian-container</artifactId>
```
slaskawi: This dependency is needed when creating Arquillian tests which are executed inside a container.

integration-test/src/test/java/org/jboss/pnc/integration/deployments/Deployments.java
```diff
@@ -45,19 +44,19 @@ private static void setTestableWar(EnterpriseArchive ear) {
...
-    }
-
     private static void addEar(Archive<?> webArchive, PomEquippedResolveStage mavenResolver) {
         File[] manuallyAddedLibs = mavenResolver.resolve("org.jboss.pnc:ear-package:ear:1.0-SNAPSHOT").withoutTransitivity().asFile();
         Stream.of(manuallyAddedLibs).forEach(lib -> webArchive.merge(ShrinkWrap.create(ZipImporter.class).
                 importFrom(lib).as(GenericArchive.class)));
     }
+
+    private static void addTestCommonWithTransitives(EnterpriseArchive webArchive, PomEquippedResolveStage mavenResolver) {
+        File[] manuallyAddedLibs = mavenResolver.resolve("org.jboss.pnc:test-common").withTransitivity().asFile();
```
slaskawi: We need test categories for Arquillian

pom.xml
```diff
@@ -927,6 +952,79 @@
...
+        <configuration>
+          <properties>
+            <property>
+              <name>listener</name>
+              <value>org.jboss.pnc.test.listener.PrintTestNameListener</value>
+            </property>
+          </properties>
+        </configuration>
+        <executions>
+          <execution>
```
slaskawi: Each tests category is executed in separate phase - this gives us flexibility (e.g. run unit tests with fork count 2C; remote and container tests without forks).

pom.xml
```diff
@@ -966,34 +1111,75 @@
         <persistence.hibernate.hbm2ddl.auto>create</persistence.hibernate.hbm2ddl.auto>
       </properties>
     </profile>
-    
+
```
slaskawi: All profiles which modify properties are placed here. This increases readability.

test-arquillian-container/pom.xml
```diff
@@ -0,0 +1,275 @@
...
+                  </artifactItems>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+
+    <profile>
```
slaskawi: All Wildfly/EAP downloading profiles are placed here. We want to download everything only once.

test-common/src/main/java/org/jboss/pnc/test/listener/PrintTestNameListener.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.util.logging.Logger;
+
+public class PrintTestNameListener extends RunListener {
+
+    private Logger logger = Logger.getLogger(PrintTestNameListener.class.getName());
+
+    @Override
+    public void testStarted(Description description) {
+        logger.info("Running test: " + description.getClassName() + "#" + description.getMethodName());
```
slaskawi: Hopefully this will make debugging on Jenkins a bit easier.

slaskawi:
retest this please

slaskawi:
@rnc @matejonnet @mareknovotny @vibe13 @pgier @jdcasey Please take a look at this when you have a while.

rnc:
Looks fine ; talked to @slaskawi offline. 

I think its removed the capability to use a local running Wildfly via `-Darquillian.container.type=remote` (we can always add it back if we want to keep it)

While it might be cleaner to move the test code into appropriate modules this can be done another time. It could well be a fine line on where to put tests in future (depends on whether the test is exercising multiple layers)

slaskawi:
@matejonnet @mareknovotny @vibe13 @pgier @jdcasey last chance, I'll address Nick's comments soon and then I will integrate it

slaskawi:
@rnc Updated, it should be fine now.

matejonnet:
What about adding your pull request line notes to code as comments?

slaskawi:
Added test modules description in `pom.xml`s.

docker-environment-driver/src/test/java/org/jboss/pnc/environment/docker/DockerEnvironmentDriverRemoteTest.java
```diff
@@ -31,22 +15,36 @@
...
+
 /**
  * Unit tests for DockerEnnvironmentDriver
- * 
- * @author Jakub Bartecek <jbartece@redhat.com>
  *
+ * @author Jakub Bartecek <jbartece@redhat.com>
  */
 @RunWith(Arquillian.class)
+@Category({RemoteTest.class, ContainerTest.class})
```
mareknovotny: Why is here 2 categories assigned? why should Remote tests be launched in container testing group?

mareknovotny:
@slaskawi  Are you fixing README.md for running tests too ?

docker-environment-driver/src/test/java/org/jboss/pnc/environment/docker/DockerEnvironmentDriverRemoteTest.java
slaskawi: Now you need 2 profiles to run this test - `-Pcontainer-tests` and `-Premote-tests`. Now those 2 groups are separate.

slaskawi:
@mareknovotny Very good point, I'm on it.

slaskawi:
@mareknovotny READ.ME Comment addressed

slaskawi:
@mareknovotny READ.ME Comment addressed

docker-environment-driver/src/test/java/org/jboss/pnc/environment/docker/DockerEnvironmentDriverRemoteTest.java
mareknovotny: I think this test is RemoteTest and is launched in  arquillian-weld-se-embedded container, so there will be duplicated container test IMO. WDYT?

slaskawi: In that case I should have used only RemoteTests category. I'll 
introduce this change shortly.

slaskawi: Double checked - this was a container-based test before my changes (see [1]). It should behave exactly the same way after my changes - container-based test which connects to external resources. That's why we need to 2 profiles there.

[1] https://github.com/project-ncl/pnc/pull/243/files#diff-d67f3d8a4636c41bf582c63a0f397bc9L141

slaskawi:
retest this please

slaskawi:
I see some very weird test failures: http://10.3.10.33:8081/job/NCL_SNAPSHOT/229/

Reverting.

slaskawi:
Reverted: https://github.com/project-ncl/pnc/commit/8956db187b6df90302a85a56939429bde32d406e

Will update it tomorrow

PR #254, opened 2015-04-01
Ncl 666

pgier:
Require null entity ID in the body of the request when the ID is available in the URL.


pnc-rest/src/main/java/org/jboss/pnc/rest/endpoint/UserEndpoint.java
```diff
@@ -49,6 +53,7 @@ public UserRest getSpecific(@ApiParam(value = "User id", required = true) @PathP
     @ApiOperation(value = "Creates new User")
     @POST
     public Response createNew(@NotNull @Valid UserRest userRest, @Context UriInfo uriInfo) {
+        validateIdIsNull(userRest);
```
slaskawi: Just a thought - wouldn't be better to validate inside `userProvider`? This way the endpoint would be anemic and testing might be done one step lower - on provider layer. We wouldn't need Arquillian and possibly plain JUnit would be sufficient this way...

pnc-rest/src/main/java/org/jboss/pnc/rest/endpoint/ProjectEndpoint.java
```diff
@@ -59,8 +62,10 @@ public Response createNew(@NotNull @Valid ProjectRest projectRest, @Context UriI
     @ApiOperation(value = "Updates an existing Project")
     @PUT
     @Path("/{id}")
-    public Response update(@ApiParam(value = "Project id", required = true) @PathParam("id") Integer productId,
+    public Response update(@ApiParam(value = "Project id", required = true) @PathParam("id") Integer id,
             @NotNull @Valid ProjectRest projectRest, @Context UriInfo uriInfo) {
+        validateIdIsNull(projectRest);
```
slaskawi: And some another idea - Couldn't we use Validation API here and create some annotation? The entity is already processed by Validation api, so I think - all we need to do is to create just another validation annotation and use it here[1].

[1] https://docs.jboss.org/hibernate/validator/4.0.1/reference/en/html/validator-customconstraints.html

pnc-rest/src/main/java/org/jboss/pnc/rest/endpoint/UserEndpoint.java
pgier: Yeah, I think that's a good idea.  It will require changes to the method params in the provider classes, but I think that's ok.  Taking this a step further, do you think it would be good to make the endpoint classes all interfaces, so that they have no functionality and only define the API?

pnc-rest/src/main/java/org/jboss/pnc/rest/endpoint/ProjectEndpoint.java
pgier: I thought about using bean validation for these checks, but I wasn't sure how well it would work because the validations are situational, i.e. in this case you want the ids to be null, but in other cases you might require the id to be filled in.  I wasn't sure how to implement a conditional validation like that using annotations.

slaskawi: Actually I think in that case we would need 2 validation annotations. One for validating if `id` is present and the second one for validating if it's missing.

PR #255, opened 2015-04-01
Fixed executing remote tests without turning on its profile

slaskawi:
@rnc Sorry, this was a silly mistake :D


slaskawi:
retest this please

PR #256, opened 2015-04-01
[NCL-559]: Load configuration only once

jbartece:


common/src/main/java/org/jboss/pnc/common/Configuration.java
```diff
@@ -33,19 +37,25 @@
      * @return Loaded configuration
      * @throws ConfigurationParseException Thrown if configuration file couldn't be loaded or parsed
      */
-    public T getModuleConfig(Class<T> moduleClass) throws ConfigurationParseException {
+    @SuppressWarnings("unchecked")
+    public <T extends AbstractModuleConfig> T getModuleConfig(Class<T> moduleClass) throws ConfigurationParseException {
```
slaskawi: I believe we may end up with many configurations here, since there is no synchronization...

I believe we have a couple of options:
* Double check locking pattern (which is also broken :D)
* Use Guava's Cache Loader (or Infinispan!)
* Use EJB @Singleton with locking

The fastest is the third one, but I think the best one is the second.

common/src/test/java/org/jboss/pnc/common/json/moduleconfig/JenkinsBuildDriverModuleConfigTest.java
```diff
@@ -10,12 +10,11 @@
     @Test
     public void test() {
```
slaskawi: Ok, since we got the survey results, please refactor this test also:
* Add some meaningful name
* Divide into 3 blocks //given //when //then
* Get rid of this nasty catch block

common/src/main/java/org/jboss/pnc/common/Configuration.java
jbartece: Do we really need locks here? I think that the risk is just that one config will be loaded twice, but the locks has overhead too. Just a though, but I can change it to singleton for now

common/src/test/java/org/jboss/pnc/common/json/moduleconfig/JenkinsBuildDriverModuleConfigTest.java
jbartece: I didn't write it so I just changed what was needed, but you are right. I'll fix it

common/src/main/java/org/jboss/pnc/common/Configuration.java
slaskawi: I believe we should have at least some locking here. It won't happen probably on test environment, but believe - debugging this stuff on production is a nightmare...

common/src/test/java/org/jboss/pnc/common/json/moduleconfig/JenkinsBuildDriverModuleConfigTest.java
slaskawi: +1
> This opportunistic refactoring is referred to by Uncle Bob as following the boy-scout rule - always leave the code behind in a better state than you found it. If everyone on the team is doing this, they make small regular contributions to code base health every day.[1]

[1] http://martinfowler.com/bliki/OpportunisticRefactoring.html

common/src/main/java/org/jboss/pnc/common/Configuration.java
matejonnet: Common module should not use EJBs.

slaskawi: @matejonnet It uses CDI already, so why not EJB?

matejonnet: In contrast to EJB, CDI can be easily added to a Java SE application.

slaskawi: That's a good point, :+1:

rnc: Double check locking works in JDK5 and up? : 
http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html
http://www.javamex.com/tutorials/double_checked_locking_fixing.shtml

slaskawi: @rnc With the "volatille" trick - yes. Java Memory Model (JSR133) which was introduced in JDK5 standardizes it.

jbartece: @matejonnet @rnc @slaskawi So add synchronized to the method and volatile to the cache field?

slaskawi: @jbartece Slightly different - with a small synchronized block on `this`. Take a look at the example:
```
// Works with acquire/release semantics for volatile
// Broken under current semantics for volatile
  class Foo {
        private volatile Helper helper = null;
        public Helper getHelper() {
            if (helper == null) {
                synchronized(this) {
                    if (helper == null)
                        helper = new Helper();
                }
            }
            return helper;
        }
    }
```
taken from http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

Now the point of this whole trick is very simple - if the field is already initialized - we want to return quickly without acquiring a monitor (to be specific monitor on `this`). Now if the field is not initialized, we obtain the monitor and check it once again - in that brief moment some other thread could initialize it and because we use synchronization - there is a "happens-before" relation between those 2 events. Finally, if the field is still not initialized - we do the initialization with holding a monitor.

jbartece: @slaskawi I got the example, but in our case we have just one instance and the field is initiated in constructor. So I think we do not need the volatile. Or am I wrong?

I expect it could look like this, right?:

    public <T extends AbstractModuleConfig> T getModuleConfig(Class<T> moduleClass) throws ConfigurationParseException {
        if(configCache.containsKey(moduleClass))
            return (T) configCache.get(moduleClass);
        
        synchronized(this) {
            if(configCache.containsKey(moduleClass))
                return (T) configCache.get(moduleClass);
            
            try (InputStream configStream = this.getConfigStream()) {
                log.info("Loading configuration for class: " + moduleClass);
                String configString = StringUtils.replaceEnv(IoUtils.readStreamAsString(configStream));
                
                T config = new ConfigurationJSONParser().parseJSONConfig(configString, moduleClass);
                configCache.put(moduleClass, config);
                return config;
            } catch (IOException e) {
                throw new ConfigurationParseException("Config could not be parsed", e);
            }
        }
    }

slaskawi: @jbartece Yes, that's correct.

jbartece: @slaskawi Thanks for the explanation ;-)

jbartece:
retest this please

PR #265, opened 2015-04-02
[NCL-609]: Refactor Docker driver related class; exception workflow in B...

jbartece:
...uildCoordinator

### DIFF
--- a/pnc-core/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
+++ b/pnc-core/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
@@ -7,6 +7,7 @@ import org.jboss.pnc.core.BuildDriverFactory;
 import org.jboss.pnc.core.EnvironmentDriverFactory;
 import org.jboss.pnc.core.RepositoryManagerFactory;
 import org.jboss.pnc.core.content.ContentIdentityManager;
+import org.jboss.pnc.core.exception.BuildProcessException;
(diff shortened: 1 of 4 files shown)


jbartece:
@matejonnet Can you check this PR? There is a change in BuildCoordinator (propagating environment in case of exception), which you suggested in PR https://github.com/project-ncl/pnc/pull/210. But There is a problem: If some unexpected Exception there can happen environment leak - it is not expected, but some bug can cause it. What do you think about it?

pnc-core/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
```diff
@@ -260,7 +257,7 @@ void startBuilding(BuildTask buildTask) throws CoreException {
                 }
                 return buildResult;
             } catch (BuildDriverException e) {
```
matejonnet: What about adding Throwable to make sure we catch all ?

jbartece: Yeah, it could be a solution. But if someone make some updates the code and forgets to catch all exceptions, it can happen, but maybe I'm too careful. I'll catch Throwable.

matejonnet: We must pay attention what is merged in ;)

jbartece:
retest this please

PR #266, opened 2015-04-02
[NCL-666] Allow non-null ID for updates

pgier:
The id in the entity can be non-null as long as it matches the id provided
in the URL.
Also add update operation for User entity.


slaskawi:
:+1: Looks good

PR #267, opened 2015-04-02
Finalized view of BuildConfigurationSet; fixed deletion of BuildConfigur...

vibe13:
...ationSet with return to previous state;


vibe13:
test this please

PR #270, opened 2015-04-03
NCL-566 WebSockets for REST

slaskawi:


pnc-rest/src/main/java/org/jboss/pnc/rest/debug/TestEndpoint.java
```diff
@@ -0,0 +1,32 @@
...
+@ApplicationScoped
+public class TestEndpoint {
+
+    @Inject
+    Event<BuildStatusChangedEvent> buildStatusChangedEventEvent;
+
+    @POST
+    @Path("/buildstatus/notify")
+    @ApiOperation(value = "Sends BuildStatusChangedEvent just like it was from Core, useful for testing WebSockets")
+    public void sendBuildStatusChangedEvent(BuildStatusChangedEventRest buildStatusChangedEventRest) {
```
slaskawi: @alexcreasy @vibe13 This is the most convenient way to test web sockets. It simulates changing build status from core.

test-arquillian-container/pom.xml
```diff
@@ -197,6 +201,45 @@
...
+            <executions>
+              <execution>
+                <id>start-server</id>
+                <phase>generate-test-resources</phase>
+                <goals>
+                  <goal>start</goal>
+                </goals>
+                <configuration>
+                  <!-- It looks like JBoss Maven plugin needs absolute paths -->
+                  <jbossHome>${project.basedir}/${test.server.build.dir}</jbossHome>
```
slaskawi: @rnc @pgier Unfortunately I had to make this path absolute. This is weird - wildfly plugin seems to work with both absolute and relative paths.

test-arquillian-container/pom.xml
```diff
@@ -197,6 +201,45 @@
...
+                <id>add-nio</id>
+                <phase>generate-test-resources</phase>
+                <goals>
+                  <goal>execute-commands</goal>
+                </goals>
+                <configuration>
+                  <execute-commands>
+                    <batch>true</batch>
+                    <commands>
+                      <command>/subsystem=web/connector=http/:write-attribute(name=protocol,value=org.apache.coyote.http11.Http11NioProtocol)</command>
```
slaskawi: @rnc Added NIO with invoking commands. Nice trick, thanks!

PR #274, opened 2015-04-07
Pnc keycloak

pslegr:


pslegr:
rebasing

integration-test/src/test/java/org/jboss/pnc/integration/BuildRecordRestTest.java
```diff
@@ -64,7 +72,17 @@ public static EnterpriseArchive deploy() {
     @Test
     @InSequence(-1)
     public void prepareBaseData() {
-        Response response = given().contentType(ContentType.JSON).port(getHttpPort()).when().get(BUILD_RECORD_REST_ENDPOINT);
+        try {
+            InputStream is = this.getClass().getResourceAsStream("/keycloak.json");
+            ExternalAuthentication ea = new ExternalAuthentication(is);
+            authProvider = ea.authenticate(System.getenv("PNC_EXT_OAUTH_USERNAME"), System.getenv("PNC_EXT_OAUTH_PASSWORD"));
```
matejonnet: Can you use configuration file instead of env variables?

auth/src/main/java/org/jboss/pnc/auth/ExternalAuthFacade.java
```diff
@@ -0,0 +1,103 @@
...
+import org.jboss.logging.Logger;
+import org.keycloak.adapters.HttpClientBuilder;
+
+/**
+ * @author pslegr
+ *
+ */
+public class ExternalAuthFacade {
+    public final static Logger log = Logger.getLogger(ExternalAuthentication.class);
+
```
matejonnet: Any particular reason to use "different" class for logger name?

integration-test/src/test/java/org/jboss/pnc/integration/BuildRecordRestTest.java
pslegr: On 7.4.2015 10:38, Matej Lazar wrote:
>
> In 
> integration-test/src/test/java/org/jboss/pnc/integration/BuildRecordRestTest.java 
> <https://github.com/project-ncl/pnc/pull/274#discussion_r27860248>:
>
> > @@ -64,7 +72,17 @@ public static EnterpriseArchive deploy() {
> >      @Test
> >      @InSequence(-1)
> >      public void prepareBaseData() {
> > -        Response response = given().contentType(ContentType.JSON).port(getHttpPort()).when().get(BUILD_RECORD_REST_ENDPOINT);
> > +        try {
> > +            InputStream is = this.getClass().getResourceAsStream("/keycloak.json");
> > +            ExternalAuthentication ea = new ExternalAuthentication(is);
> > +            authProvider = ea.authenticate(System.getenv("PNC_EXT_OAUTH_USERNAME"), System.getenv("PNC_EXT_OAUTH_PASSWORD"));
>
> Can you use configuration file instead of env variables?
>
Ok, will add a new module into pnc-config.json and use config properties

PR #276, opened 2015-04-07
[NCL-594]: Create non blocking wait for jenkins to boot up

jbartece:


docker-environment-driver/src/main/java/org/jboss/pnc/environment/docker/DockerInitializationMonitor.java
```diff
@@ -0,0 +1,90 @@
...
+                    onMonitorComplete.run();
+                } else {
+                    if (waiting >= MAX_CONTAINER_LOADING_TIME)
+                        throw new EnvironmentDriverException(
+                                "Jenkins server in container was not fully up and running in: "
+                                        + MAX_CONTAINER_LOADING_TIME + " seconds");
+                }
+            } catch (Exception e) {
+                onMonitorError.accept(e);
+                futureReference.get().cancel(true);
```
matejonnet: Probably it is safer to cancel the service before any other operation, just in case the other thrown an exception too.

jenkins-build-driver/src/main/java/org/jboss/pnc/jenkinsbuilddriver/JenkinsBuildMonitor.java
```diff
@@ -78,6 +78,7 @@ public void monitor(String jobName, int buildNumber, Consumer<BuildDriverStatus>
                 }
             } catch (Exception e) {
                 onMonitorError.accept(e);
+                futureReference.get().cancel(true);
```
matejonnet: Cancel before error notify?

pnc-core/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
```diff
@@ -322,16 +342,16 @@ private void stopRunningEnvironment(Throwable ex) {
         if(ex instanceof BuildProcessException) {
             BuildProcessException bpEx = (BuildProcessException) ex;
             try {
-                if (bpEx.getRunningEnvironment() != null)
-                    bpEx.getRunningEnvironment().destroyEnvironment();
+                if (bpEx.getDestroyableEnvironmnet() != null)
+                    bpEx.getDestroyableEnvironmnet().destroyEnvironment();
             } catch (EnvironmentDriverException envE) {
-                log.warn("Running environment" + bpEx.getRunningEnvironment() + " couldn't be destroyed!", envE);
+                log.warn("Running environment" + bpEx.getDestroyableEnvironmnet() + " couldn't be destroyed!", envE);
```
matejonnet: Error level instead of warn ?

docker-environment-driver/src/main/java/org/jboss/pnc/environment/docker/DockerInitializationMonitor.java
jbartece: OK

jenkins-build-driver/src/main/java/org/jboss/pnc/jenkinsbuilddriver/JenkinsBuildMonitor.java
jbartece: OK

pnc-core/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
jbartece: I think it is warning, because it causes leaks, but the application can run good in user's point of view.

matejonnet: Ok.

mareknovotny:
after all of these changes, docker remote test causes 3 running containers are left on docker host. See NCL-737

PR #278, opened 2015-04-08
NCL-649 Added Audit information to BuildConfiguration

slaskawi:


datastore/src/main/java/org/jboss/pnc/datastore/repositories/audited/AuditedBuildConfigurationRepository.java
```diff
@@ -0,0 +1,23 @@
...
+import org.jboss.pnc.datastore.audit.impl.AbstractAuditRepository;
+import org.jboss.pnc.model.BuildConfiguration;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+@ApplicationScoped
+public class AuditedBuildConfigurationRepository extends AbstractAuditRepository<BuildConfiguration, Integer> {
+
+    @Deprecated
```
pgier: Why is this marked deprecated?  What should it be replaced by?

slaskawi: CDI needs to have a default constructor for proxing. It shouldn't be 
used directly.

When we use @Deprecated annotation - your IDE will indicate that you 
shouldn't use it.

pnc-model/src/main/java/org/jboss/pnc/model/Environment.java
```diff
@@ -1,14 +1,15 @@
...
 
-import java.io.Serializable;
+import javax.persistence.*;
 
 /**
  * The Class Environment.
  */
 @Entity
-public class Environment implements Serializable {
+@Audited
```
pgier: I'm wondering if environment should be immutable instead of auditing it.

slaskawi: I assumed that for reproducing build we need a configuration plus an 
environment. That's why I added @Audited to the Environment.

But I totally agree - we may consider doing it immutable.

datastore/src/main/java/org/jboss/pnc/datastore/repositories/audited/AuditedBuildConfigurationRepository.java
pgier: Deprecated makes it sound like it will be removed eventually, but I guess there isn't a better tag to use to say that it just shouldn't be used as public api.  Can you add a javadoc comment to this constructor?

pgier:
We also need to link the build record table to the config audit table, and then remove the redundant fields from build record.  I guess that could be done in a separate commit or PR though.

datastore/src/main/java/org/jboss/pnc/datastore/repositories/audited/AuditedBuildConfigurationRepository.java
slaskawi: Ok, I'll implement it tomorrow.

slaskawi:
hmmm I'm not entirely sure if this is a good idea to put @Auditable on 
BuildRecord. This table will be huge in the future and adding auditing 
will double the numbers.

Are we sure we want to do that?

pgier:
We don't want to audit the build record, we need to link the build record to the build configuration audit table so that we don't have a bunch of duplicate information in the build record.

pnc-model/src/main/java/org/jboss/pnc/model/Environment.java
slaskawi: Ok, let's got for the easiest solution and leave audited there.

datastore/src/main/java/org/jboss/pnc/datastore/repositories/audited/AuditedBuildConfigurationRepository.java
slaskawi: done

PR #282, opened 2015-04-10
Ncl 702

pgier:
Update to for product milestones and releases.  The model and rest API changes seem to be working, but there will probably need to be some updates to the web UI.


integration-test/src/test/java/org/jboss/pnc/integration/BuildRecordSetRestTest.java
```diff
@@ -40,8 +40,7 @@
 
     public static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
-    private static final String PRODUCT_REST_ENDPOINT = "/pnc-rest/rest/product/";
-    private static final String PRODUCT_VERSION_REST_ENDPOINT = "/pnc-rest/rest/product/%d/version/";
+    private static final String PRODUCT_MILESTONE_REST_ENDPOINT = "/pnc-rest/rest/product-milestone/";
```
slaskawi: I'm afraid it may break UI. Am I right? 

@vibe13 @alexcreasy maybe you could take a look.

integration-test/src/test/java/org/jboss/pnc/integration/ProductReleaseRestTest.java
```diff
@@ -0,0 +1,176 @@
...
+        EnterpriseArchive enterpriseArchive = Deployments.baseEar();
+        logger.info(enterpriseArchive.toString(true));
+        return enterpriseArchive;
+    }
+
+    @Before
+    public void prepareData() {
+        try {
+            InputStream is = this.getClass().getResourceAsStream("/keycloak.json");
+            ExternalAuthentication ea = new ExternalAuthentication(is);
```
slaskawi: This will happen before each test. I think it should happen only once.

integration-test/src/test/java/org/jboss/pnc/integration/ProductReleaseRestTest.java
```diff
@@ -0,0 +1,176 @@
...
+            authProvider = ea.authenticate(System.getenv("PNC_EXT_OAUTH_USERNAME"), System.getenv("PNC_EXT_OAUTH_PASSWORD"));
+        } catch (IOException e) {
+            // TODO Auto-generated catch block
+            e.printStackTrace();
```
slaskawi: Just add a throws to method declaration

integration-test/src/test/java/org/jboss/pnc/integration/ProductReleaseRestTest.java
```diff
@@ -0,0 +1,176 @@
...
+                .post(PRODUCT_RELEASE_REST_ENDPOINT);
+        Assertions.assertThat(response.statusCode()).isEqualTo(201);
+
+        String location = response.getHeader("Location");
+        logger.info("Found location in Response header: " + location);
+
+        newProductReleaseId = Integer.valueOf(location.substring(location.lastIndexOf(
+                PRODUCT_RELEASE_REST_ENDPOINT) + PRODUCT_RELEASE_REST_ENDPOINT.length()));
+
+        logger.info("Created id of product version: " + newProductReleaseId);
```
slaskawi: Do we need an assertion here?

pnc-model/src/main/java/org/jboss/pnc/model/ProductRelease.java
```diff
@@ -0,0 +1,288 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2012, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
slaskawi: This header might be removed I guess

pnc-model/src/main/java/org/jboss/pnc/model/ProductRelease.java
```diff
@@ -0,0 +1,288 @@
...
+    @OneToOne(cascade = { CascadeType.PERSIST, CascadeType.MERGE, CascadeType.REFRESH })
+    @JoinColumn(name = "build_record_set_id")
+    private BuildRecordSet buildRecordSet;
+
+    public ProductRelease() {
+
+    }
+
+    /**
+     * @param version
```
slaskawi: Please add a meaningful Javadoc or remove it (I vote for the second option).

pnc-model/src/main/java/org/jboss/pnc/model/SupportLevel.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * JBoss, Home of Professional Open Source
+ * Copyright 2012, Red Hat, Inc., and individual contributors
+ * by the @authors tag. See the copyright.txt in the distribution for a
+ * full listing of individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
slaskawi: Header

pnc-model/src/main/java/org/jboss/pnc/model/SupportLevel.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.pnc.model;
+
+/**
+ * @author avibelli
```
slaskawi: Javadoc

pnc-model/src/main/java/org/jboss/pnc/model/VersionQualifier.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package org.jboss.pnc.model;
+
+/**
+ * @author avibelli
+ *
+ */
+public enum VersionQualifier {
```
slaskawi: The same here.

slaskawi:
Looks good to me, there are some small things to correct. 

There is also a thing which we can improve in some other PR. Since REST endpoints simply forwards calls to Provider layer, we could focus on testing that layer instead. This would speed up the tests.

integration-test/src/test/java/org/jboss/pnc/integration/ProductReleaseRestTest.java
pgier: You're right, I had just copied it from the existing product versions rest test.  Looks like the auth setup is kind of inconsistent between out test classes currently.  I'll can change it to "@BeforeClass" and see if that works.

integration-test/src/test/java/org/jboss/pnc/integration/ProductReleaseRestTest.java
pgier: Good point, this was again something I just copied from our existing tests without too much thought.  I'll review that in the other ITs as well.

integration-test/src/test/java/org/jboss/pnc/integration/ProductReleaseRestTest.java
slaskawi: @BeforeClass happens before Arquillian deployment. Take a look at other 
tests with AtomicBoolean trick or maybe you could come up with some 
better idea.

integration-test/src/test/java/org/jboss/pnc/integration/ProductReleaseRestTest.java
pgier: The important assertion is on line 124.  The remaining code is just parsing the ID from the URL and logging.  I guess we could assert that a valid Integer is found in the URL?

pgier:
Regarding the header and javadoc stuff, I'll review that today and get it cleaned up.  Thanks for the feedback!

integration-test/src/test/java/org/jboss/pnc/integration/ProductReleaseRestTest.java
pgier: The tests run successfully using "@BeforeClass".  I guess the authentication is just running against the server and doesn't care about the app?

PR #285, opened 2015-04-13
Noauth profile

pslegr:
Adding the ability to enable authentication as optional profile during build time


pslegr:
retest this please

PR #303, opened 2015-04-21
[NCL-764] Save the audit revision of the build config to the build recor...

pgier:
...d


pnc-core/src/main/java/org/jboss/pnc/core/builder/DatastoreAdapter.java
```diff
@@ -40,7 +40,7 @@ public void storeResult(BuildTask buildTask, BuildResult buildResult) throws Dat
             RepositoryManagerResult repositoryManagerResult = buildResult.getRepositoryManagerResult();
 
             BuildRecord buildRecord = new BuildRecord();
-            buildRecord.setLatestBuildConfiguration(buildConfiguration);
+            buildRecord.setBuildConfigurationAudited(buildConfiguration.getBuildConfigurationAudited());
             // Build driver results
```
matejonnet: Can be removed as it is done in setAuditDataToBuildRecord. I extracted it to method sa it is used also in case of storingResult with exception.

PR #306, opened 2015-04-22
NCL-746 Artifacts for Build Record Set

slaskawi:


slaskawi:
We need another approach.

PR #309, opened 2015-04-22
NCL-775 Use skipTests not maven.test.skip

rnc:


rnc:
retest this please

rnc:
retest this please

vibe13:
retest this please

vibe13:
retest this please

demo-data/src/main/java/org/jboss/pnc/demo/data/DatabaseDataInitializer.java
```diff
@@ -168,7 +168,7 @@ public void initiliazeProjectProductData() {
                     .project(project1)
                     .description("Test build config for project newcastle")
                     .environment(environment1)
-                    .buildScript("mvn clean deploy -Dmaven.test.skip")
+                    .buildScript("mvn clean deploy -DskipTests=true")
```
matejonnet: Isn't it better for faster build to skip also the compilation of tests and not just the execution ?

rnc: Perhaps for some projects eg pnc but for others e.g. fabric it breaks the build as later modules rely on test jars being built. I think building the tests but not running them is a reasonable compromise.

matejonnet: Looks like project (fabric) setup must be fixed, runtime code should not depend on test code.

rnc: Maybe but many community projects might be like this. We have to build what we are provided. Also we should probably deploy the same as them i.e. deploy the test artifacts.

matejonnet: Ok, then we can use that properly only for our project (pnc).

rnc: Timing a build on an internal server with skipTests versus maven.test.skip the difference appears fairly insignificant.

matejonnet: But using maven.test.skip will ensure we don't end up using test classes in runtime.

PR #310, opened 2015-04-22
[NCL-746] Get build records for configurations in a config set

pgier:


vibe13:
@slaskawi @matejonnet could the above failure be a problem with Docker space again?

pgier:
retest this please

PR #322, opened 2015-05-08
NCL-802 NCL-817 Product Milestones DB change

vibe13:


vibe13:
test this please

vibe13:
test this please

vibe13:
@pgier @alexcreasy @rnc  mind to review? this fixes also NCL-817

pgier:
I kind of prefer that "planned_starting_date" is just called "starting_date".  I don't think we have any reason to differentiate between planned and actual starting state.  We only care about the date we actually started working on the milestone.

vibe13:
@pgier ok thanks will change it

slaskawi:
retest this please

rnc:
I suppose if we were to have planned start date that might differ from our actual start date - which PM are more interested in. If (in future) we could provide that capability where they can compare the start/end dates compared with their planning that would be sufficient.

alexcreasy:
Yes this works for me!

PR #326, opened 2015-05-12
NCL 766 : Update versions during build

rnc:
@alexcreasy  @vibe13  : can you review please

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,6 +71,8 @@
     <version.jbpm>6.2.0.Final</version.jbpm>
     <version.keycloak>1.1.0.Final</version.keycloak>
 
+    <datetime>${timestamp}</datetime>
+
     <!-- maven-compiler-plugin -->
     <maven.compiler.target>1.8</maven.compiler.target>
     <maven.compiler.source>1.8</maven.compiler.source>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -1230,7 +1278,57 @@
...
+
+                                <property name="newversion" value="${project.rel.org.jboss.pnc:parent}"/>
+
+                                <replaceregexp file="${basedir}/bower.json"
+                                               match="version&quot;: &quot;[0-9].*&quot;"
+                                               replace="version&quot;: &quot;${newversion}&quot;"/>
+                                <replaceregexp file="${basedir}/package.json"
+                                               match="version&quot;: &quot;[0-9].*&quot;"
+                                               replace="version&quot;: &quot;${newversion}&quot;"/>
+                                <replaceregexp file="${basedir}/npm-shrinkwrap.json"
```
vibe13: @rnc These replacements (L1316-1324) scare me a bit, there are tons of versions inside those files, and versions of resolved dependencies could potentially match the PNC version. But seems that only npm-shrinkwrap.json file has many  "version": "X.Y.Z" declarations, so package.json and bower.json should be fine. We could remove the npm-shrinkwrap.json file from the replacement or enforce a more strict replacement to be sure that only the pnc-ui version gets replaced.

vibe13: All the remaining bits looks good!

vibe13:
Looks very good now! +1 also for the JBOSS_HOME env variable check!

alexcreasy:
Yes this looks good to me.

PR #333, opened 2015-05-18
Notify on build task completion.

matejonnet:


pnc-core/src/main/java/org/jboss/pnc/core/builder/BuildSetTask.java
```diff
@@ -33,25 +34,43 @@
...
 
-    private BuildStatus status;
+    private BuildSetStatus status = BuildSetStatus.NEW;
 
     private String statusDescription;
     private Set<BuildTask> buildTasks = new HashSet<>();
+    private Consumer<BuildSetStatus> onComplete;
 
-    public BuildSetTask(BuildConfigurationSet buildConfigurationSet, BuildExecutionType buildTaskType) {
+    public BuildSetTask(BuildConfigurationSet buildConfigurationSet, BuildExecutionType buildTaskType, Consumer<BuildSetStatus> onComplete) {
```
slaskawi: Why we are not using CDI events for this?

matejonnet: We must keep state associated to task life cycle, using CDI events we must define task scope to keep the state (what to do on event completion) in REST api.

matejonnet:
Retest this please.

PR #334, opened 2015-05-18
NCL-831 Remove copyright from include files. Update excludes

rnc:
@alexcreasy - updated for Angular include issues.


alexcreasy:
@rnc yes this fixes NCL-831, I'll merge now.

PR #358, opened 2015-05-27
Persistence updates

thauser:
A few more changes related to the maven module refactor. The important one is the change in the persistence.xml, which was failing to find the @Entity annotated classes due to an outdated model.jar reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ The default build with command `mvn clean install` comes with no authentication.
 use -Dauth=true together with your build command.
 Enabling authentication meand following
 1. Your backend REST endpoints will become secured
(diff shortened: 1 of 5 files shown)


thauser:
There's still more to change. Rest Endpoints are broken with this refactor, for instance.

thauser:
It's not clear to me if we want the modules to produce pnc-* artifacts still, or if things like having 'models.jar' instead of 'pnc-models.jar' are desired. Regardless, with these changes we're able to build and use basic functionality again.

pslegr:
We need to keep the names consistent, so your changes are OK ! thanks ! merging

PR #359, opened 2015-05-29
Ui validation refactoring

vibe13:


alexcreasy:
milestone-controllers.js line: 61... the jshint override shouldn't be used ..... the "result" param can just be removed from the function call on the line above i.e. function() { Notifications.success('Milestone Created'); ....

alexcreasy:
Also I think milestone and release should be submodules of version (which makes me think at some point we should plan to refactor our module names to reflect the model updates)

alexcreasy:
@vibe13 OK actually as this contains verified bugfixes and code improvements that will leave master in a better position than before, I'm going to merge this and then I'll raise a new PR for further improvements.

vibe13:
@alexcreasy thanks!

PR #361, opened 2015-05-29
Added close milestone feature

vibe13:


alexcreasy:
I think it would be better to remove the lock and unlock .png files and use the font-awesome icons included with patternfly all you have to do is use: 
Lock
#######
```HTML
<i class="fa fa-lock"></i>
```
Unlock 
#######
```HTML
<i class="fa-unlock-alt"></i>
```
a) These are vector graphics and can be scaled to any size
b) No new files for user to download so marginally faster
c) These are recommended by UX team for use with Patternfly look and feel (see: https://www.patternfly.org/styles/icons/)

vibe13:
@alexcreasy  oooh thanks, I found the lock but missed the unlock one, this is why I added those....thanks!

vibe13:
retest this please

PR #362, opened 2015-05-29
BPM Callback with BASIC Auth

sbunciak:


pslegr:
Merging this one & we will need to add remote test for whole BPM process => NCL-866

rest/src/main/java/org/jboss/pnc/rest/trigger/BuildTriggerer.java
```diff
@@ -144,4 +170,56 @@ private BuildConfigurationAudited getLatestAuditedBuildConfiguration(Integer bui
         }
         return buildConfigRevs.get(0);
     }
+    
+    /*
+     * TODO: Do not ignore certificates, rather setup servers properly.
+     */
+    private void signalBpmEvent(String uri) {
```
matejonnet: What about extracting "http client" to its own class to make it re-usable?

rest/src/main/java/org/jboss/pnc/rest/trigger/BuildTriggerer.java
```diff
@@ -144,4 +170,56 @@ private BuildConfigurationAudited getLatestAuditedBuildConfiguration(Integer bui
...
+        HttpPost request = new HttpPost(uri);
+        request.addHeader("Authorization", getAuthHeader());
+        log.info("Executing request " + request.getRequestLine());
+
+        try (CloseableHttpResponse response = httpclient.execute(request)) {
+            log.info(response.getStatusLine());
+            response.close();
+            httpclient.close();
+        } catch (IOException e) {
+            e.printStackTrace();
```
matejonnet: No actions is something goes wrong? Should we at least re-try few times?
Use log instead of printStackTrace.

rest/src/main/java/org/jboss/pnc/rest/trigger/BuildTriggerer.java
sbunciak: +1 will be done as part of NCL-866 as we will need it for the BPM integration test.

rest/src/main/java/org/jboss/pnc/rest/trigger/BuildTriggerer.java
sbunciak: The retries could be implemented in the extracted httpClient class as an optional parameter I guess. WDYT?
Sorry, my fault, I will fix the printStackTrace.

rest/src/main/java/org/jboss/pnc/rest/trigger/BuildTriggerer.java
sbunciak: Found out that we already have https://github.com/project-ncl/pnc/blob/master/common/src/main/java/org/jboss/pnc/common/util/HttpUtils.java. I suppose it would be better to reuse this class rather than creating a new one.

rest/src/main/java/org/jboss/pnc/rest/trigger/BuildTriggerer.java
matejonnet: +1 for retry in client.

rest/src/main/java/org/jboss/pnc/rest/trigger/BuildTriggerer.java
matejonnet: Agree to put it in common module. If it is not just few lines of code i would prefer having a client in its own classes.

PR #364, opened 2015-05-29
Fixed REST endpoint proxy config in Web UI.

jsenko:
REST proxy configuration in `rest-config.json` does not work for me when using `grunt serve` with custom `--target`, as described in `ui/README.md`

For example:
Let `rest-config.json` be
```
{
  "endpointsLocalhost": "localhost",
  "dev": "1.2.3.4"
}
```
Command `grunt serve --target=dev` creates proxy for `localhost` instead of `1.2.3.4`.
This fix should resolve the issue.

### DIFF
--- a/ui/Gruntfile.js
+++ b/ui/Gruntfile.js
@@ -44,21 +44,42 @@ module.exports = function (grunt) {
     proxyHost: PROXY_HOST
   };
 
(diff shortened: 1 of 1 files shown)


jsenko:
@alexcreasy can you verify?

PR #369, opened 2015-06-04
[NCL-471]: Rest get build record log returns 204 when it is empty

jbartece:


rest/src/test/java/org/jboss/pnc/rest/endpoint/BuildRecordEndpointTest.java
```diff
@@ -0,0 +1,50 @@
...
+        int logId = 1;
+        String logContent = "LOG CONTENT";
+        BuildRecordEndpoint buildRecordEndpoint = getLogsPrepareEndpoint(logId, logContent);
+
+        // then
+        assertEquals(200, buildRecordEndpoint.getLogs(logId).getStatus());
+    }
+
+    private BuildRecordEndpoint getLogsPrepareEndpoint(int logId, String logContent) {
+        // given
```
slaskawi: //given to be removed

rest/src/test/java/org/jboss/pnc/rest/endpoint/BuildRecordEndpointTest.java
```diff
@@ -0,0 +1,50 @@
...
+        BuildRecordRepository buildRecordRepository = mock(BuildRecordRepository.class);
+        BuildRecordProvider buildRecordProvider = new BuildRecordProvider(buildRecordRepository, null);
+        BuildRecordEndpoint buildRecordEndpoint = new BuildRecordEndpoint(buildRecordProvider, null);
+        BuildRecord buildRecord = mock(BuildRecord.class);
+
+        // when
```
slaskawi: //when to be removed

rest/src/test/java/org/jboss/pnc/rest/endpoint/BuildRecordEndpointTest.java
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * 
+ * @author Jakub Bartecek <jbartece@redhat.com>
+ *
+ */
+public class BuildRecordEndpointTest {
+
+    @Test
+    public void getLogsNoContentTest() {
+        int logId = 1;
```
slaskawi: Please add //given //when

rest/src/test/java/org/jboss/pnc/rest/endpoint/BuildRecordEndpointTest.java
```diff
@@ -0,0 +1,50 @@
...
+        BuildRecordEndpoint buildRecordEndpoint = getLogsPrepareEndpoint(logId, "");
+
+        // then
+        assertEquals(204, buildRecordEndpoint.getLogs(logId).getStatus());
+    }
+
+    @Test
+    public void getLogsWithContentTest() {
+        int logId = 1;
```
slaskawi: The same here

rest/src/test/java/org/jboss/pnc/rest/endpoint/BuildRecordEndpointTest.java
```diff
@@ -0,0 +1,50 @@
...
+import static org.mockito.Mockito.mock;
+
+import org.jboss.pnc.datastore.repositories.BuildRecordRepository;
+import org.jboss.pnc.model.BuildRecord;
+import org.jboss.pnc.rest.provider.BuildRecordProvider;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+/**
+ *
```
slaskawi: Add a valid Javadoc or remove it.

rest/src/main/java/org/jboss/pnc/rest/provider/BuildRecordProvider.java
```diff
@@ -119,17 +121,25 @@ public BuildRecordRest getSpecific(Integer id) {
         }
         return null;
     }
+    
 
-    public StreamingOutput getLogsForBuildId(Integer id) {
+    public String getBuildRecordLog(Integer id) {
```
slaskawi: Maybe we should remove this method and preserve only the streaming one?

jbartece: It is needed to find out if the content is empty or not. If the content is not empty the StreamingOutput is created after that

rest/src/test/java/org/jboss/pnc/rest/endpoint/BuildRecordEndpointTest.java
jbartece: Where should I add it, when I outlined it to the extra method?

rest/src/main/java/org/jboss/pnc/rest/provider/BuildRecordProvider.java
slaskawi: Ok, that makes sense.

rest/src/test/java/org/jboss/pnc/rest/endpoint/BuildRecordEndpointTest.java
slaskawi: //given //when //then should be added only to the test method's body. This blog post describes it really well: http://monkeyisland.pl/2009/12/07/given-when-then-forever/

jbartece: I have two choices:
  * put it to the method body, but there will be duplication of code in two tests
  * or outline it

So, are you for the first option?

slaskawi: I would prefer if you remove it from here and implement the test like the following:
```
@Test
    public void getLogsWithContentTest() {
        //given
        int logId = 1;
        String logContent = "LOG CONTENT";

        //when
        BuildRecordEndpoint buildRecordEndpoint = getLogsPrepareEndpoint(logId, logContent);

        // then
        assertEquals(200, buildRecordEndpoint.getLogs(logId).getStatus());
    }
```

jbartece: OK

PR #371, opened 2015-06-04
[NCL-473]: Rest get running build log returns 204 when it is empty

jbartece:


rest/src/main/java/org/jboss/pnc/rest/endpoint/RunningBuildRecordEndpoint.java
```diff
@@ -66,10 +77,22 @@ public BuildRecordRest getSpecific(@ApiParam(value = "BuildRecord id", required
...
+            @ApiResponse(code = 200, message = "Successful retrieval of the log"),
+            @ApiResponse(code = 204, message = "RunningBuild exists, but the log content is empty"),
+            @ApiResponse(code = 404, message = "RunningBuild with specified id does not exist"),
+    })
     @ApiOperation(value = "Gets specific log of a Running Build Record")
     @GET
     @Path("/{id}/log")
-    public Response getLogs(@ApiParam(value = "BuildRecord id", required = true) @PathParam("id") Integer id) {
-        return Response.ok(buildRecordProvider.getLogsForRunningBuildId(id)).build();
+    public Response getLogs(@ApiParam(value = "RunningBuild id", required = true) @PathParam("id") Integer id) {
```
slaskawi: Similar situation here - shouldn't we just return null?

slaskawi: ooh yes I see - we also have 404 for not found. Please ignore my previous comment.

slaskawi:
Looks good to me :+1:

jbartece:
retest this please

PR #372, opened 2015-06-04
[NCL-472]: Rest get all RunningBuildRecords returns 204 when no is av…

jbartece:
…ailable

### DIFF
--- a/rest/src/main/java/org/jboss/pnc/rest/endpoint/RunningBuildRecordEndpoint.java
+++ b/rest/src/main/java/org/jboss/pnc/rest/endpoint/RunningBuildRecordEndpoint.java
@@ -17,19 +17,30 @@
  */
 package org.jboss.pnc.rest.endpoint;
 
+import org.jboss.logging.Logger;
+import org.jboss.pnc.rest.provider.BuildRecordProvider;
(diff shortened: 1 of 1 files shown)


rest/src/main/java/org/jboss/pnc/rest/endpoint/RunningBuildRecordEndpoint.java
```diff
@@ -49,14 +60,22 @@ public RunningBuildRecordEndpoint(BuildRecordProvider buildRecordProvider) {
...
     }
 
+    @ApiResponses(value = {
+            @ApiResponse(code = 200, message = "Successful retrieval all RunningBuildRecords"),
+            @ApiResponse(code = 204, message = "No RunningBuildRecords available"),
+    })
     @ApiOperation(value = "Gets all running Build Records")
     @GET
-    public List<BuildRecordRest> getAll(
+    public Response getAll(
```
slaskawi: I'm not 100% but wouldn't returning a null do the job? This is mentioned in the spec as well as in StackOverflow: http://stackoverflow.com/questions/14102797/how-to-throw-http-204-status-code-using-jersey-framework-in-restful-web-service

jbartece:
retest this please

PR #373, opened 2015-06-04
NCL-881 - Create REST endpoint for retrieving ProductVersions associa…

alexcreasy:
…ted with a specific BuildConfiguration

### DIFF
--- a/rest/src/main/java/org/jboss/pnc/rest/provider/BuildConfigurationProvider.java
+++ b/rest/src/main/java/org/jboss/pnc/rest/provider/BuildConfigurationProvider.java
@@ -28,12 +28,14 @@ import org.jboss.pnc.model.BuildConfiguration;
 import org.jboss.pnc.model.BuildConfigurationSet;
 import org.jboss.pnc.model.ProductVersion;
 import org.jboss.pnc.rest.restmodel.BuildConfigurationRest;
+import org.jboss.pnc.rest.restmodel.ProductVersionRest;
 import org.springframework.data.domain.Pageable;
 
 import javax.ejb.Stateless;
 import javax.inject.Inject;
 
 import java.util.List;
+import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


alexcreasy:
Test this please

rest/src/main/java/org/jboss/pnc/rest/provider/BuildConfigurationProvider.java
```diff
@@ -173,4 +175,15 @@ public void removeProductVersion(Integer configId, Integer productVersionId) {
         buildConfigurationRepository.save(buildConfig);
     }
 
+    public List<ProductVersionRest> getProductVersions(Integer configId) {
+        BuildConfiguration buildConfig = buildConfigurationRepository.findOne(configId);
```
slaskawi: If we do it this way, we won't be able to use DB side logic, which is pretty bad. I would prefer if you used a proper repository for this and not a getter.

alexcreasy:
Hi Sebastian, I basically just copied the implementation of the /build-configuration-sets/:id/build-configurations please see BuildConfigurationSetProvider.java:144

I really need this functionality in the REST API urgently to be able to complete the tickets for the sprint. On the frontend we've had to implement some pretty horrible hacks (i.e. looping and making n REST calls where n is the array size) to retrieve an entities relations.

This particular endpoint I am stranded without, I can't even implement one of these horrible hacks.  Once retrieving a list of Configurations, it is impossible to query for it's associated ProductVersions, since the array returned contains only the ProductVersionId, yet you need a productId AND a ProductVersionId to query for one.

alexcreasy:
Is it possible we could merge this and my follow up PR (for NCL-882, coded in the same way) and I will look at rewriting these in the next sprint, including any that pre-existed that were coded in this fashion.

slaskawi:
Oh yeah, I asked Paul to rewrite this also (before my PTO). Obviously it seems it didn't happen.

Ok, if it's urgent this change is fine with me, but please create JIRA to fix it and assign it to me (or yourself if you'd like to do it :))

So, :+1: from me.

alexcreasy:
Great, thanks for understanding :)

PR #374, opened 2015-06-04
[NCL-845] Display error if Build Configuration is added to Build Configuration Set where it already exists

ahmedlawi92:
If the user attempts to add a configuration to a build configuration set that already contains the configuration the web ui will now display an error.

### DIFF
--- a/rest/src/main/java/org/jboss/pnc/rest/endpoint/BuildConfigurationSetEndpoint.java
+++ b/rest/src/main/java/org/jboss/pnc/rest/endpoint/BuildConfigurationSetEndpoint.java
@@ -186,6 +186,8 @@ public class BuildConfigurationSetEndpoint {
     public Response addConfiguration(
             @ApiParam(value = "Build Configuration Set id", required = true) @PathParam("id") Integer id,
             BuildConfigurationRest buildConfig) {
+        if (buildConfigurationSetProvider.getSpecific(id).getBuildConfigurationIds().contains(buildConfig.getId()))
+            return Response.status(Response.Status.CONFLICT).build();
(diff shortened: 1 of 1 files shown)


thescouser89:
Test this please

slaskawi:
test this please

slaskawi:
@ahmedlawi92 Are you joining PNC project for good?

rest/src/main/java/org/jboss/pnc/rest/endpoint/BuildConfigurationSetEndpoint.java
```diff
@@ -186,6 +186,8 @@ public Response build(
     public Response addConfiguration(
             @ApiParam(value = "Build Configuration Set id", required = true) @PathParam("id") Integer id,
             BuildConfigurationRest buildConfig) {
+        if (buildConfigurationSetProvider.getSpecific(id).getBuildConfigurationIds().contains(buildConfig.getId()))
```
slaskawi: Endpoints shouldn't contain any product related logic. I would prefer if you moved it to provider layer and add some tests for it.

ahmedlawi92:
@slaskawi  I'm actually only with Redhat for around 2 more months, but I'll try to contribute what I can while still here.

rest/src/main/java/org/jboss/pnc/rest/endpoint/BuildConfigurationSetEndpoint.java
ahmedlawi92: Right I'll move this to the provider and look to write some tests.

ahmedlawi92:
I made the PR from my master by accident so I'll close this and open a new PR.

PR #382, opened 2015-06-08
Return 409(Conflict) if adding BuildConfiguration to BuildConfigurationSet where the BuildConfiguration already exists.

ahmedlawi92:
Return error status if BuildConfiguration already exists in BuildConfigurationSet so that the UI will display an error.


slaskawi:
test this please

slaskawi:
Changes introduced here: https://github.com/project-ncl/pnc/pull/389

PR #385, opened 2015-06-09
NCL-760 Fix for: The pnc ui shows briefly before an unauthenticated u…

alexcreasy:
…ser is redirected to Keycloak login page


alexcreasy:
Test this please

PR #392, opened 2015-06-14
[NCL-904] Add rest endpoints for getting the audited build configuration

pgier:


rest/src/main/java/org/jboss/pnc/rest/provider/BuildRecordProvider.java
```diff
@@ -179,6 +183,15 @@ public String getSubmittedBuildLog(Integer id) {
             return null;
     }
 
+    public BuildConfigurationAuditedRest getBuildConfigurationAudited(Integer id) {
+        BuildRecord buildRecord = buildRecordRepository.queryById(id);
+        BuildConfigurationAudited buildConfigurationAudited = buildRecord.getBuildConfigurationAudited();
```
vibe13: @pgier  maybe worth checking if buildRecord exists to prevent NPE?

PR #398, opened 2015-06-17
Added tabs directive to clean up product version page.

jsenko:


vibe13:
@jsenko I like the new directive for the tabs, it surely helps to keep code cleaner and reusable (surely for BuildConfigurations and BuildConfigurationSets that are inside many pages).

But I am not completely sure about the tabs for Milestones, Releases, BuildConfigurations and BuildConfigurationSets. IMO some information should be well visible and not via tabs (i.e. Milestone and Releases should be next to each other to let user see and compare dates). Moreover, I feel that there would be too many clicks around the views (many clicks to get to the ProdVersion details, and others to get the additional info).
I reckon that there are many info in the ProdVersion detail, but I would rather have a minimize button to hide i.e. Milestone + Releases (kept as they were before this PR), and hide BuildConfig and BuildConfigSet, so that the page is less "full" of displayed data. Also, tab IMO should group similar data (i.e. BuildConfig and BuildConfigSet), but not info that don't quite fit (i.e. Milestones and BuildConfig). This is just my opinion. @alexcreasy  wdyt?

alexcreasy:
@vibe13 @jsenko  The existing layout has undergone review from the whole team at mocking stage and, while it's certainly not perfect, to change it we would need to reopen that review process.

I really like the tabs directive, it clearly reduces code duplication, and I think we should work it in at some point. Thanks for showing initiative on this and please don't feel discouraged from trying new things.

I'm going to close this PR with a view to revisiting this after the 0.5 release.

PR #401, opened 2015-06-18
[NCL-922] Remove authService rootScope binding

kelnar:


vibe13:
@kelnar Good call for having a controller that wraps the authentication service, UI should go via the controller and not access the service directly as done here. @alexcreasy wdyt? For me +1 to merge

alexcreasy:
@vibe13 yes I agree completely, nothing should be bound directly to the rootScope and certainly the service should be encapsulated within the controller.

@kelnar excellent work, thanks for this. +1 for merge.

PR #403, opened 2015-06-19
Modifications to the EnvironmentEndpoint and EnvironmentRest classes

thauser:
These changes correct errors and inconsistencies within api-docs/environments
1) Correctly define the values within buildType and operationalSystem
2) Define the model used as Environment instead of EnvironmentRest, consistent with the other Endpoints
3) Correct small misnomers in the delete method


slaskawi:
test this please

PR #404, opened 2015-06-19
Add @ApiModelProperty to the BuildStatus property in BuildConfigurationRest

thauser:
This change allows correct resolution of the BuildStatus type:

before change: 
http://pastebin.com/ywPzeFp6

after change:
http://pastebin.com/eBTyMfWE

Note that "$ref": BuildStatus becomes "type": "string" 

pnc-rest/rest/api-docs/build-configurations/

which I believe is desired here.


slaskawi:
test this please

PR #409, opened 2015-06-22
Ncl 778

pgier:


rest/src/main/java/org/jboss/pnc/rest/provider/BuildConfigurationProvider.java
```diff
@@ -255,10 +264,41 @@ public BuildConfigurationAuditedRest getRevision(Integer id, Integer rev) {
         return new BuildConfigurationAuditedRest (auditedBuildConfig);
     }
 
+    public Response getLatestBuildRecord(Integer configId) {
+        BuildConfiguration buildConfiguration = buildConfigurationRepository.queryById(configId);
+        if (buildConfiguration == null) {
+            return Response.status(Response.Status.NOT_FOUND).entity("No build configuration exists with id: " + configId).build();
+        }
+        List<BuildRecord> buildRecords = buildRecordRepository.findAllByLatestBuildConfigurationOrderByEndTimeDesc(buildConfiguration);
```
slaskawi: I would recommend using https://github.com/project-ncl/pnc/blob/master/spi/src/main/java/org/jboss/pnc/spi/datastore/repositories/SortInfoProducer.java#L17 with proper arguments. It would look simpler and you will have to write less code.

rest/src/main/java/org/jboss/pnc/rest/provider/BuildConfigurationProvider.java
```diff
@@ -255,10 +265,43 @@ public BuildConfigurationAuditedRest getRevision(Integer id, Integer rev) {
...
     }
 
+    public Response getLatestBuildRecord(Integer configId) {
+        BuildConfiguration buildConfiguration = buildConfigurationRepository.queryById(configId);
+        if (buildConfiguration == null) {
+            return Response.status(Response.Status.NOT_FOUND).entity("No build configuration exists with id: " + configId).build();
+        }
+        PageInfo pageInfo = this.pageInfoProducer.getPageInfo(0, 1);
+        SortInfo sortInfo = this.sortInfoProducer.getSortInfo(SortingDirection.DESC, "endTime");
+        List<BuildRecord> buildRecords = buildRecordRepository.queryWithPredicates(pageInfo, sortInfo, withBuildConfigurationId(configId));
```
slaskawi: You could also use this to make it even simpler: https://github.com/project-ncl/pnc/blob/master/spi/src/main/java/org/jboss/pnc/spi/datastore/repositories/api/ReadOnlyRepository.java#L12

pgier:
retest this please

rest/src/main/java/org/jboss/pnc/rest/provider/BuildConfigurationProvider.java
pgier: I'm not sure I understand?  How would I get the latest record (sort) using only the predicates?

slaskawi: ooh sorry you are completely right. Please ignore my previous comment.

PR #425, opened 2015-06-30
NCL-971 NCL-966 Build Status Change notifications over WebSocket

slaskawi:


build-coordinator/src/main/java/org/jboss/pnc/core/builder/DatastoreAdapter.java
```diff
@@ -48,13 +48,13 @@ public DatastoreAdapter(Datastore datastore) {
         this.datastore = datastore;
     }
 
-    public void storeResult(BuildTask buildTask, BuildResult buildResult) throws DatastoreException {
+    public BuildRecord storeResult(BuildTask buildTask, BuildResult buildResult, int buildRecordId) throws DatastoreException {
```
slaskawi: @matejonnet I'm enforcing a contract here... Build Record Id must be equal to Task ID.

rest/src/main/java/org/jboss/pnc/rest/notifications/model/NotificationFactory.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+@ApplicationScoped
+public class NotificationFactory {
+
+    private Map<BuildStatus, NotificationEventType> externalEvents = new HashMap<>();
+
+    public NotificationFactory() {
+        externalEvents.put(BuildStatus.REPO_SETTING_UP, NotificationEventType.BUILD_STARTED);
```
slaskawi: @matejonnet Could you please have a look at those? I think I used correct statuses - I'm interested only in 3 states - when the build starts, completes or fails.

integration-test/src/test/java/org/jboss/pnc/integration/WebSocketsNotificationTest.java
```diff
@@ -90,9 +90,9 @@ public void before() throws Exception {
     @Test
     public void shouldConnectToWebSockets() throws Exception {
         //given
-        BuildStatusChangedEvent buildStatusChangedEvent = new DefaultBuildStatusChangedEvent(BuildStatus.NEW, BuildStatus.BUILD_WAITING, 1, null);
+        BuildStatusChangedEvent buildStatusChangedEvent = new DefaultBuildStatusChangedEvent(BuildStatus.NEW, BuildStatus.BUILD_COMPLETED_SUCCESS, 1, 1);
         // How to mock BuildRecordProvider?
-        String expectedJsonResponse = "{}";//"{\"oldStatus\":\"NEW\",\"newStatus\":\"BUILD_WAITING\",\"buildTaskId\":1,\"buildExecution\":null}";
+        String expectedJsonResponse = "{\"payload\":{\"id\":1,\"eventType\":\"BUILD_COMPLETED\",\"userId\":1}}";
```
slaskawi: @alexcreasy Please take a look at final version of a build status notification.. Is it ok?

rest/src/main/java/org/jboss/pnc/rest/notifications/model/NotificationFactory.java
matejonnet: What about exposing all statues and let the client (UI) decide how to treat them ?

slaskawi: @alexcreasy The full list is here: https://github.com/project-ncl/pnc/blob/master/spi/src/main/java/org/jboss/pnc/spi/BuildStatus.java

Would you like to receive all of them in the UI or prefiltered to BUILD_STARTED, COMPLETED and FAILED?

build-coordinator/src/main/java/org/jboss/pnc/core/builder/DatastoreAdapter.java
matejonnet: What is the benefit of having a parameter vs. getting the id from task internally?

slaskawi: This way it's explicit and not hidden in implementation details.

matejonnet: +1

rest/src/main/java/org/jboss/pnc/rest/notifications/model/NotificationFactory.java
alexcreasy: Sorry I'm late responding to this: I suppose exposing all events is more flexible, it's not really any additional work on the UI end and saves further work if it turns out we need to use something?

slaskawi: @alexcreasy no worries...

I would like to avoid sending you the internal events. Mapping internal events to external acts like a safety buffer between Backend and UI. Without it we couldn't safely refactor Backend internals without breaking UI. Let's keep it separated for now. If that won't scale - we will change it...

alexcreasy: +10000000 Joshua Bloch would be proud sir. Yes I didn't think of that, we don't want implementation details to leak through our API. We should use an abstract set of events for the API that will form part of the API contract.

alexcreasy:
Great +1 to merge from me!

rest/src/main/java/org/jboss/pnc/rest/notifications/model/NotificationFactory.java
slaskawi: I just realized that those external events should be moved to SPI module :) I'll fix it tomorrow.

matejonnet: Some other "clients" might want so show more progress details. I find it useful to see that my task is eg. COLLECTING_RESULTS_* or that task failed due to BUILD_ENV_SETUP and not because of the BUILD_COMPLETED_WITH_ERROR. Or maybe "WAITING_FOR_DEPENDENCIES" tells that the particular component is still waiting to be scheduled. Don't you think we should show this to the users ?
And org.jboss.pnc.spi.BuildStatus is already a part of the SPI.

slaskawi: Yes, we can add a mapping for that if you wish (but on the other hand we 
don't have any requirements for this at the moment, so I'm not entirely 
convinced that we should) but as I mentioned before, I would like to 
preserve this safety buffer between Backend and UI.

matejonnet: I find it more confusing having another status for the same thing. We already have BuildStatus, BuildSetStatus and BuildStatusChangedEvent, BuildSetStatusChangedEvent as part of SPI and already used in the REST.
https://github.com/project-ncl/pnc/blob/33ebb8c8d7dd2594f878ac1f5d1ed0cdb3c73009/rest/src/main/java/org/jboss/pnc/rest/trigger/BuildTriggerer.java#L102

slaskawi: Ouch, in my opinion this BPM part should be refactored then :(

The same problem - when you just rename a constant in Backend the BPM 
will blow up. Moreover, since you are passing event name as a string in 
callback URL you won't get any error during compile time. It will blow 
up in runtime. I would like to avoid this kind of problems...

matejonnet: Well its part of SPI, you don't just rename stuff there. And it would be the same with the other status type.

slaskawi: Ok I need this PR for further work, so I'm merging it. Let's continue the discussion here.

sbunciak: The signal event name (for bpm) is intentionally passed in URL to avoid storing it in 2 places (PNC & BPM server). Thus, the entiner callback url construction is up to BPM server and PNC just tries to invoke what has been passed to it. If you believe it should be refactor, then please go ahead, I'm not against it.

In either way, this code will need to be amended at some point because we might need to report only final state of the build to BPM server. Currently, it reports other events, like repo creation If I'm not mistaken.

PR #431, opened 2015-07-02
NCL-968 Web Sockets for BuildSet notifications

slaskawi:


build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
```diff
@@ -105,8 +105,7 @@ public BuildTask build(BuildConfiguration buildConfiguration, User userTriggered
         BuildSetTask buildSetTask = new BuildSetTask(
                 this,
                 buildConfigurationSet,
-                BuildExecutionType.STANDALONE_BUILD,
-                () -> datastore.getNextBuildConfigSetRecordId());
+                BuildExecutionType.STANDALONE_BUILD, userTriggeredBuild, datastore.getNextBuildConfigSetRecordId());
```
matejonnet: Any particular reason for eager id fetching?

slaskawi: This way `BuildSetTask` doesn't know anything about this supplier (loose coupling). BTW it was eager also before (id was extracted during `BuildSetTask` construction). This is why it didn't make any sense to me...

matejonnet: +1 if it was called in constructor it doesn't make any sense

PR #432, opened 2015-07-05
Ncl 929

pgier:


test-arquillian-container/pom.xml
```diff
@@ -198,6 +205,38 @@
                 </configuration>
               </execution>
               <execution>
+                <id>deploy-hsql-jdbc-driver</id>
+                <phase>generate-test-resources</phase>
+                <goals>
+                  <goal>deploy-artifact</goal>
+                </goals>
+                <configuration>
+                  <groupId>org.hsqldb</groupId>
```
slaskawi: Why do we need it? Can't we use default Wildfly's DS?

pom.xml
```diff
@@ -658,6 +658,13 @@
         <version>${version.org.json}</version>
       </dependency>
 
+      <dependency>
+        <groupId>org.hsqldb</groupId>
```
slaskawi: The same question - what is the advantage of using explicit Database to Wildfly/EAP's default one?

test-arquillian-container/pom.xml
pgier: H2 does not support the data type "timestamp with time zone".

alexcreasy:
This PR causes failures for me when I try to deploy locally and also prevents from deploying on the autodeploy server, I'm reverting this PR for now.

PR #433, opened 2015-07-06
NCL-823 Build Agent routing

slaskawi:


termd-build-driver/src/main/java/org/jboss/pnc/rest/notifications/model/TermdBuildDriver.java
```diff
@@ -0,0 +1,98 @@
...
+    public RunningBuild startProjectBuild(BuildConfiguration buildConfiguration, final RunningEnvironment runningEnvironment)
+            throws BuildDriverException {
+
+        return new RunningBuild() {
+
+            @Override
+            public void monitor(Consumer<CompletedBuild> onComplete, Consumer<Exception> onError) {
+                onComplete.accept(new CompletedBuild() {
+                    @Override
+                    public BuildDriverStatus getCompleteStatus() {
```
slaskawi: @matejonnet I think we should remove it.. There is already a build status in BuildResult.

termd-build-driver/src/main/java/org/jboss/pnc/rest/notifications/model/TermdBuildDriver.java
```diff
@@ -0,0 +1,98 @@
...
+                    @Override
+                    public BuildDriverResult getBuildResult() throws BuildDriverException {
+                        return new BuildDriverResult() {
+                            @Override
+                            public String getBuildLog() throws BuildDriverException {
+                                return "Logs... logs everywhere...";
+                            }
+
+                            @Override
+                            public BuildDriverStatus getBuildDriverStatus() throws BuildDriverException {
```
slaskawi: This should not throw exception probably...

termd-build-driver/src/main/java/org/jboss/pnc/rest/notifications/model/TermdBuildDriver.java
```diff
@@ -0,0 +1,98 @@
...
+                                return BuildDriverStatus.SUCCESS;
+                            }
+
+                            @Override
+                            public RunningEnvironment getRunningEnvironment() {
```
slaskawi: Why are we returning running environment all over the place? I think it should be only returned by RunningBuild...

termd-build-driver/src/main/java/org/jboss/pnc/rest/notifications/model/TermdBuildDriver.java
matejonnet: Looks like it can be removed as it is not used.

PR #434, opened 2015-07-06
Add method to shutdown ExecutorService.

ahmedlawi92:
This will be helpful when running local builds, as the fact that there is no way to shutdown the executor causes the Java process to hang.

### DIFF
--- a/build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
+++ b/build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
@@ -49,10 +49,7 @@ import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.event.Event;
 import javax.inject.Inject;
 import java.util.*;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.Executor;
-import java.util.concurrent.Executors;
+import java.util.concurrent.*;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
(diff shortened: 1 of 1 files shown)


build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
```diff
@@ -378,4 +375,8 @@ private boolean isBuildAlreadySubmitted(BuildTask buildTask) {
     Event<BuildSetStatusChangedEvent> getBuildSetStatusChangedEventNotifier() {
         return buildSetStatusChangedEventNotifier;
     }
+
+    public void shutdownCoordinator(){
```
slaskawi: I would be good to invoke it somewhere... Maybe a `@Predestroy` method or something similar?

build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
```diff
@@ -378,4 +375,8 @@ private boolean isBuildAlreadySubmitted(BuildTask buildTask) {
...
+        executor.shutdown();
```
slaskawi: And you would probably need 'shutdownNow()'

build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
matejonnet: I would leave it without annotation to have only external control.

matejonnet:
Test this please.

slaskawi:
test this please

PR #438, opened 2015-07-09
NCL-823 Termd Build Driver basic implementation

slaskawi:


rest/src/main/java/org/jboss/pnc/termdbuilddriver/DefaultNotificationFactory.java
```diff
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jboss.pnc.rest.notifications.model;
+package org.jboss.pnc.termdbuilddriver;
```
matejonnet: Rename package to pncbuilddriver.

termd-build-driver/src/main/java/org/jboss/pnc/termdbuilddriver/commands/TermdCommandInvoker.java
```diff
@@ -0,0 +1,69 @@
...
+
+                String data = "{\"action\":\"read\",\"data\":\"" + command + "\\r\\n\"}";
+                termdTerminalConnection.sendAsBinary(ByteBuffer.wrap(data.getBytes()));
+                if(logConsumer.isPresent()) {
+                    termdTerminalConnection.addLineConsumer(logConsumer.get());
+                }
+                termdTerminalConnection.addLineConsumer(logLine -> {
+                    logs.append(logLine).append("\n");
+                    if(logLine.contains("% ")) {
+                        numberOfCommandsReceived.countDown();
```
matejonnet: Can you use events instead of parsing response log. See Client.connectStatusListenerClient & WEB_SOCKET_LISTENER_PATH = "/socket/process-status-updates"

termd-build-driver/src/main/java/org/jboss/pnc/termdbuilddriver/commands/TermdCommandInvoker.java
```diff
@@ -0,0 +1,69 @@
...
+                    }
+                });
+
+                numberOfCommandsReceived.await();
```
matejonnet: With events mentioned above you probably also don't need to block the thread.

termd-build-driver/src/main/java/org/jboss/pnc/termdbuilddriver/TermdBuildDriver.java
```diff
@@ -0,0 +1,102 @@
...
+    public RunningBuild startProjectBuild(BuildConfiguration buildConfiguration, final RunningEnvironment runningEnvironment)
+            throws BuildDriverException {
+
+        TermdRunningBuild termdRunningBuild = new TermdRunningBuild(runningEnvironment, buildConfiguration);
+        TermdCommandInvoker invoker = new TermdCommandInvoker(URI.create(runningEnvironment.getJenkinsUrl()));
+
+        checkoutSources(termdRunningBuild, invoker)
+                .thenCompose(returnedTermdRunningBuild -> invokeBuild(returnedTermdRunningBuild, invoker))
+                .thenCompose(returnedTermdRunningBuild -> completeBuild(returnedTermdRunningBuild, invoker));
+
```
matejonnet: You should probably catch exceptions from performCommand and mark build as failed when exceptions occur ?

termd-build-driver/src/main/java/org/jboss/pnc/termdbuilddriver/commands/TermdCommandInvoker.java
slaskawi: Fair point.... 

I think I actually will have to block at some point but that would be much better. Thanks for the hint.

spi/src/main/java/org/jboss/pnc/spi/environment/RunningEnvironment.java
```diff
@@ -51,4 +52,9 @@
      */
     RepositorySession getRepositorySession();
 
+    /**
+     * @return Returns a build directory.
+     */
+    Path getWorkingDirectory();
```
matejonnet: Do we need to care about workingDirectory? Wouldn't be better to work with relative paths?

termd-build-driver/src/main/java/org/jboss/pnc/termdbuilddriver/commands/TermdCommandInvoker.java
```diff
@@ -0,0 +1,103 @@
...
+                invokedCommands.add(invocatedCommandResult);
+                return invocatedCommandResult;
+            } catch (Exception e) {
+                throw new TermdCommandExecutionException("There was a problem when invoking command " + command, e);
+            }
+        });
+    }
+
+    protected String getLogsDirectory() {
+        return workingDirectory.toAbsolutePath().toString();
```
matejonnet: Log directory is not linked to working directory it is defined via agent boot command line. We should define it in setting on driver side too.

spi/src/main/java/org/jboss/pnc/spi/environment/RunningEnvironment.java
slaskawi: I think we should. Some build script may require absolute paths. Additionally, unit tests are invoked in temp directory.

termd-build-driver/src/main/java/org/jboss/pnc/termdbuilddriver/commands/TermdCommandInvoker.java
slaskawi: Already done: https://github.com/project-ncl/pnc/pull/443

PR #445, opened 2015-07-10
Small fixes to HTTP Client based on PR #362

sbunciak:


rest/src/main/java/org/jboss/pnc/rest/trigger/BuildTriggerer.java
```diff
@@ -178,39 +181,22 @@ private void signalBpmEvent(String uri) {
...
         HttpPost request = new HttpPost(uri);
         request.addHeader("Authorization", getAuthHeader());
         log.info("Executing request " + request.getRequestLine());
 
-        try (CloseableHttpResponse response = httpclient.execute(request)) {
+        try (CloseableHttpResponse response = httpClient.execute(request)) {
             log.info(response.getStatusLine());
             response.close();
-            httpclient.close();
+            httpClient.close();
```
slaskawi: I guess we should use try with resources also here right?

rest/src/main/java/org/jboss/pnc/rest/trigger/BuildTriggerer.java
```diff
@@ -178,39 +181,22 @@ private void signalBpmEvent(String uri) {
...
-        try {
-            sslSF = new SSLConnectionSocketFactory(builder.build(), SSLConnectionSocketFactory.ALLOW_ALL_HOSTNAME_VERIFIER);
-        } catch (KeyManagementException | NoSuchAlgorithmException e1) {
-            e1.printStackTrace();
-        }
 
-        CloseableHttpClient httpclient = HttpClients.custom().setSSLSocketFactory(sslSF)
-                .setHostnameVerifier(new AllowAllHostnameVerifier()).build();
+        /*
+         * TODO: Do not ignore certificates, rather setup servers properly.
```
slaskawi: This comment should be removed IMO

rest/src/main/java/org/jboss/pnc/rest/trigger/BuildTriggerer.java
sbunciak: @slaskawi you mean something like the following?
```java
    try (CloseableHttpClient httpClient = HttpUtils.getPermissiveHttpClient()) {
        try (CloseableHttpResponse response = httpClient.execute(request)) {
            log.info(response.getStatusLine());
        }
    } catch (IOException e) {
        log.error("Error occured executing the callback.", e);
    }
```

slaskawi: Yeah I think so... BTW it looks pretty awkward that a response needs calling close...

PR #448, opened 2015-07-13
NCL-966 NCL-968 Web Socket notifications expose SPI statuses

slaskawi:
@alexcreasy @matejonnet As we agreed I'm exposing SPI statuses in this PR. Please note, that I've also changed `BuildType` to `BuildStatus` (BuildType name didn't make sense in this context)


alexcreasy:
This gets a +1 from me!

PR #462, opened 2015-07-20
Make swagger work irrespective of the base url

thescouser89:
Right now when viewing Swagger documentation from a local deployment, the documentation will try to execute commands to ${demo.pnc.ip.address}, when what we would like to do is execute commands to localhost so that we do not mess up with the demo server, or suffer from CORS.

This commit attempts to fix this by setting the links relative to the base url.

### DIFF
--- a/rest/src/main/resources/swagger.properties
+++ b/rest/src/main/resources/swagger.properties
@@ -16,4 +16,4 @@
 # limitations under the License.
 #
 
-baseUrl=http://${demo.pnc.ip.address}/pnc-rest/rest
+baseUrl=/pnc-rest/rest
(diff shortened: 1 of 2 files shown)


thescouser89:
test me please

thescouser89:
There is another place that is still using the variable, but it's not related to swagger and I'm not sure if removing the variable will break things.

slaskawi:
Yes, there is a WebSocket tester. Since it has been already implemented in UI I think you can just remove it.

slaskawi:
Closing... added fix here: https://github.com/project-ncl/pnc/pull/470

PR #471, opened 2015-07-24
NCL-1082 Fixed Project validation bug

slaskawi:


thescouser89:
@slaskawi Sorry for commenting on a closed PR, but could you help me understand why `shouldUpdateBuildConfiguration` method in `BuildConfigurationRestTest.java` did not catch the validation bug? I've looked at it and I thought that this method would catch the bug, but apparently it didn't. I don't quite get why.

rest/src/main/java/org/jboss/pnc/rest/provider/ProjectProvider.java
```diff
@@ -103,7 +103,8 @@ public Integer update(Integer id, ProjectRest projectRest) throws ConflictedEntr
 
     private void validateBeforeSaving(ProjectRest projectRest) throws ConflictedEntryException {
         Project project = projectRepository.queryByPredicates(withProjectName(projectRest.getName()));
-        if(project != null) {
+        //don't validate against myself
```
slaskawi: @thescouser89 Probably because of this comment :D It was validating against itself but from DB.... argh... silly mistake..

PR #472, opened 2015-07-24
NCL-1094 use correct Endpoint to add version to product

ahmedlawi92:


ahmedlawi92:
test this please

slaskawi:
@alexcreasy Could you please take a look at it?

alexcreasy:
I'm not sure this REST endpoint for creating a ProductVersion is consistent with our existing REST endpoints.

To create a BuildConfiguration we use `POST /build-configurations` we don't use `POST /projects/addBuildConfiguration`.

It also leads to non-idiomatic code. For example in the UI you are creating a new product-version by calling `new PncRestClient.Product()` rather than `new PncRestClient.Version()`.

I'm actually deprecating the PncRestClient gradually and the underlying Product and ProductVersion models will be invoked directly, I can see it getting mroe confusing at this point.

@ahmedlawi92 @slaskawi - what do you think?

ahmedlawi92:
@alexcreasy Hey, I could see the creation being moved to the ProductVersionEndpoint, but I think that the creation of the ProductVersion and its being added the corresponding Product should be coupled together.

So something along the lines of POST /product-versions where you provide the ProductVersion details and the Product object.

I don't think it should be separated into two different operations such as POST /product-versions where you create the ProductVersion and POST /products/:id/product-versions to add the created ProductVersion to the Product.

alexcreasy:
@ahmedlawi92 Yes I completely agree, I was thinking `POST /product-versions` which should reject any request that doesn't contain a valid `productId` field. This is how the `/build-configurations` endpoints behave with regards to their parent Project.

slaskawi:
Backend code looks nice! Small question - do we need any integration tests around this functionality? Or this is not critical and we are ok without them?

slaskawi:
retest this please

ahmedlawi92:
@slaskawi Hey, I actually just updated the integration tests to test the addition of the product-version from the new endpoint. I'm not sure if its critical but there were already tests in place for the old endpoint. I don't think it really hurts to have these tests there.

slaskawi:
@ahmedlawi92 Ohh yes, sorry - you're absolutely right. :+1: from me.

PR #489, opened 2015-08-06
Fix triggerBuilds to return correct taskId

ruhan1:
I find /build-configurations/{id}/build callbackUrl does not work. After debugging into it, I find the culprit is in triggerBuilds which returns the configuration Id instead of taskId.


slaskawi:
test this please

slaskawi:
Integrated thanks!

PR #491, opened 2015-08-06
BuildCoordinator.shutdown() should also shutdown the dbexecutor

ahmedlawi92:
Need to shutdown all executors to end the java process.


pgier:
Rebased and merged this commit in another PR (#505)

PR #494, opened 2015-08-07
Ncl 1032

pslegr:
Decouple PNC module config's from common module. Common is able to be reused by other projects (BA atm)


pslegr:
retest this please

pslegr:
retest this please

moduleconfig/src/main/java/org/jboss/pnc/common/json/moduleprovider/PncConfigProvider.java
```diff
@@ -0,0 +1,82 @@
...
+import org.jboss.pnc.common.json.moduleconfig.BuildDriverRouterModuleConfig;
+import org.jboss.pnc.common.json.moduleconfig.DockerEnvironmentDriverModuleConfig;
+import org.jboss.pnc.common.json.moduleconfig.JenkinsBuildDriverModuleConfig;
+import org.jboss.pnc.common.json.moduleconfig.MavenRepoDriverModuleConfig;
+import org.jboss.pnc.common.json.moduleconfig.TermdBuildDriverModuleConfig;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.jsontype.NamedType;
+
+public class PncConfigProvider <T extends AbstractModuleConfig> implements ConfigProvider<T> {
```
matejonnet: Repeating the package name (pnc) in class name is considered as anti-pattern.

build-coordinator/src/test/java/org/jboss/pnc/core/BuildDriverFactoryTest.java
```diff
@@ -47,6 +49,7 @@ public void shouldSkipDriversWhichCanNotBuild() throws Exception {
         factory.getBuildDriver(BuildType.JAVA);
     }
 
+    @Ignore
```
matejonnet: Could you add a comment why the test is ignored ?

auth/src/main/java/org/jboss/pnc/auth/ExternalAuthFacade.java
```diff
@@ -52,7 +53,8 @@
      */
     public ExternalAuthFacade(InputStream keycloakConfiguration) throws Exception{
```
matejonnet: Repeating the package name (auth) in class name is considered as anti-pattern. And abbreviations should be avoided in public API.

moduleconfig/src/main/java/org/jboss/pnc/common/json/moduleprovider/PncConfigProvider.java
```diff
@@ -0,0 +1,82 @@
...
+      this.type = type;  
+      moduleConfigs = new ArrayList<>();  
+      moduleConfigs.add(new ProviderNameType(JenkinsBuildDriverModuleConfig.class,"jenkins-build-driver"));
+      moduleConfigs.add(new ProviderNameType(TermdBuildDriverModuleConfig.class,"termd-build-driver"));
+      moduleConfigs.add(new ProviderNameType(BuildDriverRouterModuleConfig.class,"build-driver-router"));
+      moduleConfigs.add(new ProviderNameType(MavenRepoDriverModuleConfig.class,"maven-repo-driver"));
+      moduleConfigs.add(new ProviderNameType(DockerEnvironmentDriverModuleConfig.class,"docker-environment-driver"));
+      moduleConfigs.add(new ProviderNameType(AuthenticationModuleConfig.class,"authentication-config"));
+      moduleConfigs.add(new ProviderNameType(BpmModuleConfig.class,"bpm-config"));
+    }
```
matejonnet: To avoid updates in common, can each module register itself?

matejonnet:
There are multiple files (integration-test module) in the commits that has organized imports and no other changes. Please try to keep this changes (if required) in a separated commit.

auth/src/main/java/org/jboss/pnc/auth/ExternalAuthFacade.java
pslegr: It is already long there in the master and auth* is not actually part of this commit. We can open separate issue for this

build-coordinator/src/test/java/org/jboss/pnc/core/BuildDriverFactoryTest.java
pslegr: Did fail, with no reason behind a this test is not related with the change. Need to investigate furher on this one

moduleconfig/src/main/java/org/jboss/pnc/common/json/moduleprovider/PncConfigProvider.java
pslegr: Can be, but not in this case ... as the Pnc in the class name defines the provider for PNC project.

moduleconfig/src/main/java/org/jboss/pnc/common/json/moduleprovider/PncConfigProvider.java
pslegr: can not, actually configs are "project" related and not module related -> all modules represents a subtype in the json config files, so they must be registered all of them with the provider

pslegr: And, this is exactly what avoids update in common ! is located in the module config, which MUST be updated for every module adding. On the other hand the common module is general now and does't require update with every new pnc module

PR #497, opened 2015-08-10
[NCL-1109] Refactor build coordinator

pgier:
Remove BuildTasksTree
Update test cases


build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
```diff
@@ -104,40 +102,69 @@ public BuildTask build(BuildConfiguration buildConfiguration, User userTriggered
...
+                buildType,
                 getProductMilestone(buildConfigurationSet));
-        build(buildSetTask, userTriggeredBuild);
+
+        ContentIdentityManager contentIdentityManager = new ContentIdentityManager();
+        String topContentId = contentIdentityManager.getProductContentId(buildSetTask.getBuildConfigurationSet().getProductVersion());
+        String buildSetContentId = contentIdentityManager.getBuildSetContentId(buildSetTask.getBuildConfigurationSet());
+
+        // Loop to create the build tasks
+        for(BuildConfiguration buildConfig : buildSetTask.getBuildConfigurationSet().getBuildConfigurations()) {
```
slaskawi: How about extracting this to a method? This way the comment could be removed...

build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
```diff
@@ -104,40 +102,69 @@ public BuildTask build(BuildConfiguration buildConfiguration, User userTriggered
...
+                    buildSetContentId,
+                    buildContentId,
+                    BuildExecutionType.COMPOSED_BUILD,
+                    user,
+                    buildSetTask,
+                    datastoreAdapter.getNextBuildRecordId());
+            buildSetTask.addBuildTask(buildTask);
+        }
+        // Loop again to set dependencies
+        for (BuildTask buildTask : buildSetTask.getBuildTasks()) {
```
slaskawi: The same here

build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
```diff
@@ -153,25 +180,17 @@ private ProductMilestone getProductMilestone(BuildConfigurationSet buildConfigSe
...
-    private void build(BuildSetTask buildSetTask, User userTriggeredBuild) throws CoreException {
+    private void build(BuildSetTask buildSetTask) throws CoreException {
 
-        BuildTasksTree buildTasksTree = new BuildTasksTree(
-                this,
-                buildSetTask,
-                userTriggeredBuild,
-                () -> datastoreAdapter.getNextBuildRecordId());
-
-        Predicate<Vertex<BuildTask>> acceptOnlyStatus = (vertex) -> {
```
slaskawi: If I remember correctly this comes from JBoss commons dependency. Can we remove it?

build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildSetTask.java
```diff
@@ -126,7 +126,22 @@ public void addBuildTask(BuildTask buildTask) {
...
     }
 
-    public Integer getId() {
+    /**
+     * Get the build task which contains the given build configuration
+     * 
+     * @param buildConfig
+     * @return The build task with the matching configuration, or null if there is none
+     */
+    public BuildTask getBuildTask(BuildConfiguration buildConfig) {
```
slaskawi: This doesn't seem to be a good fit for this method. Maybe it should belong to some utility class?

build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildTask.java
```diff
@@ -188,7 +204,14 @@ public String getBuildContentId() {
     }
 
     void addWaiting(BuildTask buildTask) {
```
slaskawi: Since the field name changed, maybe we should also update this method's name?

build-coordinator/src/test/java/org/jboss/pnc/core/test/buildCoordinator/ReadDependenciesTest.java
```diff
@@ -42,20 +43,21 @@
...
-                null);
-        User user = User.Builder.newBuilder().id(1).build();
-        BuildTasksTree buildTasksTree = new BuildTasksTree(buildCoordinator, buildSetTask, user,() -> taskIdGenerator.incrementAndGet());
+        User user = User.Builder.newBuilder().id(1).username("test-user").build();
+        BuildSetTask buildSetTask = null;
+        try {
+            buildSetTask = buildCoordinator.createBuildSetTask(buildConfigurationSet, user, BuildExecutionType.COMPOSED_BUILD);
+        } catch (CoreException e) {
+            // TODO Auto-generated catch block
+            e.printStackTrace();
```
slaskawi: replace with `fail();`

build-coordinator/src/test/java/org/jboss/pnc/core/test/buildCoordinator/ReadDependenciesTest.java
```diff
@@ -42,20 +43,21 @@
...
+        }
+        //BuildTasksTree buildTasksTree = new BuildTasksTree(buildCoordinator, buildSetTask, user,() -> taskIdGenerator.incrementAndGet());
```
slaskawi: Remove commented code

build-coordinator/src/test/java/org/jboss/pnc/core/test/buildCoordinator/ReadDependenciesTest.java
```diff
@@ -42,20 +43,21 @@
...
 
-        Assert.assertEquals("Missing projects in tree structure.", 5, buildTasksTree.getBuildTasks().size());
+        Assert.assertEquals("Missing build tasks in set.", 5, buildSetTask.getBuildTasks().size());
```
slaskawi: Maybe static import?

build-coordinator/src/test/java/org/jboss/pnc/core/test/buildCoordinator/event/StatusUpdatesTest.java
```diff
@@ -181,18 +185,29 @@ public void BuildSetTaskCallbacksShouldBeCalled() throws DatastoreException {
...
-                BuildExecutionType.COMPOSED_BUILD,
-                null);
-        new BuildTasksTree(buildCoordinator, buildSetTask, null, () -> buildTaskIdSupplier.incrementAndGet());
+        User user = User.Builder.newBuilder().id(1).username("test-user").build();
+        BuildSetTask buildSetTask = null;
+        try {
+            buildSetTask = buildCoordinator.createBuildSetTask(buildConfigurationSet, user, BuildExecutionType.COMPOSED_BUILD);
+        } catch (CoreException e) {
+            // TODO Auto-generated catch block
+            e.printStackTrace();
```
slaskawi: `fail()`

build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildCoordinator.java
matejonnet: We need commons also for String variable replacements ${var}.

pgier: Yes, I have removed the jboss-common-core dependency from the build-coordinator module.

build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildSetTask.java
pgier: I think it's ok here.  The BuildSetTask contains a bunch of BuildTasks, so you can have convenient ways to access them.  In a utility class you would still have to pass both the BuildSetTask and the BuildConfiguration, and then do the same operation.

build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildTask.java
pgier: I agree, I left it there at first for compatibility, but it's better to just remove it.

build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildSetTask.java
matejonnet: I find it ok being here. Looks like it can be converted in stylish J8 one-liner ;) .strem.filter.findFirst ;)

PR #498, opened 2015-08-11
Revert "Ncl 1033 "

alexcreasy:
Reverts project-ncl/pnc#496


slaskawi:
Hey @pslegr!

When adding back this upgrade, could you please add also a test which will check if it works correctly?

PR #503, opened 2015-08-13
Ncl 1091

pgier:


build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildTask.java
```diff
@@ -87,6 +80,7 @@
 
     BuildTask(BuildCoordinator buildCoordinator, 
             BuildConfiguration buildConfiguration, 
+            BuildConfigurationAudited buildConfigurationAudited,
```
matejonnet: Do we really need a reference to both entities (BC and BCAudited)?

pgier: Yes, because the BuildConfigurationAudited does not contain dependency information.  So we need to at least temporarily hold on to the BuildConfiguration until we build the dependency tree of the BuildTasks.

matejonnet: What about adding a BuildConfigurationVersion field instead of audited entity to avoid duplication?

PR #504, opened 2015-08-17
[NCL-1137] Fix the entity relations and transaction scopes related to…

pgier:
… saving BuildRecords

- Reverse the entity ownership relation between build records and build record sets
- Set the start and end time of a build closer to when the build actually runs, instead of when the build coordinator task starts and ends.
- Make the DatastoreAdapter transactional so that it can fetch up to date entities which are related to the BuildRecord

### DIFF
--- a/build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildSetTask.java
+++ b/build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildSetTask.java
@@ -21,7 +21,6 @@ import org.jboss.pnc.core.events.DefaultBuildSetStatusChangedEvent;
 import org.jboss.pnc.core.exception.CoreException;
 import org.jboss.pnc.model.BuildConfigSetRecord;
 import org.jboss.pnc.model.BuildConfiguration;
-import org.jboss.pnc.model.BuildConfigurationAudited;
 import org.jboss.pnc.model.BuildConfigurationSet;
 import org.jboss.pnc.model.ProductMilestone;
 import org.jboss.pnc.spi.BuildExecutionType;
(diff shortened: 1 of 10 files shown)


build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildSetTask.java
```diff
@@ -65,6 +64,7 @@ public BuildSetTask(BuildCoordinator buildCoordinator, BuildConfigSetRecord buil
         this.buildCoordinator = buildCoordinator;
         this.buildConfigSetRecord = buildConfigSetRecord;
         this.buildTaskType = buildTaskType;
+        System.out.println("setting product milestone: " + productMilestone);
```
matejonnet: Use logging.

build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildTask.java
```diff
@@ -47,7 +46,12 @@
 
     public static final Logger log = LoggerFactory.getLogger(BuildTask.class);
 
-    private BuildRecord buildRecord;
+    private Integer id;
+    private BuildConfiguration buildConfiguration;
+    private BuildConfigurationAudited buildConfigurationAudited;
```
matejonnet: Why two configurations? Bare and audited ?

build-coordinator/src/main/java/org/jboss/pnc/core/builder/DatastoreAdapter.java
```diff
@@ -47,6 +46,10 @@
 
     private static final Logger log = Logger.getLogger(DatastoreAdapter.class);
 
+    // needed for EJB/CDI
```
matejonnet: There are no EJBs available in BuildCoordinator, please update miss-leading comment.

PR #507, opened 2015-08-17
The Build Configurations list show duplicate project

thescouser89:
When a user adds several build configurations with the same project, the
project name gets duplicated several times when viewing the list of
build configurations.

http://localhost:8080/pnc-web/#/configuration

This happens because the frontend will loop through the build
configurations, grab the project information for the project id
associated with each build configuration, and add the project
information to a list.

If two build configurations have the same project id, there will be
duplicates of the project info in that list. The duplicates will show up
in the UI.

This commit attempts to fix the duplication by simply adding a 'limitTo'
to the `ng-repeat` command so that the duplicates do not show up.

However, a better fix would be to use a set to store those project info
so that we are guaranteed that there are no duplicates. However there
are no built-in set data structure in ES5 right now.

### DIFF
--- a/ui/app/configuration/views/configuration.list.html
+++ b/ui/app/configuration/views/configuration.list.html
@@ -48,7 +48,7 @@
         </td>
         <td>{{ configuration.description }}</td>
         <td>
-          <div ng-repeat="project in listCtrl.projects | filter: { id: configuration.projectId }">
+          <div ng-repeat="project in listCtrl.projects | filter: { id: configuration.projectId } | limitTo: 1">
             {{ project.name }}
           </div>
         </td>


thescouser89:
retest this please

alexcreasy:
Hi Dustin, thanks very much for this PR. 

I'll review this shortly, but just wanted to draw your attention to the buckets library that we are using: https://github.com/mauriciosantos/Buckets-JS - this implements several standard data structures including sets. Roll on ES6 :)))

My initial thought is to merge this fix, then merge a better fix when it is available.

alexcreasy:
I'm happy for this to go in, merging, Thanks again!

thescouser89:
@alexcreasy I tried `bucketsjs` set implementation, where I transform `projects` variable into a set via `this.projects = buckets.Set()`, then add the results into the set via: `that.projects.add(result)`. In the UI part, I transform the set into an array so that AngularJS knows how to iterate through it.

No luck! It seems the set implementation just stores the first element.

Commit: https://github.com/thescouser89/pnc/commit/6da36ee3ea7403982d33ce043e1e5adce23fc99e

So I'm gonna give up on transforming the 'projects' list into a set for now cause it:

1. doesn't seem to work
2. might break other places that expect the 'projects' variable to be a list, but it's instead getting that weird 'Set' thing.

PR #508, opened 2015-08-18
Add response class to Swagger documentation

thescouser89:
When viewing the Swagger documentation, it can be useful to view the expected response JSON that we will receive from PNC REST endpoints.

This PR attempts to achieve this by defining the response class in the Swagger annotations.


slaskawi:
Looks good, merging

PR #511, opened 2015-08-19
Add UI Validation based on the models

thescouser89:
1. For the Configuration creation view:
   `ui/app/configuration/views/configuration.create.html`

   In `model/src/main/java/org/jboss/pnc/model/BuildConfiguration.java`,
   the fields marked with `@NotNull` are marked as required in the
   web-ui (name, environment, and project)

2. For the Milestone creation view:
   `ui/app/milestone/views/milestone.create-update.html`

   In `model/src/main/java/org/jboss/pnc/model/ProductMilestone.java`,
   the regex validation for the version is reflected in the web-ui too.
   As such, the error message has been updated.
   `ng-pattern` is also changed to `pattern` since this seems to
   activate the regex UI validation.

3. For the Release creation view:
   `ui/app/release/views/release.create-update.html`

   Same explanation as 2. Model found in:
   `model/src/main/java/org/jboss/pnc/model/ProductRelease.java`

4. For the Product Version creation view:
   `ui/app/product/views/product.version.create.html`

   In `pnc/model/src/main/java/org/jboss/pnc/model/ProductVersion.java`,
   the 'version' field has a restriction of being required and have a
   regex validation. The web-ui already marks the version field as
   required, but lacks the regex validation. The regex validation is
   therefore added in the UI.

### DIFF
--- a/ui/app/configuration/views/configuration.create.html
+++ b/ui/app/configuration/views/configuration.create.html
@@ -33,8 +33,9 @@
     <div class="form-group">
       <label for="input-project" class="col-sm-2 control-label">Project</label>
       <div class="col-sm-10">
-      <select ng-model="createCtrl.data.projectId">
-      <option ng-repeat="project in createCtrl.projects" value="{{ project.id }}">
+      <select ng-model="createCtrl.data.projectId" required>
+          <option value="">Select a Project</option>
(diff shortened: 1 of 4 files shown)


thescouser89:
@vibe13 @jsenko Could you guys please review? :smile:

vibe13:
@thescouser89 everything good, thanks for this!

PR #519, opened 2015-08-26
NCL-1199 Removed start/end-time constrains

slaskawi:


slaskawi:
retest this please

pgier:
This seems wrong to me to just create a random start and end time if the build task doesn't contain one.  Why was it not set in the build task in this situation?  Can the times be set to the build task before calling this method?

build-coordinator/src/main/java/org/jboss/pnc/core/builder/DatastoreAdapter.java
```diff
@@ -119,11 +121,13 @@ public void storeResult(BuildTask buildTask, Throwable e) throws DatastoreExcept
      * @return The new (unsaved) build record
      */
     private BuildRecord createBuildRecord(BuildTask buildTask) {
+        Date startTime = Optional.ofNullable(buildTask.getStartTime()).orElse(new Date());
```
matejonnet: Start date can be miss-leading, better to keep it empty than wrong.

PR #522, opened 2015-08-26
Fix record view

thescouser89:
This attempts to fix an issue in the Records page due to refactorings in #517 .

There is also another commit to add displayName for both the log and output tab of records, so that the breadcrumbs show the record id.


thescouser89:
@jdg-bot :metal:

PR #538, opened 2015-09-09
add ApiModelProperty for status

thauser:
This annotation should likely be added here, as it is for Configuration's BuildStatus field.


slaskawi:
:+1:

PR #546, opened 2015-09-15
NCL-1229 Use dedicated response entities for endpoints

slaskawi:
* Upgrade Swagger to 1.5.3


thauser:
Hi Sebastian,

I've deployed your NCL-1229 branch locally, and I think we need a few minimal changes. It seems as though the following 
```
    @ApiOperation(value = "Gets a specific BuildRecordSet", response = BuildRecordSetSingleton.class)
    @ApiResponses(value = {
            @ApiResponse(code = SUCCESS_CODE, message = SUCCESS_DESCRIPTION),
            @ApiResponse(code = INVLID_CODE, message = INVALID_DESCRIPTION),
            @ApiResponse(code = NOT_FOUND_CODE, message = NOT_FOUND_DESCRIPTION),
            @ApiResponse(code = SERVER_ERROR_CODE, message = SERVER_ERROR_DESCRIPTION)
    })
    @GET
    @Path("/{id}")
    public Response getSpecific(@ApiParam(value = "BuildRecordSet id", required = true) @PathParam("id") Integer id) {
        return super.getSpecific(id);
    }
```

produces this snippet in swagger.json (forgive the bad formatting)
```
"/build-record-sets/{id}": {
    "get": 
{
    "tags": 
[
    "build-record-sets"
],
"summary": "Gets a specific BuildRecordSet",
"description": "",
"operationId": "getSpecific",
"consumes": 
[
    "application/json"
],
"produces": 
[
    "application/json"
],
"parameters": 
[
    {
        "name": "id",
        "in": "path",
        "description": "BuildRecordSet id",
        "required": true,
        "type": "integer",
        "format": "int32"
    }
],
"responses": 
{
    "200": 
{
    "description": "Success with results"
},
"400": 
{
    "description": "Invalid input parameters or validation error"
},
"404": 
{
    "description": "Can not find specified result"
},
"500": 
        {
            "description": "Server error"
        }
    }
},
```
Note how there is no mention of the response class declared in the @ApiOperation. I did a test and removed the @ApiResponse(code = SUCCESS_CODE, message = SUCCESS_DESCRIPTION), and the new swagger.json looks like this:
```
"responses": {

    "200": 

{

    "description": "successful operation",
    "schema": 

    {
        "$ref": "#/definitions/BuildRecordSetSingleton"
    }

},
"400": 
{

    "description": "Invalid input parameters or validation error"

},
"404": 
{

    "description": "Can not find specified result"

},
"500": 

    {
        "description": "Server error"
    }

}
```

Without this change, the generated client from swagger-codegen thinks all of the REST calls return nothing; after changing the Endpoints to exclude an explicit @ApiOperation for HTTP 200 status, things work as expected. I'm not sure what to think of this behavior, but that's how it is :) I'll add here if I find anything more.

thauser:
If there's a way I can push changes up to the PR, I can do the work needed.

thauser:
Additionally, is there a particular reason why:
```
/build-configuration-sets/{id}/build
/build-configuration/{id}/build
```
have @Consumes(MediaType.WILDCARD) as opposed to MediaType.APPLICATION_JSON?

slaskawi:
hmmm I took a look at some Github issues regarding to this weird error [1][2]. We might also try to put HTTP 200 response 1 line below :smiling_imp:. Or remove it as you suggested - it seems Swagger assumes that GET methods always return this code if everything goes well.

BTW - No worries, I'll rebase this PR and push everything tomorrow :)

And the last thing - the `@consumes(WILDCARD)` - yeah I think we can change it to JSON.

[1] https://github.com/swagger-api/swagger-ui/issues/680
[2] https://github.com/swagger-api/swagger-ui/issues/674

slaskawi:
And thanks a lot for checking it!

slaskawi:
It looks like each `@ApiResponse` needs it's own response attribute..... sigh... we need a bit more duplication..

BTW - I also decided to leave the wildcard in triggering. Mainly because it accepts an empty POST, do adding additional headers is not so obvious in this case.

PR #548, opened 2015-09-16
NCL-1152 Added REST Endpoints for rebuilding all deps

slaskawi:


build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildScheduler.java
```diff
@@ -29,5 +29,6 @@
  * @author <a href="mailto:matejonnet@gmail.com">Matej Lazar</a>
  */
 public interface BuildScheduler {
+    boolean hasBeenBuilt(BuildTask buildTask);
```
matejonnet: Build coordinator should be the one who decides if we need to schedule a build or not. BuildScheduler is an abstraction which allows to select the executor. Executor doesn't hold the information of what was build and what not, its responsibility is to run the build and other tasks around it, trying to be as stateless as possible.

build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildTask.java
```diff
@@ -84,18 +84,12 @@
...
-              User user,
-              Date submitTime,
-              BuildSetTask buildSetTask,
-              int id,
-              Event<BuildStatusChangedEvent> buildStatusChangedEventNotifier,
-              Consumer<BuildTask> onAllDependenciesCompleted) {
+    private BuildTask(BuildConfiguration buildConfiguration, BuildConfigurationAudited buildConfigurationAudited, String topContentId,
+            String buildSetContentId, String buildContentId, User user, Date submitTime, BuildSetTask buildSetTask, int id,
+            Event<BuildStatusChangedEvent> buildStatusChangedEventNotifier, Consumer<BuildTask> onAllDependenciesCompleted,
+            boolean rebuildAll) {
```
matejonnet: Please arrange the parameters in new lines to increase readability.

build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildScheduler.java
slaskawi: I'm not entirely sure if this is the right approach... consider the following use cases:

1. Each scheduler might use different policies to check whether a BC has been built. For example I can imagine, that BPM might check additionally Brew (or any other repository) whereas Local builder might check only DB.

2. BC should be checked just *before* the build (think about building 2 products with the same BC built concurrently). This can only be done at scheduler level. BuildCoordinator just takes the BuildTask and passes it to the scheduler.

To wrap it up - if I move this check to the build coordinator, we might not be able to meet those 2 use cases. The question is - is it ok?

build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildTask.java
slaskawi: +1

build-coordinator/src/main/java/org/jboss/pnc/core/builder/BuildScheduler.java
matejonnet: 1. I don't see a reason for a different requirement to check if the builds exists or not. And if they do we should extract that functionality to a new class rather than scheduler. p.s.: local is meant as local to coordinator, which is likely to be dropped once BPM work-flow is in completely in place.

2. the check is already (this PR) called in the BuildCoordinator, which is OK, I would move just the method that does the checking.

slaskawi: Okey dokey, I'll prepare an update tomorrow. Maybe it would be a good idea to extract this checking-if-build-exists logic into a separate class? This way it would be very easy to find and extend it (if needed). I have a strong feeling that we will change it a couple of times in the future.

matejonnet: +1 for extracting it out.

slaskawi: extracted

slaskawi:
Comments addressed - merging

PR #553, opened 2015-09-17
Add BuildScheduler selector.

matejonnet:
@slaskawi Do you find ok the renaming build-driver config to more general?

### DIFF
--- a/build-coordinator/src/main/java/org/jboss/pnc/core/builder/coordinator/bpm/BpmBuildScheduler.java
+++ b/build-coordinator/src/main/java/org/jboss/pnc/core/builder/coordinator/bpm/BpmBuildScheduler.java
@@ -48,7 +48,6 @@ import java.util.function.Consumer;
 /**
  * @author <a href="mailto:matejonnet@gmail.com">Matej Lazar</a>
  */
-@Alternative
 @ApplicationScoped
 public class BpmBuildScheduler implements BuildScheduler {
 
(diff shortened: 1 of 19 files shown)


build-coordinator/src/main/java/org/jboss/pnc/core/builder/coordinator/bpm/BpmBuildScheduler.java
```diff
@@ -58,13 +57,18 @@
 
     private BpmCompleteListener bpmCompleteListener;
 
-    private final URL instanceUrl;
```
slaskawi: Why URL was bad? Currently we miss the validation piece...

common/src/main/resources/pnc-config.json
```diff
@@ -2,6 +2,11 @@
 "@class":"ModuleConfigJson","name":"pnc-config",
         "configs":[
           {
+            "@module-config":"system-config",
+            "buildDriverId":"termd-build-driver",
```
slaskawi: You probably need to change also NCL scripts.

build-coordinator/src/main/java/org/jboss/pnc/core/builder/coordinator/bpm/BpmBuildScheduler.java
matejonnet: It required valid URL also if you don't use BPM and it caused an error due to our env variable reference. Validation is still done in later stage, when you are connecting to BPM server. I prefer the first option but a compromise is required.

common/src/main/resources/pnc-config.json
matejonnet: Changing autodeploy one ...

matejonnet:
Retest this please.

PR #554, opened 2015-09-18
Update for AProx 0.24.0 with group promotion feature

jdcasey:
also implement direct promotion to shared-releases group


jdcasey:
retest this please

PR #564, opened 2015-09-21
REST Tweaks

thauser:
A few tweaks to the swagger api-docs declarations.


slaskawi:
I'n not entirely sure about this change. Since we send an empty POST to trigger a build, maybe it should accept everything? BTW - did you check if that works with the UI?

thauser:
It does indeed work with the UI; however I can see how accepting */* for an empty POST might be advantageous. How about the change to use ProductVersionPage instead of ProductVersionRest?

slaskawi:
:+1: for using ProductVersionPage :) My mistake sorry :)

PR #571, opened 2015-09-22
Revert "Disable skipping completed builds."

slaskawi:
Reverts project-ncl/pnc#555


slaskawi:
I need to add a check if there are no additional Audited BC created...

PR #574, opened 2015-09-22
fix NCL-1248 - reflecting Swagger ApiKeyAuthorization class changes a…

pslegr:
…fter update


slaskawi:
:+1: Thanks for the fix!

PR #591, opened 2015-09-29
NCL-1186

jsenko:
Paging for UI.
To test the paging more easily you can change the default page size in the PageFactory::build


alexcreasy:
@slaskawi would you mind also reviewing this PR as it makes REST API changes.

rest/src/main/java/org/jboss/pnc/rest/endpoint/RunningBuildRecordEndpoint.java
```diff
@@ -88,9 +70,8 @@ public RunningBuildRecordEndpoint(BuildRecordProvider buildRecordProvider) {
     @GET
     public Response getAll(@ApiParam(value = PAGE_INDEX_DESCRIPTION) @QueryParam(PAGE_INDEX_QUERY_PARAM) @DefaultValue(PAGE_INDEX_DEFAULT_VALUE) int pageIndex,
             @ApiParam(value = PAGE_SIZE_DESCRIPTION) @QueryParam(PAGE_SIZE_QUERY_PARAM) @DefaultValue(PAGE_SIZE_DEFAULT_VALUE) int pageSize,
-            @ApiParam(value = SORTING_DESCRIPTION) @QueryParam(SORTING_QUERY_PARAM) String sort,
-            @ApiParam(value = QUERY_DESCRIPTION, required = false) @QueryParam(QUERY_QUERY_PARAM) String q) {
-        return fromCollection(buildRecordProvider.getAllRunning(pageIndex, pageSize, sort, q));
```
slaskawi: Why did we remove sorting?

slaskawi: Actually it looks like this question applies to all endpoints

slaskawi:
Looks really nice from my point of view! 

Just a one question (inlined) and could you please add a test to `RSQLTest` for "like"?

rest/src/main/java/org/jboss/pnc/rest/endpoint/RunningBuildRecordEndpoint.java
jsenko: That resource does not come from the database so RSQL & Criteria API cannot be used. Before I removed the 'sort' and 'q' parameters they were there but didn't work so i didn't see a reason to keep them because that was misleading. I would have to parse the sort string and use reflection to implement the same functionality but i didn't see the need because I sort by 'id' everywhere so iIjust hardcoded that.

rest/src/main/java/org/jboss/pnc/rest/provider/BuildRecordProvider.java
```diff
@@ -69,25 +68,69 @@ public BuildRecordProvider(BuildRecordRepository buildRecordRepository, BuildCoo
         this.buildExecutor = buildExecutor;
     }
 
-    public CollectionInfo<BuildRecordRest> getAllRunning(Integer pageIndex, Integer pageSize, String sortingRsql, String rsql) {
-        if(!Strings.isNullOrEmpty(sortingRsql)) {
-            logger.warn("Sorting RSQL is not supported, ignoring");
-        }
+    public CollectionInfo<BuildRecordRest> getAllRunning(Integer pageIndex, Integer pageSize, String search) {
+        List<BuildTask> x = buildCoordinator.getActiveBuildTasks();
+        return nullableStreamOf(x)
```
matejonnet: Please use descriptive variable names instead of x.

PR #605, opened 2015-09-30
BPM Scheduler based on HTTPClient

slaskawi:
DRAFT ONLY, Please do not integrate


slaskawi:
Closing - newer KIE client seems to work (yay!)

PR #607, opened 2015-09-30
[NCL-533] Improve confirmation panels

thescouser89:
This is done by integrating sweetalert with PNC.

delete_form


thescouser89:
@emmettu This might break the PNC end-to-end testing. :smiley:

slaskawi:
Wow, looks awesome. @alexcreasy @vibe13 WDYT?

vibe13:
They look very good, like them a lot.

alexcreasy:
Thanks for this @thescouser89 - I think this is much better than what we currently have. 

Personally I think it would be better to have something based on the patternfly modal widget [see: https://www.patternfly.org/widgets/#modal]. Although I realise this will probably require more coding out of the box, sticking with patternfly widgets whenever possible is recommended to: 

1) Give a consistent JBoss look and feel to the project
2) Use UX patterns that have undergone user testing by our UX team 
3) If the UX team refresh the patternfly style in the future, we only have to upgrade the library.

That said, I'd be happy to merge this for now, with one reservation. I think the "Done" screen is a little redundant. We have the existing patternfly notification to let you know the action was successful and more importantly it requires the user to make another unnecessary click. I think the unnecessary extra click will be a bit frustrating for users.

Does anyone else have any thoughts on this?

thescouser89:
@alexcreasy I can remove the "Done" screen. I agree with your point that it's a little bit redundant.

thescouser89:
PR  #619 uses modals instead of sweetalert

PR #609, opened 2015-10-01
NCL-1276 UI authentication enhancement. The UI will only redirect to login page when the user performs an action that requires authentication.

alexcreasy:


alexcreasy:
@vibe13 @rnc wdyt? It gives us more brew like authentication system (i.e. login only required to perform write actions)

alexcreasy:
test this please

alexcreasy:
I'm going to go ahead and merge this as it will make progressing with the general UI authentication overhaul easier, if you guys wouldn't mind commenting still, we can always back this feature out if we don't like it. @pslegr I meant to tag you in the original post!

PR #611, opened 2015-10-02
replace url of Keycloak server with token string

pslegr:


pslegr:
retest this please

pslegr:
retest this please

pslegr:
retest this please

slaskawi:
Frankly - I don't like this change... so everyone that would like to test with `-Dauth` would have to use sed? It doesn't seem ok to me. There must be a better way.

PR #619, opened 2015-10-05
[NCL-533] Improve confirmation panels

thescouser89:
This is done by integrating modals with PNC.

out

### DIFF
--- a/ui/app/common/util/views/confirm-click.html
+++ b/ui/app/common/util/views/confirm-click.html
@@ -1,0 +1,26 @@
+<!--
+
+    JBoss, Home of Professional Open Source.
+    Copyright 2014 Red Hat, Inc., and individual contributors
+    as indicated by the @author tags.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<div class="modal-header">
+    <h3 class="modal-title">{{ message }}</h3>
+</div>
+<div class="modal-footer">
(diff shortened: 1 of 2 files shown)


thescouser89:
Instead of using sweetalert library in #607, use Bootstrap modals instead as suggested.

thescouser89:
@jdg-bot :raised_hands:

alexcreasy:
@thescouser89 I'm loving it thanks for this, can I be a real pain and ask for two more slight changes ?

1) The confirm buttons should use the "danger" patternfly style, so it's red and people know they are doing something destructive? 

2) I feel cancel should maybe be the default button colour (i.e. grey) rather than using the "warning" orange colour, as cancel is the safe option.

The 4 different colour styles in patternfly have semantic meaning that we should try to abide by as it makes things more intuitive for the end user. The current method makes people think cancel is possibly dangerous and submit is perfectly safe.

WDYT?

ui/app/common/util/views/confirm-click.html
```diff
@@ -0,0 +1,26 @@
...
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<div class="modal-header">
+    <h3 class="modal-title">{{ message }}</h3>
+</div>
+<div class="modal-footer">
+  <button class="btn btn-warning" type="button" ng-click="cancel()">Cancel</button>
```
alexcreasy: Could the css class be: class="btn btn-default" ?

ui/app/common/util/views/confirm-click.html
```diff
@@ -0,0 +1,26 @@
...
+  <button class="btn btn-primary" type="button" ng-click="confirm()">Confirm</button>
```
alexcreasy: could the css class be: class="btn btn-danger"

alexcreasy:
Just to emphasise this is an awesome impl, in case it got lost in the critique above :+1:  :)

thescouser89:
@alexcreasy Updated! Thanks for the suggestion :smiley:

alexcreasy:
Test this please

alexcreasy:
Perfect thanks @thescouser89 will merge when it passes CI!

thescouser89:
@jdg-bot :star:

alexcreasy:
I think I caused an extra build :worried:

PR #626, opened 2015-10-07
MAITAI-3445 Added GA to BuildConfiguration

slaskawi:


model/src/main/java/org/jboss/pnc/model/BuildConfiguration.java
```diff
@@ -60,6 +59,10 @@
     @NotNull
     private String name;
 
+    @Column(unique = true)
+    @NotNull
+    private GA ga;
```
matejonnet: Is GA general enough to use it for all kind of builds? Also for non Java builds?

slaskawi:
Closing - we will use BC.name

PR #630, opened 2015-10-08
BPM integration fixes.

matejonnet:


build-coordinator/src/main/java/org/jboss/pnc/core/builder/coordinator/BuildCoordinator.java
```diff
@@ -278,7 +278,7 @@ void processBuildTask(BuildTask buildTask) {
             activeBuildTasks.add(buildTask);
             buildScheduler.startBuilding(buildTask, onComplete);
         } catch (CoreException e) {
-            log.warn("[" + buildTask.getId() + "] Build task failed. Setting it as SYSTEM_ERROR.", e);
+            log.warn("[" + buildTask.getId() + "] Build coordination task failed. Setting it as SYSTEM_ERROR.", e);
```
slaskawi: you might use `log.warn("blah {} blan", myparam)`. This way you will avoid String concatenation

matejonnet: Yea, I need to switch to slf4j, here is jboss.logging used.

matejonnet:
Retest this please.

PR #641, opened 2015-10-13
Fix for relative path for import

thescouser89:
If the '/' in the path is omitted, PNC will try to access:

    'localhost:8080/pnc-web/da-bcg ...'

instead of:

    'localhost:8080/da-bcg ...'

This commit fixes it.


thescouser89:
@jdg-bot :sob:

PR #642, opened 2015-10-13
Manage the call to a completed BuildRecord endpoint if the BuildRecor…

vibe13:
…d is in running state, to avoid 404

@alexcreasy fix of the remaining bug


vibe13:
test this please

PR #650, opened 2015-10-14
Managed the system_error more gracefully when a triggered build fails…

vibe13:
… to start on the server (no more 404 error)


vibe13:
@alexcreasy as discussed on IRC

alexcreasy:
@vibe13 great! I'll create a ticket to make notifications more extensible in the next sprint, but this will do nicely for 0.7

PR #658, opened 2015-10-15
Several UI bugfixes.

jsenko:


jsenko:
i'll create separate PR for each commit

PR #678, opened 2015-10-20
Added UI validation in the Product Import Process pages

vibe13:


jsenko:
validation also has to be added here https://github.com/project-ncl/pnc/blob/master/ui/app/import/product/ProductImportCtrl.js#L142, to be the same as in the form. just add the regexp match check there

vibe13:
retest this please

PR #679, opened 2015-10-20
Clonning adds date at the beginning

slaskawi:


model/src/main/java/org/jboss/pnc/model/BuildConfiguration.java
```diff
@@ -547,7 +568,7 @@ public int hashCode() {
     public BuildConfiguration clone() {
         try {
             BuildConfiguration clone = (BuildConfiguration) super.clone();
-            clone.name = "_" + name;
+            clone.name = CLONED_DATE_FORMATTER.format(LocalDateTime.now()) + "_" + name;
```
slaskawi: @vibe13 @alexcreasy Just letting you know - clonned BC will have a date at the beginning.

PR #690, opened 2015-11-02
Fix inverted logic for adding a dependency. 

thauser:
validateCondition needs the boolean parameter to be true, otherwise validation will fail. Without this change, the validateCondition will only pass when configId == dependencyId, which cannot be the intent. With this fix, NCL-1264 becomes possible.


thauser:
Something seems to be up with the Testing job; it states that JAVA_HOME is not correctly defined:
Error: JAVA_HOME is not defined correctly.
  We cannot execute /opt/jdk1.8.0_25/bin/java

thauser:
Re-test this please

slaskawi:
Looks good integrating!

PR #691, opened 2015-11-04
Allow arbitray ordering of angular module declarations/definitions.

jsenko:


jsenko:
its better just to rename the files...

PR #699, opened 2015-11-17
Ncl 1447 / NCL-1338- CurrentMilestoneId attribute is not updated in ProductVersion

vibe13:
@alexcreasy @matejonnet this should be cherry picked to 0.7 branch also

### DIFF
--- a/rest/src/main/java/org/jboss/pnc/rest/restmodel/ProductVersionRest.java
+++ b/rest/src/main/java/org/jboss/pnc/rest/restmodel/ProductVersionRest.java
@@ -64,8 +64,8 @@ public class ProductVersionRest implements GenericRestEntity<Integer> {
         this.id = productVersion.getId();
         this.version = productVersion.getVersion();
         this.productId = productVersion.getProduct().getId();
-        this.currentProductMilestoneId = productVersion.getCurrentProductMilestone() != null ? productVersion
-                .getCurrentProductMilestone().getId() : null;
+        this.currentProductMilestoneId = productVersion.getCurrentProductMilestone() != null
+                ? productVersion.getCurrentProductMilestone().getId() : null;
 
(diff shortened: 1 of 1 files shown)


vibe13:
test this please

vibe13:
test this please

vibe13:
@alexcreasy  any comments on this?

rest/src/main/java/org/jboss/pnc/rest/restmodel/ProductVersionRest.java
```diff
@@ -158,6 +158,21 @@ public ProductVersion toProductVersion(Product product) {
...
+            List<ProductMilestone> productMilestones = nullableStreamOf(productVersion.getProductMilestones())
+                    .filter(productMilestone -> productMilestone.getId().equals(currentProductMilestoneId))
+                    .collect(Collectors.toList());
+
+            if (!productMilestones.isEmpty()) {
+                productVersion.setCurrentProductMilestone(productMilestones.get(0));
+            } else {
+                productVersion.setCurrentProductMilestone(null);
+            }
+        } else {
```
alexcreasy: Is this else clause and the one nested above it necessary? It looks to me as if the current product milestone would already be null if they were invoked? Probably I don't understand the context of this, but just thought I'd check :)

vibe13: I wanted to make sure that if a milestoneId is updated to null from a non-null value, or if it is set to a non-existing id it would be set to null

alexcreasy: OK cool.

alexcreasy:
@vibe13 :+1: from me!

PR #708, opened 2015-11-23
NCL-1454 - Model description fields are VARCHAR255 but should support arbitrary length strings

alexcreasy:


alexcreasy:
Test this please

PR #712, opened 2015-12-01
NCL-1478

thauser:
Fix for NCL-1478; wrap the cloned BuildConfiguration within the Singleton construct.


thauser:
Problem with the commits in the PR. Resubmitting.

PR #713, opened 2015-12-01
fix for NCL-1478; wrap rest entity in Pagination 

thauser:
Fix for NCL-1478, properly wrap the cloned BC in a Singleton instance.


thauser:
These changes are included in my latest PR. Closing.

PR #714, opened 2015-12-02
Build configuration endpoint changes

thauser:
Modify the 'build' method in BuildConfigurationEndpoint to correctly indicate that a 200 response has a BuildRecordSingleton, instead of nothing. This corrects the swagger documentation on this method.

Wrap the response in the 'clone' method of  BuildConfigurationEndpoint instead of returning the raw BuildConfigurationRest


slaskawi:
Looks good, integrated

PR #716, opened 2015-12-03
NCL-1472 Triggering builds and returning BuildTasks is atomic

slaskawi:


slaskawi:
@matejonnet Could you please do a Code Review?

PR #718, opened 2015-12-07
Use Undertow WebSocket client instead of container provided.

matejonnet:


termd-build-driver/pom.xml
```diff
@@ -61,9 +61,8 @@
     </dependency>
 
     <dependency>
-      <groupId>org.jboss.spec.javax.websocket</groupId>
-      <artifactId>jboss-websocket-api_1.0_spec</artifactId>
-      <scope>provided</scope>
+      <groupId>io.undertow</groupId>
+      <artifactId>undertow-websockets-jsr</artifactId>
```
slaskawi: I think we should use library provided by the container. This way we package it with our WAR/EAR.

matejonnet: The EAP version we use still uses Tomcat.

PR #730, opened 2015-12-14
NCL-1496 Product Import: Reset modal is displayed after process finis…

jsenko:
…hed.


vibe13:
@jbartece @thescouser89 looks good to me

PR #732, opened 2015-12-14
Changed Type annotation of BuildLogs to a non deprecated one; forced …

vibe13:
…Lazy fetching; specified an additional property to use streams (may be overridden by driver implementation)


vibe13:
test this please

PR #743, opened 2015-12-21
NCL-1532 - Added annotations to force db index generation

vibe13:


vibe13:
@matejonnet @slaskawi @pgier these annotations should force index generations on FK, as they are not created by default

vibe13:
retest this please

vibe13:
@matejonnet @slaskawi @pgier I believe this is good to merge, also looking at execution plans in db, with indexes the performance is improved. wdyt? Since the indexes are created only with ddl create (not update), a script to create them in demo will be provided

PR #750, opened 2016-01-08
update consumes API documentation for swagger

thauser:
Slightly modify the @Consumes annotation for swagger.json. Without this slight tweak, clients generated from swagger-codegen will be incorrect.


rest/src/main/java/org/jboss/pnc/rest/endpoint/BuildConfigurationSetEndpoint.java
```diff
@@ -276,7 +276,7 @@ public Response getBuildRecords(
     })
     @POST
     @Path("/{id}/build")
-    @Consumes(MediaType.WILDCARD)
```
slaskawi: What about other methods that trigger build? Do they need the same update?

thauser:
Sorry, I haven't had the chance to look into this yet. I'll take a look and post more changes in this PR if needed.

thauser:
I've looked through the calls that build and it seems that they all have @Consumes declared appropriately. I'll open a new PR / issue if I discover more through PNC-CLI testing / development.

PR #755, opened 2016-01-13
Handle NPE when there are missing parameters

vibe13:


vibe13:
@alexcreasy CI finished

alexcreasy:
@vibe13 great it's merged, let's see what happens after it's deployed

PR #756, opened 2016-01-14
NCL-1481 - Product Import: The tree of dependencies should not automa…

vibe13:
…tically mark parents of a selected dependency


vibe13:
retest this please

PR #760, opened 2016-01-17
Ncl 1431 split build executor from build coordinator

matejonnet:


build-coordinator/pom.xml
```diff
@@ -45,6 +45,10 @@
     </dependency>
     <dependency>
       <groupId>org.jboss.pnc</groupId>
+      <artifactId>rest-model</artifactId>
```
slaskawi: Why are we extracting REST model after all? Do we plan to give it to someone?

build-coordinator/src/main/java/org/jboss/pnc/core/builder/coordinator/BuildCoordinator.java
```diff
@@ -228,29 +230,25 @@ private ProductMilestone getProductMilestone(BuildConfigurationSet buildConfigSe
...
                     .forEach(v -> processBuildTask(v));
         }
     }
 
+    private boolean rejectAlreadySubmitted(BuildTask buildTask) {
+        if (isBuildAlreadySubmitted(buildTask)) {
+            buildTask.setStatus(BuildCoordinationStatus.REJECTED);
+            buildTask.setStatusDescription("The configuration is already in the build queue.");
+            return false;
+        } else {
```
slaskawi: else can be removed

build-coordinator/src/main/java/org/jboss/pnc/core/builder/coordinator/BuildCoordinator.java
```diff
@@ -57,17 +59,17 @@
...
-    private Logger log = LoggerFactory.getLogger(BuildCoordinator.class);
+    private final Logger log = LoggerFactory.getLogger(BuildCoordinator.class);
 
     /**
      * Build tasks which are either waiting to be run or currently running.
      * The task is removed from the queue when the build is complete and the results
      * are stored to the database.
      */
-    private Queue<BuildTask> activeBuildTasks = new ConcurrentLinkedQueue<>(); //TODO garbage collector (time-out, error state)
+    private final Queue<BuildTask> activeBuildTasks = new ConcurrentLinkedQueue<>(); //TODO garbage collector (time-out, error state)
```
slaskawi: What are we going to do with this queue? IMO - the easiest solution is to use JMS here.

build-coordinator/src/main/java/org/jboss/pnc/core/builder/coordinator/BuildScheduler.java
```diff
@@ -33,7 +33,7 @@
  * @author <a href="mailto:matejonnet@gmail.com">Matej Lazar</a>
  */
 public interface BuildScheduler {
-    void startBuilding(BuildTask buildTask, Consumer<BuildStatus> onComplete) throws CoreException;
+    void startBuilding(BuildTask buildTask, Consumer<BuildResult> onComplete) throws CoreException, ExecutorException;
```
slaskawi: From client code perspective - why do we need 2 exceptions here? Both are unrecoverable...

build-coordinator/src/test/java/org/jboss/pnc/core/builder/coordinator/bpm/BpmBuildSchedulerTest.java
```diff
@@ -48,6 +46,7 @@
 
 public class BpmBuildSchedulerTest {
 
+    @Ignore //TODO what exactly are we testing here?
```
slaskawi: We are checking if properties are extracted based on configuration.

build-executor-mock/src/main/java/org/jboss/pnc/mock/environmentdriver/BuildExecutorMock.java
```diff
@@ -0,0 +1,136 @@
...
+        } catch (InterruptedException e) {
+            log.warn("Mock build interrupted.", e);
+        }
+        return buildPassed;
+    }
+
+
+    //    private CompletableFuture<Integer> mockBuild() {
+//        CompletableFuture<Integer> waitToCompleteFuture = new CompletableFuture<>();
+//        int sleep = RandomUtils.randInt(50, 500);
```
slaskawi: to be removed?

integration-test/src/test/java/org/jboss/pnc/integration/WebSocketsNotificationTest.java
```diff
@@ -128,8 +128,8 @@ private void waitForWSClientConnection() {
 
     private void awaitFor(Supplier<Boolean> condition, int timeMs) {
         long waitUntil = System.currentTimeMillis() + timeMs;
-        while (System.currentTimeMillis() < waitUntil) {
-            if (condition.get()) {
+        while(System.currentTimeMillis() < waitUntil) {
```
slaskawi: Are you sure you are using the correct formatter? If the answer is yes - that we probably need to sync IntelliJ/Eclipse stuff...

pnc-mock/src/main/java/org/jboss/pnc/mock/model/builders/ArtifactBuilder.java
```diff
@@ -0,0 +1,39 @@
+/**
```
slaskawi: Shouldn't we call this module 'model-mock'?

rest-model/pom.xml
```diff
@@ -0,0 +1,254 @@
...
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
```
slaskawi: We probably don't this this stuff here.

rest-model/pom.xml
```diff
@@ -0,0 +1,254 @@
...
+        <filtering>true</filtering>
+      </resource>
+    </resources>
+
+    <plugins>
+      <plugin>
+        <artifactId>maven-war-plugin</artifactId>
+        <configuration>
+          <attachClasses>true</attachClasses>
+          <webResources>
```
slaskawi: and this

rest-model/pom.xml
```diff
@@ -0,0 +1,254 @@
...
+                    <executions>
+                      <execution>
+                        <id>default-copy-resources</id>
+                        <phase>process-resources</phase>
+                        <goals>
+                          <goal>copy-resources</goal>
+                        </goals>
+                        <configuration>
+                          <overwrite>true</overwrite>
+                          <outputDirectory>${project.build.directory}/${project.artifactId}/WEB-INF/</outputDirectory>
```
slaskawi: and this ;)

build-coordinator/src/main/java/org/jboss/pnc/core/builder/coordinator/BuildCoordinator.java
matejonnet: Looking at it I think there is no reason to use queue, it could be distributed set. Based on JDG cache?
In order to make it work properly in cluster (queue or cache) we must make sure all changes (buildTaks.status) are pushed into grid and all updates notified back.

build-coordinator/src/main/java/org/jboss/pnc/core/builder/coordinator/BuildScheduler.java
matejonnet: +1

pnc-mock/src/main/java/org/jboss/pnc/mock/model/builders/ArtifactBuilder.java
matejonnet: Yes, we probably gonna end up with *-mock modules to avoid circular dependencies.

rest-model/pom.xml
matejonnet: Logging is used in BuildResultSerializationTest which is part of this module.

rest-model/pom.xml
matejonnet: +1

rest-model/pom.xml
matejonnet: +1

build-coordinator/src/test/java/org/jboss/pnc/core/builder/coordinator/bpm/BpmBuildSchedulerTest.java
matejonnet: After the refactor how parameters are put together I don't think this test is relevant anymore. We have two set of parameters `processParameters` which is a map serialized to json. Testing means we are testing `ObjectMapper`.
And second is `BuildExecutionConfigurationREST` which serialization is tested else where.
The only thing remaining here is `BpmModuleConfig` which should also be tested elsewhere.
Do you think we need to test something else? 
Parameter preparation doesn't have anything with process starting, this test is too complex to test the parameters.
I would remove it, wdyt ?

build-coordinator/pom.xml
jdcasey: I was going to ask for this anyway, since we need some method of deserializing responses from PNC in the brew bridge, and it'd be helpful if we didn't have to write our own model classes for this.

build-coordinator/src/main/java/org/jboss/pnc/core/builder/coordinator/bpm/BpmBuildScheduler.java
```diff
@@ -122,103 +113,51 @@ ProcessInstance startProcess(BuildTask buildTask, Integer buildTaskSetId) throws
...
+                buildConfiguration.getName(),
+                buildConfiguration.getScmRepoURL(),
+                buildConfiguration.getScmRevision(),
+                buildConfiguration.getScmMirrorRepoURL(),
+                buildConfiguration.getScmMirrorRevision(),
+                buildConfiguration.getBuildEnvironment().getBuildType());
+
+        BuildExecutionConfigurationRest buildExecutionConfigurationREST = new BuildExecutionConfigurationRest(buildExecutionConfiguration);
+
+        return buildExecutionConfigurationREST.toString();
```
jdcasey: Why call toString() here? That method is typically more used for debugging purposes...should we be using a proper serializer framework of some sort instead?

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/MavenRepositorySession.java
```diff
@@ -136,7 +156,7 @@ public RepositoryManagerResult extractBuildArtifacts() throws RepositoryManagerE
 
         promoteToBuildContentSet();
 
-        return new MavenRepositoryManagerResult(uploads, downloads);
+        return new MavenRepositoryManagerResult(uploads, downloads, buildRepoId); //TODO buildRepoId == buildContentId ?
     }
```
jdcasey: Yep, it just needs to be updated.

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/MavenRepositorySession.java
```diff
@@ -136,7 +156,7 @@ public RepositoryManagerResult extractBuildArtifacts() throws RepositoryManagerE
 
         promoteToBuildContentSet();
 
-        return new MavenRepositoryManagerResult(uploads, downloads);
+        return new MavenRepositoryManagerResult(uploads, downloads, buildRepoId); //TODO buildRepoId == buildContentId ?
```
jdcasey: Yep, they're the same.

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/MavenRepositorySession.java
```diff
@@ -67,6 +79,7 @@
 
     // TODO: Create and pass in suitable parameters to Aprox to create the
     //       proxy repository.
+    @Deprecated
```
jdcasey: I think the buildContentSetId was a holdover from before we decided to do the common component pool. I would have been necessary for chained builds before that. Now, with CCP I think we can use the untested builds artifacts collection as a rough equivalent to the chain build set, since promotion into this untested level should be a prerequisite for any dependency projects to build.

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
```diff
@@ -45,7 +45,10 @@
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
-import static org.jboss.pnc.mavenrepositorymanager.MavenRepositoryConstants.*;
+import static org.jboss.pnc.mavenrepositorymanager.MavenRepositoryConstants.DRIVER_ID;
+import static org.jboss.pnc.mavenrepositorymanager.MavenRepositoryConstants.PUBLIC_GROUP_ID;
+import static org.jboss.pnc.mavenrepositorymanager.MavenRepositoryConstants.SHARED_IMPORTS_ID;
+import static org.jboss.pnc.mavenrepositorymanager.MavenRepositoryConstants.UNTESTED_BUILDS_GROUP;
```
jdcasey: I'm not 100% sure hard coding these build groups is a good idea, but I suppose the first group after a successful build should be relatively stable. Still, it might make sense to move this into configuration so we can rename it if needed.

maven-repository-manager/src/test/java/org/jboss/pnc/mavenrepositorymanager/fixture/TestBuildExecution.java
```diff
@@ -80,33 +69,8 @@ public void setBuildContentId(String buildContentId) {
         this.buildContentId = buildContentId;
     }
 
-    @Override
-    public String getProjectName() {
-        return projectName;
-    }
-
     public void setProjectName(String projectName) {
```
jdcasey: Probably worth cleaning up these setters for the removed fields

build-coordinator/pom.xml
matejonnet: It should be used also by bpm and it is used from our internal modules too (buildCoordinator).

build-coordinator/src/main/java/org/jboss/pnc/core/builder/coordinator/bpm/BpmBuildScheduler.java
matejonnet: Proper serialization is used in `BuildExecutionConfigurationREST.toString()` method. 
Are you suggesting to not use `toString` for this purpose and rather say `toJson`?

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
matejonnet: I would leave hardcoded DRIVER_ID, agree for other to move to config.
Actually my update to use static imports was not intentional.

maven-repository-manager/src/test/java/org/jboss/pnc/mavenrepositorymanager/fixture/TestBuildExecution.java
matejonnet: +1

PR #766, opened 2016-01-20
[NCL-1525] validate input length

thescouser89:
We add the upper input limit based on the database contraints.

Also, a few validations were added.


thescouser89:
@jdg-bot :blush:

alexcreasy:
@thescouser89 looks good, merging!

PR #769, opened 2016-01-21
Fix error state handling.

matejonnet:


slaskawi:
Could we add some tests for it?

matejonnet:
There are some.

PR #775, opened 2016-01-25
Ncl 1534

vibe13:


vibe13:
retest this please

vibe13:
@matejonnet Hi Matej, mind taking a look? Thanks!

rest/src/main/java/org/jboss/pnc/rest/notifications/websockets/SessionBasedAttachedClient.java
```diff
@@ -39,8 +43,45 @@ public boolean isEnabled() {
...
+
+        // The sendText methos returns before the message is transmitted. The returned Future object is used to track the
+        // progress of the transmission. The Future's get method returns null upon successful completion. Errors in transmission
+        // are wrapped in the ExecutionException thrown when the Future object is queried.
+        Future<Void> future = session.getAsyncRemote().sendText(outputConverter.apply(messageBody));
+        
+        // wait for completion (forever, no timeout)
+        if (future.isDone()) {
+            future.get();
+        }
```
matejonnet: Returning immediately is the purpose of AsyncRemote, you can provide SendHandler as 2nd parameter if you need a callback notification when sending completes. Which would be the right way to do it IMHO.
Current implementation is the same as if BasicRemote was used, which blocks the thread.

vibe13:
@matejonnet Yep right, thanks for the review. What about this :-) ?

vibe13:
retest this please

matejonnet:
Looks ok, if you can get rid of synchronized blocks it would be even better ;)

rest/src/test/java/org/jboss/pnc/rest/notifications/websockets/DefaultNotifierTest.java
```diff
@@ -66,44 +71,67 @@ public void shouldSendAMessage() throws Exception {
...
+        // when
+        notifier.sendMessage(messageBody);
 
-        //then
-        assertThat(notifier.getAttachedClientsCount()).isEqualTo(0);
+        // then
+        verify(attachedClient, new Times(0)).sendMessage(messageBody, notifier.getCallback());
     }
 
+    // @Test
```
vibe13: @matejonnet I didn't really manage to rewrite the test below

vibe13:
@matejonnet got rid of synchronized blocks during the sendMessage, but kept when removing objects. also catched the exceptions at least to debug ConcurrentModifications. wdyt?

ui/app/common/events/events-services.js
```diff
@@ -92,14 +92,18 @@
                 payload: event.payload
               };
 
-              switch (event.payload.buildStatus) {
-                case 'REPO_SETTING_UP':
+              switch (event.payload.buildCoordinationStatus) {
```
vibe13: @matejonnet Matej, any status which you think we should ignore in UI? All these will issue a Notification ("The build... has started") or ("The build xx completed, failed, etc)

matejonnet: Leave them all in for now, it would be nice to see what is happening in the background.

rest/src/test/java/org/jboss/pnc/rest/notifications/websockets/DefaultNotifierTest.java
matejonnet: That is something @slaskawi would know better.

vibe13:
retest this please

vibe13:
@matejonnet @slaskawi  I have added a missing test for failures in notifications, so we should be good to go. Can you please take a look? Would like to merge it today as it should fix Websockets problems. Thanks!

vibe13:
retest this please

PR #777, opened 2016-01-26
Ncl 1303

slaskawi:
Hey @vibe13! Could you please have a look? This is some prep work for unifying REST interfaces (final sorting and filtering will have to occur in-memory).


vibe13:
@slaskawi sorry for the soo late reply, looks great to me :-) thanks!

PR #798, opened 2016-02-05
Use angular-loading-bar to show progress

thescouser89:
When using frontend web frameworks like Angular JS, sometimes it's
really hard to know when a page is loading after we click on a button.

For e.g if clicking the button redirects us to another page that takes a
long time to load, we don't have any visual feedback that the other page
is being loaded, prompting the user to re-press the button again.

In traditional web frameworks, our browser would show a spinner that
gives us visual feedback that it is still loading.

Using angular-loading-bar adds the visual feedback currently lacking in
our Web UI.

I've chosen to use the color orange for the bar since we don't use that color anywhere on PNC Web-UI, and is therefore easily seen, not yet too distracting.

### DIFF
--- a/ui/app/_pnc.js
+++ b/ui/app/_pnc.js
@@ -19,6 +19,7 @@
 
  (function() {
   var app = angular.module('pnc', [
+    'angular-loading-bar',
     'ui.router',
     'ui.bootstrap',
     'patternfly.notification',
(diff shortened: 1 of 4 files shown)


alexcreasy:
@thescouser89 I think this is a great idea, I can't really see the orange loading bar from the gif very well, I'll download and give it a go, but in general this is a big +1 from me.

thescouser89:
By the way, my choice of color is arbitrary! so feel free to choose a better color :)

pgier:
I think this needs a rebase

thescouser89:
Yup forgot to do it yesterday :(

PR #805, opened 2016-02-10
Ncl 1281

pgier:
Refactoring db model artifacts to have a more clear distinction between imported and built artifacts.


pgier:
retest this please

PR #808, opened 2016-02-11
[NCL-1268] Add ImportedArtifact model entity which contains download …

pgier:
…url and date

Made the Artifact entity abstract because it should never be created directly


pgier:
Ok, I guess this PR is broken

PR #822, opened 2016-02-17
[NCL-1648] Use combination of identifier and checksum when searching …

pgier:
…for matching artifacts


jbartece:
Looks good to me :+1:

PR #837, opened 2016-02-25
[NCL-1649] Allow a single BuiltArtifact to be linked to multiple buil…

pgier:
…d records

Change relation of BuildRecord to BuiltArtifact from OneToMany to ManyToMany.
If a new build generates artifacts with matching identifier and checksum
then the BuildRecord can link to the existing artifact instead of creating a
new artifact with duplicate identifiers.

### DIFF
--- a/datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
+++ b/datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
@@ -25,6 +25,7 @@ import org.jboss.pnc.model.BuildConfigurationAudited;
 import org.jboss.pnc.model.BuildRecord;
 import org.jboss.pnc.model.BuildRecordSet;
 import org.jboss.pnc.model.BuildStatus;
+import org.jboss.pnc.model.BuiltArtifact;
 import org.jboss.pnc.model.User;
 import org.jboss.pnc.spi.datastore.Datastore;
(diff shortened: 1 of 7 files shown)


pgier:
retest this please

datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
```diff
@@ -95,23 +97,48 @@ public BuildRecord storeCompletedBuild(BuildRecord.Builder buildRecordBuilder, S
...
+        for (BuiltArtifact builtArtifact : builtArtifacts) {
+            Artifact artifactFromDb = artifactRepository
+                    .queryByPredicates(withIdentifierAndChecksum(builtArtifact.getIdentifier(), builtArtifact.getChecksum()));
+            if (artifactFromDb == null) {
+                BuiltArtifact buildArtifactFromDb = (BuiltArtifact)artifactRepository.save(builtArtifact);
+                savedArtifacts.add(buildArtifactFromDb);
+            } else if (artifactFromDb instanceof BuiltArtifact) {
+                savedArtifacts.add((BuiltArtifact)artifactFromDb);
+            } else {
+                // TODO: how should we handle when a built artifact matches an existing imported artifact?
```
matejonnet: This should be an error IMO. We should not store the artifact with the same GAV.

PR #845, opened 2016-02-26
[NCL-1586] fixing TermBuildDriverRemoteTest and making it local 

michalszynkiewicz:


termd-build-driver/src/test/java/org/jboss/pnc/termdbuilddriver/TermdBuildDriverTest.java
```diff
@@ -151,4 +157,45 @@ public void shouldSetLogsInBuildExecution() throws Exception {
 //        verify(buildExecutionMock, times(1)).clearLogsWebSocketLink();
     }
 
+    @Test(timeout = 60_000)
+    public void shouldFetchFromGitAndBuild() throws Exception {
+        //given
+        Path tmpRepo = Files.createTempDirectory("tmpRepo");
+        String repoPath = "file://" + tmpRepo.toAbsolutePath().toString() + "/test-repo";
+        ZipUtils.unzipToDir(tmpRepo, "/repo.zip");
```
slaskawi: Looks good but I'm not a big fan for unzipping it here... Seems to be a bit fragile...

Would it be possible to create a directory in `src/test/resources` with almost empty `.git` and simulate that everything is there?

slaskawi: Clarified - we can't use a synthetic git repo here.

slaskawi:
:+1

termd-build-driver/src/test/java/org/jboss/pnc/termdbuilddriver/TermdBuildDriverTest.java
matejonnet: Do we need to test git at all? It is important that we test communication with BuildAgent. From a BuildDriver perspective it should be the same if you run `git clone` or `whoami`, important is that command is executed and result received.

PR #849, opened 2016-02-28
[NCL-1350]  improve local dev env and test setup

psakar:
prevent failure of mvn compile without clean
fix generate-test-resources configuration
remove obsolete file
allow to create aquillian Deployment for integration test from build target
postpone preparation of application server to pre-integration-test phase


matejonnet:
It looks ok to me. @slaskawi any comments ?

PR #859, opened 2016-03-02
Add tooltips and total page number to pagination directive

vibe13:
@alexcreasy Thought that adding the total number of pages would be good, plus tooltips
pagination


alexcreasy:
@vibe13  +1!!

vibe13:
@alexcreasy  thanks!

PR #861, opened 2016-03-02
NCL-1572 Fixes UX issues and mitigates performance issues for user, but does not solve underlying scalability issues.

alexcreasy:


alexcreasy:
@vibe13 Could you give this a review? This doesn't solve the scaling issues (it will improve it on some pages though) but it fixes the UX issues where the panels become empty while refreshing and the user therefore can't use them if they're refreshing regularly. 

I have a fix coming for the performance issues but they have complex logic (to check whether a build should be displayed on the current page and where, there are a surprising number of edge cases to this), but I wanted to have something better to revert to if we can't get them stable in time.

alexcreasy:
retest this please

PR #876, opened 2016-03-04
[NCL-1685] storing results of a failed build

michalszynkiewicz:


build-coordinator/src/main/java/org/jboss/pnc/core/builder/datastore/DatastoreAdapter.java
```diff
@@ -106,11 +106,41 @@ public void storeResult(BuildTask buildTask, BuildResult buildResult) throws Dat
             log.debugf("Storing results of buildTask [%s] to datastore.", buildTask.getId());
             datastore.storeCompletedBuild(buildRecordBuilder, buildTask.getBuildRecordSetIds());
         } catch (Exception e) {
-            storeResult(buildTask, e);
+            storeErroneousResult(buildTask, e);
+        }
+    }
+
+
+    public void storeFailedResult(BuildTask buildTask, BuildResult buildResult) throws DatastoreException {
```
matejonnet: Does it make sense to have another method for storing the failed build? 
Isn't the result in both cases mostly the same?

michalszynkiewicz: There's indeed some code duplucation. I'll reduce it. For now I think that two methods are clearer, but I'll try out merging too. (after some tests start to show that it works ;) )

Thank you for reviewing the changes!

matejonnet: Once you merge in public method others can use it and "for now" statement turns in long removal procedure.
I also wonder if changes on coordinator side are required for this bug fix?

michalszynkiewicz: Do you mean BuildCoordinator class itself or do you count DatastoreAdapter as a coordinator side too?
If it's the latter - yes they are required.

michalszynkiewicz: Here a version with one method for storing failed and successful builds:

michalszynkiewicz:
replaced with: https://github.com/project-ncl/pnc/pull/890

PR #882, opened 2016-03-07
warning if command contains mvn and not deploy

matedo1:


alexcreasy:
Nice, clean solution +1 for merging.

PR #897, opened 2016-03-10
[NCL-1706] sealing removing submitted task

michalszynkiewicz:
An attempt to fix build retrying for twitter4j

### DIFF
--- a/build-coordinator/src/main/java/org/jboss/pnc/core/builder/coordinator/BuildCoordinator.java
+++ b/build-coordinator/src/main/java/org/jboss/pnc/core/builder/coordinator/BuildCoordinator.java
@@ -217,7 +217,6 @@ public class BuildCoordinator {
                         datastoreAdapter.storeResult(buildTask, buildResult);
                         coordinationStatus = BuildCoordinationStatus.DONE_WITH_ERRORS;
                     } else {
-                        removeSubmittedTask(buildTask);
                         throw new BuildCoordinationException("Failed task should have set exception or failed reason status.");
                     }
                 } else {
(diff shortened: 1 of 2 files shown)


build-coordinator/src/main/java/org/jboss/pnc/core/builder/coordinator/BuildCoordinator.java
```diff
@@ -217,7 +217,6 @@ void processBuildTask(BuildTask buildTask) {
                         datastoreAdapter.storeResult(buildTask, buildResult);
                         coordinationStatus = BuildCoordinationStatus.DONE_WITH_ERRORS;
                     } else {
-                        removeSubmittedTask(buildTask);
                         throw new BuildCoordinationException("Failed task should have set exception or failed reason status.");
```
matejonnet: This can be optimized further, It does not make sense to throw an exception and catch it few lines bellow.

matejonnet:
As this doesn't look a fix for something, please merge it to master only.

matejonnet:
Ok to put it in 0.8 as it fixes what you pointed out, a call of removeSubmittedTask in case of any uncaught exception.

rest-model/src/main/java/org/jboss/pnc/rest/utils/JsonOutputConverterMapper.java
```diff
@@ -47,6 +50,7 @@ public static String apply(Object objectToBeConverted) {
             try {
                 return mapper.writeValueAsString(objectToBeConverted);
             } catch (JsonProcessingException e) { //TODO remove exception masking
+                log.warn("Could not convert object to JSON", e);
```
michalszynkiewicz: @Matej: What do you think about logging the object here too?
The only cons that I can think of here is that we can have some sensitive data that might end up in the log in this case. What is your opinion?

matej: I'm probably not the one you're looking for. ;)

michalszynkiewicz: true :D editing the comment.

PR #914, opened 2016-03-18
NCL-1734 proper copying of auth web.xml

michalszynkiewicz:


rest/pom.xml
```diff
@@ -202,13 +202,12 @@
     </plugins>
   </build>
 
-  <profiles>    
+  <profiles>
      <profile>
         <id>auth</id>
         <activation>
             <property>
                 <name>auth</name>
```
michalszynkiewicz: with this config, if we build without `-Dauth`, no auth will be set up. But if we use `-Dauth`, either **with or without value**, **rest will be built with auth**. So even if someone uses `-Dauth=false`, auth will be enabled.

@matejonnet is this ok for us?

matejonnet: Shouldn't be the profile activation managed by -P flag? Do we need a SystemProperty -Dauth somewhere else? If yes, can we set it in the profile?

michalszynkiewicz: Before my changes we relied on this property. It's also used for tests to determine if they should run with auth or not. 

I may try to change it. So far I didn't find a way to set a property in a profile.

matejonnet: Then leave as is (-D) for 0.8 and try to make it work with -P for the next release.

PR #929, opened 2016-03-30
Upgrade Indy dep, and seal the tracking record after the build completes

jdcasey:
Sealed to ensure it's preserved on disk for future use, and to enable retrieval of the tracking record.


jdcasey:
[NCL-1635]

jbartece:
I think we should not merge it into the 0.8 branch

jdcasey:
Sounds fine. I'll close this, then.

PR #930, opened 2016-03-30
Upgrade Indy version, and seal the tracking record after the build completes

jdcasey:
Seal tracking record to ensure it's preserved on disk for future use, and to make it available for download.


jdcasey:
[NCL-1635] This is required before we can upgrade to Indy 0.99.2

maven-repository-manager/pom.xml
```diff
@@ -47,6 +47,10 @@
       <artifactId>spi</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.jboss.logging</groupId>
+      <artifactId>jboss-logging</artifactId>
+    </dependency>
+    <dependency>
```
matejonnet: Can you switch to slf4j, to slowly unify on one logging.

jdcasey: I've only included that to get the embedded Indy instance to run during tests. It's needed by Infinispan (I actually use logback + slf4j-api in Indy itself)

matejonnet: Test scope ?

jdcasey: it's in provided scope, in dependencyManagement below. It may make sense to put it in test scope, but I'm not 100% sure. Provided scope will not propagate into the dependency graph though.

matejonnet: I thought of "test" scope to prevent accidental use in our production code.

jdcasey: Okay, I've moved it to test scope and verified that the tests succeed (in this module, at least). I'm still having trouble with later modules in the build that I believe are unrelated (they haven't ever worked on this machine).

jbartece:
retest this please

jbartece:
retest this please

jbartece:
@jdcasey Tests are still failing. Probably needs some fix.

jdcasey:
The problem was a hibernate-search-orm dependency in the folo (tracking add-on) model jar's classpath, which was bringing in jboss-logging. Hibernate was only used for annotations in this module, so I turned that scope to provided and it's working on my localhost.

I'll cut a new Indy micro release today and update this PR when it's ready.

jdcasey:
@matejonnet @jbartece Okay, tests are passing. Any objections to me merging this?

jbartece:
:+1:

jdcasey:
The new version of Indy is 0.99.2.1.

PR #931, opened 2016-03-31
[NCL-1803] storing build completion on incomplete request

michalszynkiewicz:


rest/src/main/java/org/jboss/pnc/rest/endpoint/BuildTaskEndpoint.java
```diff
@@ -90,11 +95,25 @@ public Response buildTaskCompleted(
...
+        } else {
+            bpmCompleteListener.notifyCompleted(taskId, buildResult.toBuildResult());
+        }
         return Response.ok().build();
     }
 
+    private BuildResult erroneousBuildResult() {
+        return new BuildResult(Optional.empty(),
+                Optional.empty(), Optional.empty(), Optional.empty(),
+                Optional.of(BuildExecutionStatus.DONE_WITH_ERRORS));
```
matejonnet: In case of invalid request BuildExecutionStatus.SYSTEM_ERROR fits better.

PR #936, opened 2016-04-04
[NCL-1827] Delete group (merged) metadata for promoted artifacts

jdcasey:
both promotions to shared-imports and build output.


jdcasey:
@matejonnet I'm having the same problem with this build. I'm seeing a failure in the docker-environment-driver locally, and when I comment out that module, I get a ConcurrentModificationException in the build-executor tests.

matejonnet:
We should remove docker environment driver from the default build as we don't use it (NCL-1743).

PR #937, opened 2016-04-05
[NCL-1770] Remove BuildRecordSet

pgier:
Make a direct one to many relation between product milestone and build record
Only allow a build configuration to be associated with a single product version at a time


pgier:
This still needs more work before merging

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/datastore/DatastoreAdapter.java
```diff
@@ -107,11 +107,8 @@ public void storeResult(BuildTask buildTask, BuildResult buildResult) throws Dat
                 return;
             }
 
-            // Build driver results
-            buildRecordBuilder.endTime(Date.from(Instant.now()));
```
matejonnet: We don't need to set endTime ?

matejonnet:
Nice, it looks like this updated simplifies many things.

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/datastore/DatastoreAdapter.java
pgier: The end time is now set in initBuildRecordBuilder method.  Reviewing this, it looks like setting the endTime should also be removed from storeResult with an error and storeRejected since they are now redundant.

matejonnet: There is also duplication in the `initBuildRecordBuilder` itself.
I wonder if it makes sense to use `endTime` from the `buildTask` or just use `Instant.now` as when we store the result the operation is completed.

pgier:
retest this please

PR #939, opened 2016-04-06
NCL-1769 product import start and finish issues

matedo1:


ui/bower.json
```diff
@@ -9,7 +9,7 @@
     "font-awesome": "4.2.0",
     "jquery": "2.1.3",
     "angular-resource": "1.3.15",
-    "angular-bootstrap": "0.13.0",
```
alexcreasy: Out of interest, what does upgrading the angular-bootstrap version fix?

matedo1: There is new isOpen property for tooltips, you can check it here https://github.com/angular-ui/bootstrap/blob/master/CHANGELOG.md#features-11 , I used this new property on several places like product-import-tree.html:53

alexcreasy: ahh, yes that is useful!

alexcreasy:
+1 merging!

PR #940, opened 2016-04-08
[NCL-1754] fixed Last build status in the error log

michalszynkiewicz:


build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/datastore/DatastoreAdapter.java
```diff
@@ -72,7 +73,9 @@ public BuildConfigSetRecord saveBuildConfigSetRecord(BuildConfigSetRecord buildC
      * @return The latest revision of the given build configuration
      */
     public BuildConfigurationAudited getLatestBuildConfigurationAudited(Integer buildConfigurationId) {
-        return datastore.getLatestBuildConfigurationAudited(buildConfigurationId);
+        BuildConfigurationAudited buildConfigAudited = datastore.getLatestBuildConfigurationAudited(buildConfigurationId);
+        buildConfigAudited.getProject().getBuildConfigurations().forEach(BuildConfiguration::getId);
```
matejonnet: Why is this loop required here ?

michalszynkiewicz: I've just tested it. 

`getBuildConfigurations()` is not enough, `LazyInitializationException` is still thrown on accessing the `buildConfigurations` of `Project`.

Long story: AFAIR, we put the stuff that is fetched here in `submittedBuilds`. Then, later, handling a different request, we take a build from it and try to get build configurations from the build.

matejonnet: @pgier @vibe13 have dealt with some LazyIE we already had.

michalszynkiewicz: @matejonnet I don't understand. Are their changes in some other PRs and I can rebase on them?

vibe13: @michalszynkiewicz @matejonnet  If we are going to keep the LAZY fetching, I saw an earlier implementation. It basically does not get rid of the loop, but it's inside a dedicated class:  pnc/rest/src/main/java/org/jboss/pnc/rest/utils/HibernateLazyInitializer.java used inside  pnc/rest/src/main/java/org/jboss/pnc/rest/trigger/BuildTriggerer.java

PR #941, opened 2016-04-11
NCL-1848

thauser:
Operations which return Artifacts should return ArtifactPage, not BuildRecordPage.


pgier:
retest this please

PR #953, opened 2016-04-20
Fix for NCL-1878

emmettu:
Remove hard-coding of the 409 response code error message. This allows the correct error message from the REST API to be displayed.

### DIFF
--- a/ui/app/services.js
+++ b/ui/app/services.js
@@ -208,9 +208,6 @@
             case 404:
               Notifications.error('Requested resource not found');
               break;
-            case 409:
-              Notifications.error('Build rejected because the same build configuration is already running', rejection);
-              break;
             default:
               handleError(rejection);
               break;


jbartece:
@michalszynkiewicz  Could you review this PR, please?

ui/app/services.js
```diff
@@ -208,9 +208,6 @@
             case 404:
               Notifications.error('Requested resource not found');
               break;
-            case 409:
```
michalszynkiewicz: After removal of this case branch, will the proper error message show up?

I'm all hands for it. My only concern is that it may result in appropriate messages not showing up if backend returns something that UI is not able to parse and display. And I don't see any backend changes here.

emmettu: The correct error messages display now. With this 409 switch case in place the UI always returns the "build still running" error message even though 409 is used to notify the user of conflicts that are not necessarily a result of a running build, such as trying to create a configuration with the same name as an existing one.

This didn't require any back end changes because the UI was ignoring the true error message and returning its own, the REST API was doing the correct thing.

All the endpoints that return a 409 look good to me, the UI should be able to parse them.

michalszynkiewicz: Cool :+1: 
I'm merging.

PR #957, opened 2016-04-26
NCL-1922

emmettu:
Map BuildConflictExceptions to error responses. This is done by adding a new the BuildConflictExceptionMapper, and adding a new constructor for ErrorResponseRest that takes BuildConflictExceptions. Lastly, the try-catch block that catches BuildConflictExceptions in the trigger method of BuildConfigurationEndpoint has been removed so the BuildConflictException can be thrown to the mapper.

### DIFF
--- a/rest/src/main/java/org/jboss/pnc/rest/endpoint/BuildConfigurationEndpoint.java
+++ b/rest/src/main/java/org/jboss/pnc/rest/endpoint/BuildConfigurationEndpoint.java
@@ -152,8 +152,8 @@ public class BuildConfigurationEndpoint extends AbstractEndpoint<BuildConfigurat
 
     @ApiOperation(value = "Creates a new Build Configuration")
     @ApiResponses(value = {
-            @ApiResponse(code = INVLID_CODE, message = INVALID_DESCRIPTION, response = BuildConfigurationSingleton.class),
             @ApiResponse(code = SUCCESS_CODE, message = SUCCESS_DESCRIPTION, response = BuildConfigurationSingleton.class),
+            @ApiResponse(code = INVLID_CODE, message = INVALID_DESCRIPTION, response = ErrorResponseRest.class),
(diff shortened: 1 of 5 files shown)


rest/src/main/java/org/jboss/pnc/rest/endpoint/BuildConfigurationEndpoint.java
```diff
@@ -152,8 +152,8 @@ public Response getAll(@ApiParam(value = PAGE_INDEX_DESCRIPTION) @QueryParam(PAG
 
     @ApiOperation(value = "Creates a new Build Configuration")
     @ApiResponses(value = {
-            @ApiResponse(code = INVLID_CODE, message = INVALID_DESCRIPTION, response = BuildConfigurationSingleton.class),
             @ApiResponse(code = SUCCESS_CODE, message = SUCCESS_DESCRIPTION, response = BuildConfigurationSingleton.class),
+            @ApiResponse(code = INVLID_CODE, message = INVALID_DESCRIPTION, response = ErrorResponseRest.class),
```
michalszynkiewicz: can you fix type in INVLID_CODE btw?

emmettu: Whoops!

michalszynkiewicz: It's not your typo. It seems it's been there for a long time:)

emmettu: Yeah the replace is 111 lines, should I make a new PR for this or just commit it here?

michalszynkiewicz: Maybe separate then :D
27 kwi 2016 17:03 "Emmett Underhill" <notifications@github.com> napisał(a):

michalszynkiewicz:
:+1:

PR #958, opened 2016-04-27
NCL-910 build input queue

michalszynkiewicz:


build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildCoordinator.java
```diff
@@ -203,113 +157,149 @@ private boolean rejectAlreadySubmitted(BuildTask buildTask) {
...
+            taskSet.getBuildTasks().stream()
+                    .filter(t -> isDependentOn(failedTask, t))
+                    .forEach(t -> finishDueToFailedDependency(failedTask, t));
         }
-        return submittedBuildTasks.peek() != null;
     }
 
-    private boolean isBuildAlreadySubmitted(BuildTask buildTask) {
-        return submittedBuildTasks.contains(buildTask);
+    private boolean isDependentOn(BuildTask failedTask, BuildTask t) {
```
matejonnet: Could you add some api doc and maybe use some more descriptive parameter name instead of `t`.

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildQueue.java
```diff
@@ -0,0 +1,210 @@
...
+ * Author: Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
+ * Date: 4/18/16
+ * Time: 12:47 PM
+ */
+@ApplicationScoped
+public class BuildQueue {
+
+    private final Logger log = LoggerFactory.getLogger(BuildQueue.class);
+
+    private static final int MAX_CONCURRENT_BUILDS = 20; // TODO: move to property ?
```
matejonnet: Yes, make it configurable please.

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildQueue.java
```diff
@@ -0,0 +1,210 @@
...
+    public static final ExecutorService executorService = Executors.newFixedThreadPool(10);
```
matejonnet: This one should be configurable too.

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildQueue.java
```diff
@@ -0,0 +1,210 @@
...
+    public String toString() {
+        return "BuildQueue{" +
+                "readyTasks=" + readyTasks +
+                ", waitingTasks=" + waitingTasks +
+                ", tasksInProgress=" + tasksInProgress +
+                ", taskSets=" + taskSets +
+                '}';
+    }
+
+    @Deprecated // used only in tests!
```
matejonnet: If it's possible it would be nice to get rid of this as it a sign of a design issue.

build-coordinator/src/test/java/org/jboss/pnc/coordinator/test/ProjectWithDependenciesBuildTest.java
```diff
@@ -55,20 +57,24 @@
     Set<BuildSetStatusChangedEvent> eventsReceived = new HashSet<>();
 
     private void collectEvent(BuildSetStatusChangedEvent buildSetStatusChangedEvent) {
+        log.info("############Got: build set statuc changed event for build id: " + BUILD_SET_ID);
```
matejonnet: Remove extra markers `#`. If this message is more important than others, try to lower the "level" of others.

matejonnet:
I find a bit confusing the aggregation of four collections in one `BuildQueue` object, that could cause us problems when we want to make is clusterable, but I guess for now we are ok.
If it is not too complicated it would be nice to remove a two way dependency between `BuildCoordinator` and `BuildQueue`.

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildQueue.java
michalszynkiewicz: actually, the method can be used not only in tests. It just checks if the queue is empty.

I think is worth having this check in tests. What would you say for removing Deprecated?

matejonnet: +1 to remove the deprecated

build-coordinator/src/test/java/org/jboss/pnc/coordinator/test/ProjectWithDependenciesBuildTest.java
michalszynkiewicz: done

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildCoordinator.java
michalszynkiewicz: done

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildQueue.java
michalszynkiewicz: done

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildQueue.java
michalszynkiewicz: done

PR #966, opened 2016-05-09
NCL-1437 remove build-coordinator-mock module

tremes:


matejonnet:
It looks ok to me. @michalszynkiewicz wdyt?
There is also `BuildExecutorMock` that should be moved.

tremes:
Thank's for checking. Yes I know...I am looking at this.

spi/src/main/java/org/jboss/pnc/spi/coordinator/BuildSetTask.java
```diff
@@ -0,0 +1,71 @@
...
+import org.jboss.pnc.spi.events.BuildSetStatusChangedEvent;
+
+import java.util.Date;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * @author Tomas Remes
+ */
+public interface BuildSetTask {
```
michalszynkiewicz: Is this interface required?
Did you have some problems with mocking w/o it?

spi/src/main/java/org/jboss/pnc/spi/coordinator/BuildTask.java
```diff
@@ -0,0 +1,91 @@
...
+import org.jboss.pnc.model.BuildConfiguration;
+import org.jboss.pnc.model.BuildConfigurationAudited;
+import org.jboss.pnc.model.ProductMilestone;
+import org.jboss.pnc.model.User;
+import org.jboss.pnc.spi.BuildCoordinationStatus;
+
+/**
+ * Created by <a href="mailto:matejonnet@gmail.com">Matej Lazar</a> on 2014-12-23.
+ */
+public interface BuildTask {
```
michalszynkiewicz: ditto

pnc-mock/src/main/java/org/jboss/pnc/mock/coordinator/BuildCoordinatorMock.java
```diff
@@ -72,17 +73,20 @@ public BuildTask build(BuildConfiguration buildConfiguration, User user, boolean
...
         logger.warn("Invoking unimplemented method build");
         return Mockito.mock(BuildSetTask.class);
     }
 
-    @Override
     public List<BuildTask> getSubmittedBuildTasks() {
         return activeTasks;
     }
 
+    @Override
```
michalszynkiewicz: why do you remove @override in other places and keep it here?

spi/src/main/java/org/jboss/pnc/spi/coordinator/BuildSetTask.java
tremes: I think so (or simply didn't come up with any other better idea). `BuildCoordinator` interface defines `build` methods which has this return type and when you don't want to introduce circular dependency by returning the original `BuildTask` then you need this interface (and the one below) in spi...I guess.

pnc-mock/src/main/java/org/jboss/pnc/mock/coordinator/BuildCoordinatorMock.java
tremes: Yes the override should stay at other methods as well. I don't know why I removed it.

spi/src/main/java/org/jboss/pnc/spi/coordinator/BuildSetTask.java
michalszynkiewicz: @matejonnet can we move BuildTask and BuildSetTask to spi? It seems we could omit creating these two interfaces this way.

tremes: I guess we would need to move `DefaultBuildStatusChangedEvent` and then it could be possible. Should I give it a try?

matejonnet: Moving them to SPI makes sense, but we must make them clear DTO first. Now there are some methods with business logic like `setStatus`, `taskStatusUpdatedToFinalState`.

michalszynkiewicz: @tremes can you add a todo on BuildTask and BuildSetTask to slim them down, move it and remove the interfaces?

Other than that, this change looks good to me :)

tremes:
Ok so I added second commit to make `BuildTask` and `BuildSetTask` DTOs. It's not masterpiece so it will probably need some updates but it seems to work. :-)

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/DefaultBuildCoordinator.java
```diff
@@ -187,17 +245,23 @@ private void processBuildTask(BuildTask task) {
                         task.getId(), task.getRebuildAll(), prepareBuildTaskFilterPredicate().test(task), task.getBuildConfigurationAudited().getIdRev());
                 if(!task.getRebuildAll() && prepareBuildTaskFilterPredicate().test(task)) {
                     log.info("[{}] Marking task as REJECTED_ALREADY_BUILT, because it has been already built", task.getId());
-                    task.setStatus(BuildCoordinationStatus.REJECTED_ALREADY_BUILT);
+                    notifyBuildStatusChanged(task, BuildCoordinationStatus.REJECTED_ALREADY_BUILT);
+                    task.setHasFailed(BuildCoordinationStatus.REJECTED_ALREADY_BUILT.hasFailed());
```
michalszynkiewicz: I think you may leave setting hasFailed in the the build task

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/DefaultBuildCoordinator.java
```diff
@@ -147,8 +151,8 @@ public BuildSetTask build(BuildConfigurationSet buildConfigurationSet, User user
                 rebuildAll,
                 getBuildStatusChangedEventNotifier(),
                 () -> datastoreAdapter.getNextBuildRecordId());
-        BuildSetStatusChangedEvent event = buildSetTask.setStatus(BuildSetStatus.NEW);
-        buildSetStatusChangedEventNotifier.fire(event);
+        notifyBuildSetStatusChanged(buildSetTask, BuildSetStatus.NEW);
```
michalszynkiewicz: it might be more readable if notifyBuildSetStatusChanged was called e.g. updateBuildSetStatus and contained setStatus

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/DefaultBuildCoordinator.java
```diff
@@ -187,17 +245,23 @@ private void processBuildTask(BuildTask task) {
...
                 task.setStartTime(new Date());
-                task.setStatus(BuildCoordinationStatus.BUILDING); //status must be updated before startBuild as if build takes 0 time it complete before having Building status.
+                notifyBuildStatusChanged(task,BuildCoordinationStatus.BUILDING);
+                task.setStatus(BuildCoordinationStatus.BUILDING);//status must be updated before startBuild as if build takes 0 time it complete before having Building status.
+                task.setHasFailed(BuildCoordinationStatus.BUILDING.hasFailed());
             }
             buildScheduler.startBuilding(task, onComplete);
         } catch (CoreException | ExecutorException e) {
             log.debug(" Build coordination task failed. Setting it as SYSTEM_ERROR.", e);
+            notifyBuildStatusChanged(task,BuildCoordinationStatus.SYSTEM_ERROR);
```
michalszynkiewicz: The same as with `notifyBuildSetStatusChanged` - I think it would be better to have all the methods that are executed when the status is changed called from one place.

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/DefaultBuildCoordinator.java
tremes: Well yes I've had something like you suggested but if it contains `setStatus` then it will need to return `BuildTask`. And yes we can add overloaded version of this method to accept some String as status description as well. Don't know what's more readable...

spi/src/main/java/org/jboss/pnc/spi/coordinator/BuildTask.java
```diff
@@ -67,7 +64,7 @@
 
     private ProductMilestone productMilestone;
 
-    private boolean hasFailed = false;
+    private AtomicBoolean hasFailed = new AtomicBoolean();
```
michalszynkiewicz: If we need `AtomicBoolean` here, don't we need `AtomicReference` for status?

michalszynkiewicz: Why do you need `AtomicBoolean`?

tremes: Yes it's not needed. I will keep this hasFailed as it was.

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildQueue.java
```diff
@@ -188,20 +184,10 @@ public BuildTask take() throws InterruptedException {
     }
 
 
-    private boolean isBuildAlreadySubmitted(BuildTask buildTask) {
+    public synchronized boolean isBuildAlreadySubmitted(BuildTask buildTask) {
```
matejonnet: Why does it have to be synchronized ?

tremes: I think `waitingTasks` is not thread safe as ArrayList.

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/DefaultBuildCoordinator.java
michalszynkiewicz: I don't understand. Why does it have to return anything? BuildTask is mutable...

if status desc is always set, I'd add it too. Then we have all things that happen when status is changed in one place and we're sure we won't forget to fix some if there's some bug in it.

tremes: Disregard my comment. I was confused by my local changes. Sorry. Will update.

michalszynkiewicz:
:+1:  Looks good to me.

You should probable squash commits. Right, @matejonnet ?

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildQueue.java
matejonnet: But there are read-only methods used only?

matejonnet:
As you prefer, I usually don't squash them unless there is some ugly stuff in-there ;)

PR #969, opened 2016-05-10
NCL-1886 Improve the user friendliness of Product import process

matedo1:


ui/app/styles/directives.css
```diff
@@ -120,4 +120,24 @@
...
+    color: #3f9c35;
+}
+.bc-tree .node-warning {
+    color: #ec7a08;
+}
+.bc-tree .node-critical {
+    color: #cc0000;
+}
+
+.bc-tree .node-selected .node-sucess,
```
alexcreasy: I'm just wondering, could we use the patternfly built in css classes ".text-success .text-warning .text-danger" rather than custom css classes? This would ensure we match up with the patternfly colour scheme if it changes slightly for example.

matedo1: +1, I created new commit for this

alexcreasy:
Great thanks for the changes, I'll merge this now.

PR #974, opened 2016-05-13
NCL-1965 marking build as system error on Error

michalszynkiewicz:
1. proper logging on error when build ends with error (unlike exception, error is rethrown)
1. fixed synchronization of buildQueue

### DIFF
--- a/build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildQueue.java
+++ b/build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildQueue.java
@@ -40,17 +40,20 @@ import java.util.stream.Collectors;
 /**
  * <h3>Build task queue.</h3>
  *
- * The queue consists of 4 collections:
+ * The queue consists of 5 collections:
  * <ul>
  * <li>taskSets - set of currently processed task sets</li>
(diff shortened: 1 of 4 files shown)


michalszynkiewicz:
@tremes @matejonnet can you take a look?

tremes:
IMO introducing next collection in this class doesn't make understanding to this logic easier but that's just minor. Another "cosmetic" are imports in BuildConfiguration class (I know there is an issue to provide code templates.:-)). Otherwise +1.

michalszynkiewicz:
I didn't have any other idea to fix the race condition between `isBuildAlreadySubmitted` and `take`...

With imports I thought we're trying not to use package improts, so I configured idea this way and it did that...

matejonnet:
I also have thoughts about 4 collections. Maybe we should take some other approach as once we make this clusterable the synchronized methods won't fit in very nice.

Imports +1 for not using packet imports.

michalszynkiewicz:
Maybe we don't need 100% thread safety here?

If it's ok for isBuildAlreadySubmitted and getSubmittedBuildTask to be just more or less thread-safe, the code can be simplified and we could remove `synchronized`.

Here's a version without `synchronized`: https://github.com/project-ncl/pnc/pull/984

PR #1009, opened 2016-06-02
NCL-1956 Use separate deployment URL to deploy directly to hosted repo

pkocandr:


thescouser89:
retest this please

pkocandr:
retest this please

jdcasey:
@pkocandr +1 from me, but do we know what needs to change in terms of how the settings.xml gets generated? It may need to be updated...?

pkocandr:
I checked that before and this should be the only necessary change being made, because the settings.xml template already contains altDeploymentRepsoitory with value "aprox::default::${AProxDeployUrl}". The placeholder is replaced by the getDeployUrl() result.

PR #1012, opened 2016-06-03
Ncl 2033

pgier:
Refactor model tests
Fix bug where a duplicate artifact can be added to a build record


build-coordinator/src/test/java/org/jboss/pnc/coordinator/test/MockUser.java
```diff
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-package org.jboss.pnc.model.mock;
+package org.jboss.pnc.coordinator.test;
```
matejonnet: Can we move it to pnc-mock module ?

pgier: Ok, I moved it to the pnc.mock.model package in the mock module.

PR #1014, opened 2016-06-06
Add title when action button is disabled

matedo1:


matedo1:
Hi @alexcreasy , I did small improvement for "Disable action buttons when not logged in"

alexcreasy:
@matedo1 Did this work for you? I originally tried adding a title to the element, but apparently most browsers won't display this for a disabled element? That said I only tested in chrome. I'm more than happy to merge this if you got it working!

alexcreasy:
@matedo1 again sorry for the delay, merging now!

PR #1015, opened 2016-06-06
Use field injection and flush for stream in UIConfigurationServlet

jbartece:
@matedo1 @matejonnet


jbartece:
Matej reviewed it, I'm merging it

PR #1019, opened 2016-06-07
NCL-1343 Implement validation for scm url when using web

matedo1:


alexcreasy:
@matedo1 great, yes, I think this is the best idea for UI SCM field validation yet. If people complain too much about the lack of immediate feedback we can look to change this later, but disallowing valid SCM URLs will cause serious problems.

PR #1025, opened 2016-06-08
[NCL-2083] Add validation of artifact data returned from repo manager

pgier:


maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/MavenRepositorySession.java
```diff
@@ -79,6 +84,9 @@
     private final RepositoryConnectionInfo connectionInfo;
     private boolean isSetBuild;
 
+    @Inject
```
michalszynkiewicz: Is there any situation in which this @Inject works?

pgier: I thought it would inject the validator when running in container, but I could be wrong.

PR #1052, opened 2016-06-22
NCL-2105 Allow only https:// and git:// scm urls as input for cloning…

matedo1:
… in DA


alexcreasy:
This is fantastic, great that it integrates with ng-modal. Merging

PR #1055, opened 2016-06-23
[NCL-2144] keep the build alive after failure

michalszynkiewicz:
1. added a UI option: "Build and keep alive after failure" (in the same menu where "Build forcing all rebuilds" is). If the option is selected, it also forces rebuilds
1. added a query param for `/build-configurations/{id}/build`: `keepPodAliveOnFailure`. The parameter is independent of `rebuildAll` - the two params are passed together if user selects "Build and keep alive after failure"
1. Environment destroying is skipped when `keepPodAliveOnFailure` is enabled


matejonnet:
It looks ok. +1 for merging.

PR #1060, opened 2016-06-24
[NCL-2168] Make build agent route configurable

thescouser89:
Changes made:

- Add 'buildAgentHost' variable into openshift driver configuration
- Use that new variable to setup routes
- Use that new variable to setup the service name


thescouser89:
build build build
pass pass pass

PR #1061, opened 2016-06-24
[NCL-2168] Don't specify host in route-name

thescouser89:
We shouldn't do that since Openshift will complain that the new route
name does not conform to the DNS convention. We are going to revert that
change


thescouser89:
compile compile compile

PR #1062, opened 2016-06-24
[NCL-2168] Fix internal path for pod

thescouser89:


thescouser89:
Woops I didn't fully understand what was going on, this commit is wrong.

PR #1071, opened 2016-07-01
[NCL-2204] Change DNS policy of build agent pod

thescouser89:
The current DNS policy for the build agent pod created by PNC on
Openshift is Default.

This causes us issues since the 'Default' policy doesn't add the
internal DNS nameserver of Openshift inside the build agent pod. We
would really like to have the internal DNS nameserver to solve NCL-2201.

Using the dns policy of 'ClusterFirst' should add to `/etc/resolv.conf`
the internal DNS nameserver of Openshift


thescouser89:
@matejonnet Could you review this PR? This is part of NCL-2201

thescouser89:
Note: This commit, if merged, should also be backported to v0.9 branch

PR #1072, opened 2016-07-01
[NCL-2204][v0.9] Change DNS policy of build agent pod

thescouser89:
The current DNS policy for the build agent pod created by PNC on
Openshift is Default.

This causes us issues since the 'Default' policy doesn't add the
internal DNS nameserver of Openshift inside the build agent pod. We
would really like to have the internal DNS nameserver to solve NCL-2201.

Using the dns policy of 'ClusterFirst' should add to `/etc/resolv.conf`
the internal DNS nameserver of Openshift


thescouser89:
Note: already commited to master branch: #1071 

Just creating another PR so as to remember to backport it to v0.9 when we confirm the fix works on master branch.

If we prefer to cherry-pick from master to v0.9 branch, please just close the request

PR #1073, opened 2016-07-01
NCL-2112 Loading bar for specific socket communication

matedo1:


matedo1:
retest this please

ui/app/common/daclient/jsonrpc.js
```diff
@@ -20,158 +20,182 @@
...
-          $log.debug('Making RPC request: ' + JSON.stringify(request, null, 2));
+            $log.debug('Making RPC request: ' + JSON.stringify(request, null, 2));
 
-          return deferred.promise;
-        }
-
-        function handleRpcResponse(response) {
-          $log.debug('Received RPC response: ' + JSON.stringify(response, null, 2));
+            that.interceptors.forEach(function(interceptor) {
+              interceptor().requestStarted(deferred.promise, cfpLoadingBar);
```
alexcreasy: I wouldn't inject the cfpLoading bar here, this tightly couples the interceptor logic. I would just inject the promise for each interceptor and access the loading bar where you create the interceptor itself. I'll write a general comment with how I would achieve this.

alexcreasy:
I would make a few changes here just to make the interceptor code more generic:

- Each interceptor should be a named angular service/factory etc with the required methods, rather than an anonymous function.
- Doing it this way allows you to inject the cfpLoading bar into the specific interceptor you want to use it in, rather than injecting it into all the interceptors.
- To achieve this: 
  1) in the config block, simply push the string name of the angular service to the interceptors array instead of an anonymous function.
  2) In the jsonrpc client you can use the $injector service when you iterate over the interceptors array calling $injector.get(interceptor) to get the instance of the specific interceptor (Use a try catch block and rethrow a better, more specific exception than the obscure $injector failed one you'll get).

alexcreasy:
An alternative way if you prefer to pass the anonymous function instead (I've never tried this but a quick look at the docs tells me it's possible) is to iterate over the interceptors array and call $injector.invoke on the function, that will allow you to provide apply dependency injection to function.

alexcreasy:
@matedo1 👍  merging.

PR #1088, opened 2016-07-19
Merge keycloak.json to pnc-config.json file.

tremes:
This is going to fail since it seems that Indy needs to be updated to use keycloak 1.9.8.Final as well (NCL-2246) and the pull request CI build needs an update as welll. @matejonnet @michalszynkiewicz please review and maybe it would be also good if someone from keycloak team could review as well. 

I think the next step could be to store these values directly in standalone.xml but I don't know the way how to obtain them from standalone.xml yet and question is whether it has some advantage over the current config approach.


auth/src/main/resources/pnc-config.json
```diff
@@ -5,7 +5,10 @@
             "@module-config":"authentication-config",
             "username":"${env.PNC_EXT_OAUTH_USERNAME}",
             "password":"${env.PNC_EXT_OAUTH_PASSWORD}",
-            "baseAuthUrl":"${env.PNC_EXT_REST_BASE_URL}"
+            "auth-server-url":"${env.PNC_EXT_REST_BASE_URL}",
+            "realm-public-key": "MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAjG+B9bN0bQNDzpazVrpIG5SLEGmz7iehC/IjKLkFhwetP+nIVQi5hn59osJNrQ/vTGDdEYWLProLzkhJHRQzwZs2ucvyolaL/FcQ+l7UCxrewyhSsIubra14dJwhglKqQS+ZD2nRcQuiMJ97J9D9gqnk8z31mUO1Oq57HSIEwq/VH8YEWsOSnvd1ZDoYiKffnrghq1DTaEgPZm3aKPMmPkAS3n7ClC2Doww8r18EOEh+y8cea0M3fOpFL81137likn1m0phSHmdJ4MLVhUwgFn50UDGTjzB3pFAkUWE7s473C4zxAamLA1740hnUgXoCL36KKaMbZLQ4ZJnQEZQL6wIDAQAB",
+            "resource": "pncdirect",
+            "realm": "pncredhat"
```
matejonnet: This values should not be hard-coded.

tremes: @matejonnet AFAIK you need them to construct the initial request and token at https://github.com/project-ncl/pnc/blob/master/auth/src/main/java/org/jboss/pnc/auth/ExternalAuthentication.java#L87 or do you want them to be configurable via env properties as well?

auth/src/main/java/org/jboss/pnc/auth/ExternalAuthFacade.java
```diff
@@ -44,20 +44,22 @@
...
+    AuthenticationModuleConfig config;
     
     /**
      * Load configuration from config json file
      * 
      * @param keycloakConfiguration
      * @throws Exception
      */
-    public ExternalAuthFacade(InputStream keycloakConfiguration) throws Exception{
+    public ExternalAuthFacade() throws Exception{
```
matejonnet: It looks like this class is used in integration tests only, do we need it ?

auth/src/main/java/org/jboss/pnc/auth/ExternalAuthentication.java
```diff
@@ -31,45 +32,33 @@
...
-import org.keycloak.adapters.installed.KeycloakInstalled;
+import org.keycloak.common.VerificationException;
+import org.keycloak.common.util.KeycloakUriBuilder;
 import org.keycloak.constants.ServiceUrlConstants;
 import org.keycloak.representations.AccessToken;
 import org.keycloak.representations.AccessTokenResponse;
 import org.keycloak.util.JsonSerialization;
-import org.keycloak.util.KeycloakUriBuilder;
 
 public class ExternalAuthentication {
```
matejonnet: It looks like this class is used in integration tests only, do we need it ?

auth/src/main/java/org/jboss/pnc/auth/ExternalAuthFacade.java
tremes: This one could be likely removed.

auth/src/main/java/org/jboss/pnc/auth/ExternalAuthentication.java
tremes: This one could be moved to integration-tests because we still need some piece of code to authenticate user and token in the tests.

tremes:
I am going to prepare another one.

PR #1090, opened 2016-07-21
NCL-2077

dans123456:
fixed swagger annotation systemimageid for environments

### DIFF
--- a/rest-model/src/main/java/org/jboss/pnc/rest/restmodel/BuildEnvironmentRest.java
+++ b/rest-model/src/main/java/org/jboss/pnc/rest/restmodel/BuildEnvironmentRest.java
@@ -125,8 +125,8 @@ public class BuildEnvironmentRest implements GenericRestEntity<Integer> {
         return systemImageId;
     }
 
-    public void setImageId(String imageId) {
-        this.systemImageId = imageId;
+    public void setImageId(String systemImageId) {
+        this.systemImageId = systemImageId;
     }
 
     public Map<String, String> getAttributes() {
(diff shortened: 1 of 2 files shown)


rest-model/src/main/java/org/jboss/pnc/rest/restmodel/BuildEnvironmentRest.java
```diff
@@ -125,8 +125,8 @@ public String getImageId() {
         return systemImageId;
     }
 
-    public void setImageId(String imageId) {
-        this.systemImageId = imageId;
+    public void setImageId(String systemImageId) {
```
pgier: I wouldn't think the parameter name here would have any effect.  Do the getter and setter method names also need to be changed from ImageId to SystemImageId?

dans123456:
I found this very odd too.. Just changing the var name in the setter from imageId to systemImageId fixed the swagger doc. I'll change the method names for consistency.

rest-model/src/main/java/org/jboss/pnc/rest/restmodel/BuildEnvironmentRest.java
```diff
@@ -121,12 +121,12 @@ public void setSystemImageRepositoryUrl(String systemImageRepositoryUrl) {
...
-    public String getImageId() {
+    public String getSystemImageId() {
         return systemImageId;
     }
 
-    public void setImageId(String imageId) {
-        this.systemImageId = imageId;
+    public void setSystemImageId(String systemImageId) {
+        this.systemImageId = systemImageId;
     }
```
matejonnet: Does this updated break our api back-compatibility?

dans123456: Nothing has changed with the api. The change I made corrected the generated swagger docs. Prior to my commit the field name in the docs was "imageId" rather than the correct "systemImageId". For some odd reason changing the setter name and param fixed the problem.

thescouser89:
retest this please

thescouser89:
Failure probably caused by internal server issue. Will tweak it then rerun the tests afterwards.

dans123456:
retest this please

thescouser89:
retest this please

PR #1092, opened 2016-07-21
Ncl 2257

dans123456:


rest-model/src/main/java/org/jboss/pnc/rest/restmodel/ProjectRest.java
```diff
@@ -47,6 +48,7 @@
 
     private String projectUrl;
 
+    @ApiModelProperty(notes = "Projects cannot set build configurations", readOnly = true)
```
matejonnet: What is this comment trying to say ?

PR #1093, opened 2016-07-21
Ncl 2257

dans123456:


pgier:
I think it looks fine, but see my comment on PR #1090 before merging the first commit.
This PR probably shouldn't include the NCL-2077 commit.

PR #1097, opened 2016-07-27
Ncl 2272 remove build driver status and use build status instead

matejonnet:


matejonnet:
Re-test this please.

thescouser89:
Are there changes required on the maitai side for that change?

matejonnet:
@thescouser89 it should not affect maitai process, I also got successful build with this updates being deployed.

PR #1103, opened 2016-08-03
NCL-2218 REST API for BPM tasks, incl. BC creation workflow

jsenko:


bpm/src/main/java/org/jboss/pnc/bpm/task/BpmBCCreationTask.java
```diff
@@ -0,0 +1,72 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author Jakub Senko
+ */
+@ToString(callSuper = true)
+public class BpmBCCreationTask extends BpmTask {
```
matejonnet: Please avoid using abbreviations like BC.

rest/src/main/java/org/jboss/pnc/rest/endpoint/BpmEndpoint.java
```diff
@@ -0,0 +1,168 @@
...
+
+
+    @ApiOperation(value = "Start BC creation task.", response = Singleton.class)
+    @ApiResponses(value = {
+            @ApiResponse(code = SUCCESS_CODE, message = "Success")
+    })
+    @POST
+    @Path("/tasks/start-build-configuration-creation")
+    @Consumes(MediaType.APPLICATION_JSON)
+    public Response startBCCreationTask(
```
matejonnet: Use complete name.

PR #1104, opened 2016-08-03
Add labels to Buildrecord and expose them via REST.

matejonnet:


matejonnet:
Re-test this please!

pgier:
Commit looks ok to me, but I wonder if the new field should be called something like "attributes" instead of "labels".  To me, label implies just a text string, instead of a key-value pair.  We've also had some discussions in the past about adding a new field to build record called "tag" or "label" which would be just a text string used to group certain build records.  For example a label called "eap-test-builds" or something similar.

Also the equivalent field in BuildEnvironment is called "attributes", so it would be consistent with that entity.

matejonnet:
I remember the discussion on labelling, we can probably handle all tagging/labelling with this map, without another specific field. Do you see some potential issues with it ?

The inspiration for key/value labels is from Kubernetes [1] and OpenShift, I must say that I found that approach quite handy to use.

1: http://kubernetes.io/docs/user-guide/labels/

pgier:
I don't see any technical problem with it, so I'm ok with the merge.  I would just argue that "attribute" is a better term.  I'm not sure why Kubernetes chose "label" instead of "attribute" or "property", maybe they already had something else using those terms.

thescouser89:
Looks like this change won't affect Dependency Analysis. :+1:

jrenaat:
@matejonnet @jbartece This was **my** issue ... I am soooo pissed at both of you ... y-L
(just poking at you guys)

PR #1113, opened 2016-08-08
NCL-2273

dans123456:
Create/update milestones form checks for version uniqueness. 

Please comment on anything that seems odd as this is my first time working with ajs

### DIFF
--- a/ui/app/milestone/factories/get-milestones.js
+++ b/ui/app/milestone/factories/get-milestones.js
@@ -21,17 +21,16 @@
 
     var module = angular.module('pnc.milestone');
 
-    module.factory('versionFactory', ['$http',
+    module.factory('versionFactory', ['ProductMilestoneDAO',
 
-        function ($http) {
-            var restUrl = '../pnc-rest/rest/product-milestones/product-versions/',
-                factory = {};
+        function (ProductMilestoneDAO) {
+            var factory = {};
 
             factory.checkUniqueValue = function (productVersionId, milestoneVersion, productVersion) {
-                return $http.get(restUrl + productVersionId, {cache: true}).then(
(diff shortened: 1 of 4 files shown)


dans123456:
retest this please

ui/app/milestone/factories/get-milestones.js
```diff
@@ -0,0 +1,47 @@
...
+    var module = angular.module('pnc.milestone');
+
+    module.factory('versionFactory', ['$http',
+
+        function ($http) {
+            var restUrl = '../pnc-rest/rest/product-milestones/product-versions/',
+                factory = {};
+
+            factory.checkUniqueValue = function (productVersionId, milestoneVersion, productVersion) {
+                return $http.get(restUrl + productVersionId, {cache: true}).then(
```
alexcreasy: It's better that you inject our REST client object here, our base URLs are dynamically configured using the PNC main config file and this will take care of it for you. You can do this by:
- Inject ProductMilestoneDAO
- Call ProductMilestoneDAO.getByProductVersion({ versionId: productVersionId }).$promise.then( ....

alexcreasy:
@dans123456 nothing odd looking here, just the one change to make to use one of our existing APIs so that our URLs are dynamically generated.

ui/app/milestone/factories/get-milestones.js
dans123456: getByProductVersion does not exist. Do you mean _getByProductVersion? Is meant to be a "private" method though? I'm guessing due to the leading underscore.

alexcreasy: Sorry. that should be "getAllForProductVersion" 
This line here, decorates the private method and provides a new public method: 
https://github.com/project-ncl/pnc/blob/master/ui/app/common/restclient/services/ProductMilestoneDAO.js#L58

PR #1114, opened 2016-08-09
NCL-2292 attempt to fix BPM process parameters

michalszynkiewicz:


bpm/src/main/java/org/jboss/pnc/bpm/BpmTask.java
```diff
@@ -150,6 +150,7 @@ public Long getProcessInstanceId() {
             throw new CoreException("Could not serialize process parameters '" +
                     parameters + "'.", e);
         }
+
         actualParameters.put("pncBaseUrl", config.getPncBaseUrl());
         actualParameters.put("taskId", taskId);
         return actualParameters;
```
ruhan1: Maybe these two are not needed? i guess all such params are wrapped in processParameters.  
         actualParameters.put("pncBaseUrl", config.getPncBaseUrl());
         actualParameters.put("taskId", taskId);

PR #1127, opened 2016-08-11
[NCL-2166] Externalize build agent templates

thescouser89:
With this commit, you can now override which definition file to use for
the Openshift Driver at runtime.

To override:

- pnc-builder-pod definition, pass the flag '-Dv1_pnc-builder-pod-file=<file>'
- pnc-builder-service definition, pass the flag '-Dv1_pnc-builder-service-file=<file>'
- pnc-builder-route definition, pass the flag '-Dv1_pnc-builder-route-file=<file>'
- pnc-builder-ssh-service definition, pass the flag '-Dv1_pnc-builder-ssh-service-file=<file>'

The flag could be set in your `standalone.conf` in your EAP deployment.

### DIFF
--- a/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/Configurations.java
+++ b/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/Configurations.java
@@ -28,30 +28,49 @@ import java.lang.invoke.MethodHandles;
 /**
  * @author <a href="mailto:matejonnet@gmail.com">Matej Lazar</a>
  */
-public enum Configurations {
+public class Configurations {
 
-    V1_PNC_BUILDER_POD("v1_pnc-builder-pod.json"),
-    V1_PNC_BUILDER_SERVICE("v1_pnc-builder-service.json"),
-    V1_PNC_BUILDER_ROUTE("v1_pnc-builder-route.json"),
-    V1_PNC_BUILDER_SSH_SERVICE("v1_pnc-builder-ssh-service.json");
+    private static final String DEFAULT_V1_PNC_BUILDER_POD = "v1_pnc-builder-pod.json";
+    private static final String DEFAULT_V1_PNC_BUILDER_SERVICE_FILE = "v1_pnc-builder-service.json";
+    private static final String DEFAULT_V1_PNC_BUILDER_ROUTE_FILE = "v1_pnc-builder-route.json";
+    private static final String DEFAULT_V1_PNC_BUILDER_SSH_SERVICE_FILE = "v1_pnc-builder-ssh-service.json";
 
     private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass().getName());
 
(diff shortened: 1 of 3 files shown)


thescouser89:
@matejonnet Please review! :smiley:

openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/Configurations.java
```diff
@@ -28,30 +28,49 @@
 /**
  * @author <a href="mailto:matejonnet@gmail.com">Matej Lazar</a>
  */
-public enum Configurations {
+public class Configurations {
```
matejonnet: Why did you change it from enum to class, now the implementation is more verbose and it doesn't seem to be more readable.

openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/Configurations.java
```diff
@@ -28,30 +28,49 @@
...
+    private static final String DEFAULT_V1_PNC_BUILDER_SERVICE_FILE = "v1_pnc-builder-service.json";
+    private static final String DEFAULT_V1_PNC_BUILDER_ROUTE_FILE = "v1_pnc-builder-route.json";
+    private static final String DEFAULT_V1_PNC_BUILDER_SSH_SERVICE_FILE = "v1_pnc-builder-ssh-service.json";
 
     private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass().getName());
 
     private static final String CONFIGURATIONS_FOLDER = "openshift.configurations/";
 
-    private final String filePath;
+    private static final String CUSTOM_V1_PNC_BUILDER_POD_PROPERTY = "v1_pnc-builder-pod-file";
```
matejonnet: Drop "v1_" from the property names. The default files are v1, but user might specify v2 format (once available/supported).

matejonnet:
What about making it configurable via pnc-config.json in "openshift-environment-driver" section, to keep all the configurations in one place ?

openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/Configurations.java
thescouser89: Will do!

openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/Configurations.java
thescouser89: I changed it to class to change the intent of the 'class' from being fairly static (enum) to more dynamic (with property name to specify template file). I can change it back to enum though and add the logic in `getContentsAsString()`

thescouser89:
Yeah I can put that in the pnc-config.json. Will override that PR with the new change

thescouser89:
New PR created to include the changes: #1144 

Closing this one

PR #1129, opened 2016-08-15
NCL-2215 New paging API

alexcreasy:
This addresses a number of issues with the current API:
    - Automatically applies paging functionality where necessary, there's no need to add decorating functions calls to ng-resource services.
    - Adds ability to filter queries by a field, order by a field and change page size.
    - Returns promises so as doesn't mess up ui router resolves
    - decouples paging and pagination API.
    - decouples searching and pagination. 
    - Adds new features such as ability to change page size, or add query parameters to the search.
    - Doesn't always cause a wildcard RSQL query to be performed even when there is no search term.

### DIFF
--- a/ui/app/_pnc.js
+++ b/ui/app/_pnc.js
@@ -39,7 +39,8 @@
     'pnc.import',
     'pnc.report',
     'pnc.properties',
-    'pnc.common.authentication'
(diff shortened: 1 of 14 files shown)


alexcreasy:
Test this please

PR #1131, opened 2016-08-16
NCL-2293 ssh credentials moved to rest api

michalszynkiewicz:


openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
```diff
@@ -170,8 +170,7 @@ private void initDebug() {
 
             debugData.setSshServiceInitializer(d -> {
                 Integer port = startSshService();
-                d.setSshPort(port);
-                d.setSshHost(route.getHost());
+                d.setSshCommand("ssh worker@" + route.getHost() + " -p " + port);
```
pgier: Maybe should use a final String for the username "worker".

pgier:
Code looks fine to me.  Would it be better to just always return the ssh credentials in the BuildRecord even if the current user is not the one who started the build?

michalszynkiewicz:
@pgier it was decided that it would be unsafe. For now we're showing credentials for everybody, but later we want to hide it...

openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
michalszynkiewicz: done

PR #1144, opened 2016-08-18
[NCLi-2166] Externalize build agent templates

thescouser89:
Changes:

- `OpenshiftBuildAgentConfig` created so that we can put the pod
  definition stuff into our pnc-config.json if desired. The definition
  in the config file looks like this:
  ```json
  {
    "@module-group": "pnc",
    "configs": [
        {
            "@module-config": "openshift-build-agent",
            "pncBuilderPod": {...},
            "pncBuilderService": {...},
            "pncBuilderRoute": {...},
            "pncBuilderSshRoute": {...}
        },
        ...
    ]
  }
  ```
  You don't need to specify a key in the config if you only want to use
  the default definition for that key.

  The JSON object for each key will be interpreted as a string and used
  as the pod definition.

- `Configurations` will by default read from the default pod definition
  files if the `openshift-build-agent` section is not specified in the
  config files

- The default pod definition files were renamed to drop the 'v1_' in
  their names, to future-proof for when the definition files switch to
  `v2`. This change was done because the `v1_` was also hardcoded in the
  PNC code, and upgrading to `v2` would also require a change in the
  codebase. Not ideal.

- The enums were also renamed to drop the `v1` for the same reasons as
  the previous section.


michalszynkiewicz:
Looks good to me!

To make sure I understand: We either use the file from `/openshift-environment-driver/src/main/resources/openshift.configurations/pnc-builder-ssh-service.json` or inline the configuration in pnc-config.json, right?

thescouser89:
@michalszynkiewicz Right! So if you don't specify anthing in pnc-config.json it'll use the one in openshift-environment-driver.

michalszynkiewicz:
:+1:

PR #1150, opened 2016-08-25
NCL-2187: brew push dtos

michalszynkiewicz:


janinko:
We need to pass around also callback information in the same way [repour](https://github.com/project-ncl/repour) does.
This means PNC/BPM will send request to causeway:
```
{
  "milestoneId": long,
  "callback": {
    "url": "...",
    "method": "..."}
}
``
with **URL** being the callback url in BPM and **method** either POST or PUT (defaults to POST).
Causeway will immidietly respond with:
```
{
    "callback": {
        "id": "...",
    },
}
```
where **id** is some random unique string identifing the request.

When Causeway finishes processing everything, it will contact BPM on the provided callback url with:
```
{
    "foo": "bar",
    "callback": {
        "status": int,
        "id": str,
    }
}
```
where **id** is the same id that causeway provided in the first response and **status** is HTTP status of the task.

File: rest-model/src/main/java/org/jboss/pnc/rest/restmodel/causeway/BuildImportResultRest.java [hunk 1]
```diff
@@ -0,0 +1,57 @@
...
+public class BuildImportResultRest {
+    /**
+     * id of pnc build record
+     */
+    private Integer buildRecordId;
+
+    /**
+     * build id assigned by brew
+     */
+    private String brewBuildId;
```
janinko: You want to have it as string or rather Integer?

PR #1156, opened 2016-08-30
Store execution root name and version in build record.

jsenko:
- This is (in case of Maven) GAV of the POM being built. This information comes
   from Repour/PME and has to be stored in the build record to be used in ~~brew bridge push~~ release process.
 - NCL-2345

### DIFF
--- a/model/src/main/java/org/jboss/pnc/model/BuildRecord.java
+++ b/model/src/main/java/org/jboss/pnc/model/BuildRecord.java
@@ -164,6 +164,24 @@ public class BuildRecord implements GenericEntity<Integer> {
     @Setter
     private String sshPassword;
 
+    /**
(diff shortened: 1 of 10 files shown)


model/src/main/java/org/jboss/pnc/model/BuildRecord.java
```diff
@@ -164,6 +164,14 @@
     @Setter
     private String sshPassword;
 
+    @Getter
+    @Setter
+    private String executionRootName;
```
matejonnet: What should be the executionRootName and where do we get it from?
Can you add some comments. (the same for version)

matejonnet:
Please use more generic term "Release process" instead of "brew bridge".

PR #1159, opened 2016-08-30
Remove deprecated scmMirror* and add scmExternal* fields to Build Configuration.

jsenko:


matejonnet:
Lets remove the deprecated fields if they are not used.

jsenko:
They are kept just in the Rest entities for backwards compatibility. They will be removed in the next release.

PR #1172, opened 2016-09-09
NCL-2302: enabled non-categorized tests when running with -Pcontainer…

michalszynkiewicz:
…-tests

### DIFF
--- a/build-executor/src/test/java/org/jboss/pnc/executor/BuildExecutionTest.java
+++ b/build-executor/src/test/java/org/jboss/pnc/executor/BuildExecutionTest.java
@@ -50,7 +50,6 @@ import javax.inject.Inject;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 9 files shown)


build-executor/src/test/java/org/jboss/pnc/executor/BuildExecutionTest.java
```diff
@@ -180,34 +179,13 @@ public void shouldNotContinueBuildOnMavenError() throws ExecutorException, Inter
         assertNoState(statusChangedEvents, BuildExecutionStatus.COLLECTING_RESULTS_FROM_REPOSITORY_MANAGER);
     }
 
-    @Test
-    public void shouldNotDestroyEnvironmentWhenPodKeptAliveSet() throws ExecutorException, InterruptedException, TimeoutException {
```
michalszynkiewicz: the test was no longer valid (env destroying is skipped at a deeper level)

process-managers/pom.xml
```diff
@@ -67,6 +67,11 @@
       <artifactId>javax.inject</artifactId>
       <scope>provided</scope>
     </dependency>
+
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+    </dependency>
```
tremes: <scope>test</scope>?

rest/src/test/java/org/jboss/pnc/rest/endpoint/BuildRecordEndpointTest.java
```diff
@@ -105,58 +98,6 @@ public void shouldGetLogsWithContent() {
...
-
-        // then
-        SshCredentials sshCredentials = record.getSshCredentials();
-        assertThat(sshCredentials).isNotNull();
-        assertThat(sshCredentials.getCommand()).isEqualTo(sshCommand);
-        assertThat(sshCredentials.getPassword()).isEqualTo(sshPassword);
-    }
-
-    @Test
-    public void shouldNotReturnSshCredentialsForOtherUsers() {
```
tremes: These two tests are also not valid?

michalszynkiewicz: I'm changing the functionality there ATM. I'm adding them in a different place

process-managers/pom.xml
michalszynkiewicz: Yes, I missed that. Thanks.I'll add `<scope>test</scope>` in the parent

PR #1179, opened 2016-09-14
Simplified log message for milestone release failure

michalszynkiewicz:
Exception message is now skipped from the push log.
It contained details that are irrelevant for the user.


process-managers/src/main/java/org/jboss/pnc/managers/ProductMilestoneReleaseManager.java
```diff
@@ -113,7 +113,7 @@ public boolean noReleaseInProgress(ProductMilestone milestone) {
             return release;
         } catch (CoreException e) {
             log.error("Error trying to start brew push task for milestone: {}", milestone.getId(), e);
-            release.setLog("Brew push BPM task creation failed. Check log for more details " + e.getMessage() + "\n");
+            release.setLog("Brew push BPM task creation failed.\nCheck log for more details.\n");
             release.setStatus(MilestoneReleaseStatus.SYSTEM_ERROR);
```
matejonnet: Isn't it useful to keep the e.getMessage ?

PR #1184, opened 2016-09-14
[NCL-2410] External pnc-build-agent defs are null

thescouser89:
Linked to NCL-2166

In NCL-2166, we add the ability to externalize pnc-build-agent pod
definitions in pnc-config.json, and the ability to only specify one
definition, not all of them.

Unfortunately, the second ability doesn't work. If one pod definition is
defined, but the others are not, the final definition used will be 'null'

This commit fixes this issue.


thescouser89:
retest this please

PR #1192, opened 2016-09-15
Update Artifact javadoc.

matejonnet:


model/src/main/java/org/jboss/pnc/model/Artifact.java
```diff
@@ -105,13 +105,18 @@
     private String deployUrl;
 
     /**
-     * The record of the build which produced this artifact
+     * The record of the build which produced this artifact.
+     * Usualy hare should be only one build record that produced this artifact.
```
pgier: Typo s/hare/there/

model/src/main/java/org/jboss/pnc/model/Artifact.java
```diff
@@ -105,13 +105,18 @@
...
+     * in such case we link the BuildRecord to the same artifact.
      */
     @ManyToMany(mappedBy = "builtArtifacts")
     private Set<BuildRecord> buildRecords;
 
     /**
-     * The builds which depend on this artifact
+     * The builds which depend on this artifact.
+     * Build records of the dependencies that were downloaded during the build
+     * and has matching record in the PNC.
```
pgier: I think this wording is a bit confusion.  How about something like this?
"The list of builds which depend on this artifact.  
For example, if the build downloaded this artifact as a Maven dependency."

matejonnet:
@pgier thanks for the comments.

PR #1194, opened 2016-09-19
Ncl 2420 store multiple checksums

matejonnet:


model/src/main/java/org/jboss/pnc/model/Artifact.java
```diff
@@ -71,11 +76,27 @@
     @Column(updatable=false)
     private String identifier;
 
+    /**
+     * Checksum is unique artifact identifier.
+     * In case it is not defined, it is calculated from defined checksum fields (md5, sha1, sha256)
+     */
```
jbartece: What's is this checksum meant to be? I suppose we should remove i

matejonnet: Removed. The idea was to keep for more flexibility, but it can cause confusion.

PR #1210, opened 2016-09-30
[NCL-2142] UI toasts notification needs improving

thescouser89:
This is the first commit adding the ability to create fancy toast
notifications.

* New Features Added
  With that, we'll be able to show toast notifications with a link in
  it. The link will execute a callback where, for example, you can open
  a new link.

  You can also add a kebab menu (declicous!) to perform several actions,
  where each action will have a callback function associated with it.

* Notification Drawer
  This commit also adds support for a notification drawer, where all the
  notifications displayed are shown there also. This is useful in case
  you missed a notification.

  Clicking on the notification in the drawer will mark the notification
  as read.

  Clicking on 'X' on the notification when it is first shown will also
  mark the notification as read in the notification drawer.

* pncNotify Factory
  A `pncNotify` factory is provided to glue together the notifications
  shown in the Notifications Drawer and the notifications themselves.

  As such, we should now only use that factory when displaying
  notifications instead on relying on using `Notifications` service
  directly.

* Usage
  To define a link in the notification, first inject the `pncNotify`
  factory, then do this call:

  ```javascript
  pncNotify.info('this is a message', 'link title', linkCallbackFunc);
  ```

  To define a kebab menu in the notification, first inject the
  `pncNotify` factory, then do this call:

  ```javascript
  // actionFn is where you'll specify your callback function
  menu = [
      {
          name: 'Action',
          title: 'Perform an action',
          actionFn: performAction
      },
      {
          name: 'Another Action',
          title: 'Do something else',
          actionFn: performAction
      }
  ];

  pncNotify.info('this is a message', undefined, undefined, menu);
  ```

  Note that the 'x' in the notifications won't be shown anymore when you
  use the kebab menu.

  If you want to define both a link and a menu:

  ```javascript
  pncNotify.info('this is a message', 'link title', linkCallbackFunc, menu);
  ```

There will be another commit to exploit all these new features in the
current PNC UI codebase.

### DIFF
--- a/ui/app/_pnc.js
+++ b/ui/app/_pnc.js
@@ -33,6 +33,7 @@
     'pnc.milestone',
     'pnc.release',
     'pnc.common.directives',
+    'pnc.common.notifications',
     'pnc.common.websockets',
     'pnc.common.events',
     'pnc.configuration-set-record',
--- a/ui/app/common/events/_events.js
+++ b/ui/app/common/events/_events.js
@@ -44,9 +44,9 @@
     'eventTypes',
     'BuildRecordDAO',
     'authService',
-    'Notifications',
+    'pncNotify',
     function($log, $rootScope, webSocketBus, eventTypes, BuildRecordDAO,
-             authService, Notifications) {
+             authService, pncNotify) {
       var scope = $rootScope.$new();
 
       //TODO: When backend functionality is available these notifications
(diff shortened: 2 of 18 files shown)


thescouser89:
An example of the commit in action:

thescouser89:
@alexcreasy please review! I'm pretty sure I messed up something somewhere

thescouser89:
Added the licenses to the html files. Hopefully this will make the testsuite happy

thescouser89:
Oh crap good point! I think I put 'pncService' as a placeholder and totally forgot to change it. Maybe change it to 'pncNotify'? To signify it's from pnc and not from some external library?

alexcreasy:
Yeah, namespacing it is probably for the best. pncNotify sounds good to me!

thescouser89:
pncService changed to pncNotify

alexcreasy:
Awesome +1! I want my Kebab toast!!!

thescouser89:
Should I create another PR for 0.10 branch? or should I cherry-pick it?

PR #1214, opened 2016-10-03
Removed deprecated scmMirror* fields.

jsenko:
This change is not backwards compatible and clients should update their rest request models.


matejonnet:
+1 for merge

PR #1215, opened 2016-10-03
NCL-2450 Add PUT /prouct-versions/:id/build-configuration-sets endpoint

alexcreasy:
to allow bulk changing of associations


alexcreasy:
@michalszynkiewicz would you mind reviewing this please?

PR #1238, opened 2016-10-11
Add build configs to config sets after BC creation.

jsenko:


jsenko:
Do not merge this yet, UI support for this must be available at the same time.

PR #1241, opened 2016-10-11
Unify how the parameters are passed to the BPM.

matejonnet:


jsenko:
Please wait until i have time to verify that it'll work on BPM side. I'll merge it after.

matejonnet:
I think Henry is already on it, see NCL-2526 and NCL-2301.

PR #1251, opened 2016-10-18
Increase timeout for build container startup

jbartece:


jbartece:
Increase of the timeout until we have proper configurable solution - @jsenko  is on the proper solution

PR #1252, opened 2016-10-18
Replace PullingMonitor timeout constants with config parameters.

jsenko:
@matejonnet hi, this change causes a problem when parsing configuration, I could not figure out why. Can you take a look? Reproduce: mvn clean test


thescouser89:
I'll close it since #1255 is supposed to fix the build failures.

PR #1255, opened 2016-10-18
Replace PullingMonitor timeout constants with config parameters

thescouser89:
This PR is to fix some of the issues in #1252

### DIFF
--- a/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/monitor/PullingMonitor.java
+++ b/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/monitor/PullingMonitor.java
@@ -24,6 +24,7 @@ import org.jboss.pnc.common.json.moduleconfig.OpenshiftEnvironmentDriverModuleCo
 import org.jboss.pnc.common.json.moduleprovider.PncConfigProvider;
 import org.jboss.pnc.common.util.ObjectWrapper;
 import org.jboss.pnc.common.util.TimeUtils;
+import org.jboss.pnc.environment.openshift.OpenshiftEnvironmentDriver;
 import org.jboss.util.collection.ConcurrentSet;
 
 import javax.annotation.PreDestroy;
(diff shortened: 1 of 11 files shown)


thescouser89:
@jsenko I don't know what was up with the error in `PullingMonitorTest`. I formatted the pnc-config.json it was reading so that it used the new pnc-config format. I have no idea why we didn't hit that error before if the config format was wrong. Maybe the tests were disabled? No idea.

As for the integration tests, I added a default constructor so that JBoss could get its act together and do the injection properly.

I hope I didn't break anything! :speak_no_evil:

openshift-environment-driver/src/main/java/org/jboss/pnc/environment/monitor/PullingMonitor.java
```diff
@@ -16,14 +16,20 @@
  * limitations under the License.
  */
 
-package org.jboss.pnc.common.monitor;
+package org.jboss.pnc.environment.monitor;
```
matejonnet: Why moving this one to the environment module, can't we keep it generic ?

openshift-environment-driver/src/main/java/org/jboss/pnc/environment/monitor/PullingMonitor.java
```diff
@@ -54,14 +54,23 @@
...
 
+    @Inject
+    private Configuration configuration;
+
     public PullingMonitor() {
+    }
+
+    @Inject
+    public PullingMonitor(Configuration configuration) {
+        this.configuration = configuration;
```
matejonnet: There already is a `monitor`method that takes timeout as a parameter. There should be no need to make this class dependent on the configuration.

jsenko:
The move is because of the circular dependency if common (where PullingMonitor was) would depend on configuration and vice versa. This in turn is because the monitor method is used 4 times, which would require the timeout configuration parameters to be passed each time. Instead they are supplied directly by the monitor. Is the move more problematic that this?

PR #1260, opened 2016-10-19
Use config parameters instead of PullingMonitor timeout constants.

jsenko:


openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
```diff
@@ -193,25 +195,32 @@ public void monitorInitialization(Consumer<RunningEnvironment> onComplete, Consu
...
             logger.info("Route [{}].", route.getName());
         }
 
         //logger.info("Waiting to start a pod [{}], service [{}].", pod.getName(), service.getName());
     }
 
+    private void monitor(Runnable onMonitorComplete, Consumer<Exception> onMonitorError, Supplier<Boolean> condition) {
+        pullingMonitor.monitor(onMonitorComplete, onMonitorError, condition,
+                environmentConfiguration.getBuildStartCheckIntervalSeconds(),
+                environmentConfiguration.getBuildStartCheckIntervalSeconds(),
```
jsenko: this should be the timeout

PR #1268, opened 2016-10-25
NCL-1903: improve indy promotion validation error format.

jrenaat:
Improve indy promotion validation error format.
Added debug test for this.


jbartece:
As shown by test results IndyPromotionValidationtest.java  file miss copyright header and also there is a typo in the filename

PR #1271, opened 2016-10-25
test fix

michalszynkiewicz:


jbartece:
:+1:

datastore/src/test/resources/test-persistence.xml
```diff
@@ -27,7 +27,7 @@
     <jta-data-source>java:jboss/datasources/NewcastleTestDS</jta-data-source>
 
     <!-- Load the model entities -->
-    <jar-file>lib/model-0.10-SNAPSHOT.jar</jar-file>
+    <jar-file>lib/model-0.10.2-SNAPSHOT.jar</jar-file>
```
matejonnet: Jars need a version. Do we need to list them here? If we do, can we generate this file somehow ?

PR #1278, opened 2016-10-26
Return a promise in all cases to fix ssh-credentials button not being…

alexcreasy:
… hidden when not available


alexcreasy:
Test this please

PR #1297, opened 2016-11-02
[NCL-2539] Configuration loaded multiple times

thescouser89:
We instead use the Application Scope PNC to inject the already loaded configuration.

### DIFF
--- a/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/Configurations.java
+++ b/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/Configurations.java
@@ -26,6 +26,7 @@ import org.jboss.pnc.common.util.IoUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.inject.Inject;
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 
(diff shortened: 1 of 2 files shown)


michalszynkiewicz:
the jenkins errors are not related to the change, so if the code is ok - feel free to merge.

michalszynkiewicz:
retest this please

openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/Configurations.java
```diff
@@ -45,11 +46,11 @@
 
     private final String filePath;
 
+    @Inject
```
matejonnet: Does Inject work in the enum ?

thescouser89: Yup it totally won't work. Verified by using some test code. Ah will refactor!

matejonnet: Basically Inject works only in objects instantiated by the CDI. Meaning no Inject when you use new MyClass() or MyClass.fieldOrMethod(), which is kinda the enum case.

PR #1309, opened 2016-11-08
[NCL-2539] Configuration loaded multiple times

thescouser89:
We instead use the Application Scope PNC to inject the already loaded configuration.

This is an updated PR to #1297 . I have removed the `@Inject` in the enum, and instead provide the configuration to the method requiring `configuration` via the method parameter.


openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
```diff
@@ -85,6 +87,9 @@
     private Route route;
     private Service sshService;
 
+    @Inject
+    private Configuration configuration;
```
matejonnet: OpenshiftStartedEnvironment is also instantited using new, thus not managed by the CDI.

thescouser89: Changed! Thanks for catching this again!

I'm not really happy with my fix though.

thescouser89:
Woah sorry about dimissing the review. I don't know how to just say "It's been resolved" on Github.

thescouser89:
Closed! Updated the PR in #1408

PR #1316, opened 2016-11-09
NCL-2496: Enable authentication in Indy

pkocandr:
For all operations it uses now the token of currently logged in user.
Therefore we had to drop the global repos setup at PNC startup, because
there is no user signed in. But it was not used anyway, because the
setup did not contain all the needed repos.

To be able to add the support we needed to use separate Indy client
for each build. That is achieved by calling init() at the beginning of
each RepositoryManager's methods and the in completeExecution there is
the connection close() called. Another option was to close the
connection after the client is not used for a period of time, but this
approach was easier to implement. We'll see if that works well.

Regarding the tests I used null token for now. I think there is the mock
used, so it should be ok. Not sure if there are any integration tests
using a live Indy instance.

### DIFF
--- a/build-executor/src/main/java/org/jboss/pnc/executor/DefaultBuildExecutor.java
+++ b/build-executor/src/main/java/org/jboss/pnc/executor/DefaultBuildExecutor.java
@@ -175,7 +175,7 @@ public class DefaultBuildExecutor implements BuildExecutor {
         try {
             RepositoryManager repositoryManager = repositoryManagerFactory.getRepositoryManager(ArtifactRepo.Type.MAVEN);
             BuildExecution buildExecution = buildExecutionSession.getBuildExecutionConfiguration();
-            return repositoryManager.createBuildRepository(buildExecution);
+            return repositoryManager.createBuildRepository(buildExecution, buildExecutionSession.getAccessToken());
         } catch (Throwable e) {
             throw new BuildProcessException(e);
         }
@@ -364,6 +364,14 @@ public class DefaultBuildExecutor implements BuildExecutor {
         } else {
             buildExecutionSession.setEndTime(new Date());
         }
+        
+        String accessToken = buildExecutionSession.getAccessToken();
+        log.debug("Closing Maven repository manager [" + buildExecutionSession.getId() + "].");
+        try {
+            repositoryManagerFactory.getRepositoryManager(ArtifactRepo.Type.MAVEN).close(accessToken);
+        } catch (ExecutorException executionException) {
+            buildExecutionSession.setException(executionException);
+        }
 
         //check if any of previous statuses indicated "failed" state
         if (buildExecutionSession.isCanceled()) {
@@ -374,7 +382,7 @@ public class DefaultBuildExecutor implements BuildExecutor {
             buildExecutionSession.setStatus(BuildExecutionStatus.DONE);
         }
 
-        log.debug("Removing buildExecutionTask [" + buildExecutionSession.getId() + "] form list of running tasks.");
+        log.debug("Removing buildExecutionTask [" + buildExecutionSession.getId() + "] from list of running tasks.");
         runningExecutions.remove(buildExecutionSession.getId());
 
         return null;
--- a/maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
+++ b/maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
@@ -59,7 +59,7 @@ import static org.jboss.pnc.mavenrepositorymanager.MavenRepositoryConstants.SHAR
 import static org.jboss.pnc.mavenrepositorymanager.MavenRepositoryConstants.UNTESTED_BUILDS_GROUP;
 
 /**
- * Implementation of {@link RepositoryManager} that manages an <a href="https://github.com/jdcasey/indy">AProx</a> instance to
+ * Implementation of {@link RepositoryManager} that manages an <a href="https://github.com/jdcasey/indy">Indy</a> instance to
  * support repositories for Maven-ish builds.
  * <p>
  * Created by <a href="mailto:matejonnet@gmail.com">Matej Lazar</a> on 2014-11-25.
--- a/maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
+++ b/maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
@@ -17,9 +17,12 @@
  */
 package org.jboss.pnc.mavenrepositorymanager;
 
+import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.commonjava.indy.client.core.Indy;
 import org.commonjava.indy.client.core.IndyClientException;
+import org.commonjava.indy.client.core.auth.IndyClientAuthenticator;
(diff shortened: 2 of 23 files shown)


build-executor/src/main/java/org/jboss/pnc/executor/DefaultBuildExecutor.java
```diff
@@ -123,15 +123,19 @@ public BuildExecutionSession startBuilding(
...
+	        CompletableFuture.supplyAsync(() -> configureRepository(buildExecutionSession), executor)
+	                .thenApplyAsync(repositoryConfiguration -> setUpEnvironment(buildExecutionSession, repositoryConfiguration, debugData), executor)
+	                .thenComposeAsync(startedEnvironment -> waitForEnvironmentInitialization(buildExecutionSession, startedEnvironment), executor)
+	                .thenComposeAsync(runningBuild -> runTheBuild(buildExecutionSession), executor)
+	                //no cancellation after this point
+	                .thenApplyAsync(completedBuild -> optionallyEnableSsh(buildExecutionSession, completedBuild), executor)
+	                .thenApplyAsync(completedBuild -> retrieveBuildDriverResults(buildExecutionSession, completedBuild), executor)
+	                .thenApplyAsync(nul -> retrieveRepositoryManagerResults(buildExecutionSession), executor)
+	                .handleAsync((nul, e) -> completeExecution(buildExecutionSession, e), executor);
+        } finally {
```
matejonnet: All the exceptions should be caught by "handleAsync".

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
```diff
@@ -84,23 +97,25 @@ public RepositoryManagerDriver(Configuration configuration) {
...
+                throw new IllegalStateException("Failed to setup shared-releases or shared-imports hosted repository: "
+                        + e.getMessage(), e);
+            }
         }
+        return indy;
     }
 
-    @PreDestroy
-    public void shutdown() {
-        indy.close();
```
matejonnet: No need to close it anymore ?

build-executor/src/main/java/org/jboss/pnc/executor/DefaultBuildExecutor.java
pkocandr: You mean it is not needed to put this in a finally block?

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
matejonnet: I see, I missed that,

build-executor/src/main/java/org/jboss/pnc/executor/DefaultBuildExecutor.java
matejonnet: handleAsync is always called

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
```diff
@@ -68,54 +72,65 @@
...
     public RepositoryManagerDriver() { // workaround for CDI constructor parameter injection bug
     }
-
-    @SuppressWarnings("resource")
+    
     @Inject
     public RepositoryManagerDriver(Configuration configuration) {
         MavenRepoDriverModuleConfig config;
-        try {
+		try {
```
matejonnet: Remove double Indent.

build-executor/src/main/java/org/jboss/pnc/executor/DefaultBuildExecutor.java
```diff
@@ -374,9 +374,17 @@ private Void completeExecution(DefaultBuildExecutionSession buildExecutionSessio
...
 
-        log.debug("Removing buildExecutionTask [" + buildExecutionSession.getId() + "] form list of running tasks.");
+        log.debug("Removing buildExecutionTask [" + buildExecutionSession.getId() + "] from list of running tasks.");
         runningExecutions.remove(buildExecutionSession.getId());
 
+        String accessToken = buildExecutionSession.getAccessToken();
+        log.debug("Closing Maven repository manager [" + buildExecutionSession.getId() + "].");
+        try {
+			repositoryManagerFactory.getRepositoryManager(ArtifactRepo.Type.MAVEN).close(accessToken);
+		} catch (ExecutorException executionException) {
```
matejonnet: Indent!

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
```diff
@@ -68,54 +72,65 @@
...
+        } catch (ConfigurationParseException e) {
+            throw new IllegalStateException("Cannot read configuration for " + DRIVER_ID + ".", e);
+        }
 
-            internalRepoPatterns = new ArrayList<>();
-            internalRepoPatterns.add(MavenRepositoryConstants.SHARED_IMPORTS_ID);
+        indyBaseUrl = config.getBaseUrl();
+        if (indyBaseUrl.endsWith("/")) {
+            indyBaseUrl = indyBaseUrl.substring(0, indyBaseUrl.length() - 1);
+        }
```
matejonnet: Use can use StringUtils.stripEndingSlash

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
pkocandr: Stupid eclipse inserted tabs instead of spaces, so the number is correct, but github shows them as 8 spaces while eclipse as 4 :-)

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
pkocandr: OK, I'll change that, but this is not my code. I just slightly moved it.

build-executor/src/main/java/org/jboss/pnc/executor/DefaultBuildExecutor.java
pkocandr: Sure, tabs again :-)

build-executor/src/main/java/org/jboss/pnc/executor/DefaultBuildExecutor.java
```diff
@@ -374,9 +374,17 @@ private Void completeExecution(DefaultBuildExecutionSession buildExecutionSessio
...
-        log.debug("Removing buildExecutionTask [" + buildExecutionSession.getId() + "] form list of running tasks.");
+        log.debug("Removing buildExecutionTask [" + buildExecutionSession.getId() + "] from list of running tasks.");
         runningExecutions.remove(buildExecutionSession.getId());
 
+        String accessToken = buildExecutionSession.getAccessToken();
+        log.debug("Closing Maven repository manager [" + buildExecutionSession.getId() + "].");
+        try {
+            repositoryManagerFactory.getRepositoryManager(ArtifactRepo.Type.MAVEN).close(accessToken);
+        } catch (ExecutorException executionException) {
+            e = executionException;
```
matejonnet: This exception "e" is not handled down-here. See ` if (e != null) {` few lines above.

PR #1319, opened 2016-11-09
Fix show credentials button

alexcreasy:


alexcreasy:
@michalszynkiewicz I've tested this both with and without and it *should* work finally... please test given how many times I've fixed this :P

PR #1322, opened 2016-11-10
Add attributes to the ProductVersion

jbartece:
Change brew tag prefix to be a classic attribute


jbartece:
I might change the size of the attributes to 255 - I limited it to 50 for the brew tag, but generally it should be bigger, I suppose

PR #1330, opened 2016-11-15
Add optimistic locking to the BuildRecord

jbartece:


matejonnet:
What are the fields that can be updated, it should be immutable, except the attributes, that should be in a separate DB table ?

jbartece:
I'm not sure I understand. It cannot be immutable as the Hibernate needs to update it in every persist/merge operation. It should be part of the entity, but it doesn't need getters/setters as it's not intended for application, but for Hibernate and it can use reflection to manipulate it.

jbartece:
But maybe better field name would be good

matejonnet:
BuildRecord should be protected from the updates. As mentioned, only the attributes can be added to it, if that can cause updates to the BuildRecord entry, we should refactor our api to prevent that.
We should also annotate the entity with updateable=false.

jbartece:
>> BuildRecord should be protected from the updates.
How is it protected? I see there setters in the BuildRecord entity and method save in the BuildRecordRepositoryImpl (inherited from AbstractRepository).
Or you mean, we should protect it from updates?

AFAIK it's not possible to make the whole entity updatable=false - we need to set it directly to the fields, which cannot be changed. 

If you mean that, it can be done and yes, then the @Version field is not necessary - I just had a feeling there can be possibly concurrent access to the entity.

matejonnet:
There can be concurrent read access, but only one write. If that is not available per entity then we should add it to the fields. Except for the attributes, where we should have "insert-only", if that is possible.

jbartece:
OK. That should be possible. I'll do a PR

jbartece:
Shouldn't we let the log, endtime and BuildStatus as updatable? 
I expect that these are updated during the build as the BuildRecord entity is created, when the build is triggered. 
I think that also these fields needs to be updatable: sshCommand, sshPassword, executionRootName, executionRootVersion

matejonnet:
Nope, the entity is immutable. Created and stored at once, without updates.

matejonnet:
Closing as we don't need the locking, we should set all the fields to updatable=false.

PR #1331, opened 2016-11-15
Fix wrong table name User

jbartece:


model/src/main/java/org/jboss/pnc/model/User.java
```diff
@@ -34,7 +34,7 @@
  * @author avibelli
  */
 @Entity
-@Table(uniqueConstraints = { @UniqueConstraint(name = "uk_user_email", columnNames = { "email" }),
+@Table(name = "UserTable", uniqueConstraints = { @UniqueConstraint(name = "uk_user_email", columnNames = { "email" }),
```
matejonnet: Why do we need to force new table name ?

jbartece: As User is restricted keyword for Postgres and deployment failed because of that

matejonnet: Any idea why it's been working up to now ?

jbartece: I messed it up in the DB Review PR - it wasn't captured by tests

matejonnet: I see it used to be called `Users` which actually sounds better than `UserTable`.

PR #1334, opened 2016-11-15
[NCL-2373] Make all threadpool configurable

thescouser89:
ThreadPool now configurable in class:

- OpenshiftEnvironmentDriver
- PullingMonitor

ThreadPool *not* made configurable in class:

- JenkinsBuildMonitor (module not built)
- BuildExecutorMock (Used for testing only)

The `OpenshiftEnvironmentDriver` is configured via the
`openshift-environment-driver` config section, with key
`executorThreadPoolSize`.

The `PullingMonitor` is configured via the system config section, with
key `monitorThreadPoolSize`.

The `monitor` sub-module is moved from `common` to `coordinator` so as the
`PullingMonitor` class, found in the `monitor` sub-module, can have
access to the configuration. The `common` module cannot have access to
the configuration's module configs because of circular dependency.

### DIFF
--- a/build-coordinator/pom.xml
+++ b/build-coordinator/pom.xml
@@ -59,6 +59,16 @@
       <groupId>org.jboss.pnc</groupId>
       <artifactId>bpm</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.jboss</groupId>
+      <artifactId>jboss-common-core</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>org.jboss.logging</groupId>
+          <artifactId>jboss-logging-spi</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
 
     <!-- Remote dependencies -->
     <dependency>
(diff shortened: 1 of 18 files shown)


build-coordinator/pom.xml
```diff
@@ -59,6 +59,16 @@
...
+    <dependency>
+      <groupId>org.jboss</groupId>
+      <artifactId>jboss-common-core</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>org.jboss.logging</groupId>
+          <artifactId>jboss-logging-spi</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
```
matejonnet: It should be with other remote dependencies.

openshift-environment-driver/pom.xml
```diff
@@ -54,6 +54,10 @@
       <groupId>org.jboss.pnc</groupId>
       <artifactId>spi</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.jboss.pnc</groupId>
+      <artifactId>build-coordinator</artifactId>
+    </dependency>
```
matejonnet: I don't think it is work introducing this dependency, due to configuration mess. PullingMonitor should stay in the common. I would rather made it configurable in "non-standard" way, via system property -D and env variable. Where system property overrides env if both are set.

thescouser89: I get what you mean. But I'm also not a fan of the non-standard way. I'd rather have everything defined in pnc-config.json rather have it spread between system property and pnc-config.json

matejonnet: An option is to make it configurable from the code before the first usage.
But being part of common, where should be no PNC specific stuff, I think it is ok to configure it via system property. To keep all the configurations in the pnc-config.json, we can set the system property programatically on config load.
Looking again at the added dependency, we cannot do that, as build-coordinator should be decoupled from the executor (environment driver and other). The plan is to extract executor as its own service.

PR #1355, opened 2016-11-23
NCL-2624 Internal and external fields added on product import

matedo1:


alexcreasy:
@matedo1 +1 this works as expected, also nice code de-duplication.

PR #1359, opened 2016-11-24
NCL-2651: display Causeway error message in the import log

michalszynkiewicz:


process-managers/src/main/java/org/jboss/pnc/managers/ProductMilestoneReleaseManager.java
```diff
@@ -170,6 +170,10 @@ private String describeCompletedPush(MilestoneReleaseResultRest result) {
         stringBuilder.append("Brew push ").append(success ? "SUCCEEDED" : "FAILED").append("\n");
         stringBuilder.append("Import details:\n");
 
+        String errorMessage = result.getErrorMessage();
+        if (errorMessage != null) {
+            stringBuilder.append("\n").append(errorMessage).append("\n");
```
janinko: Isn't the first "\n" extra?

PR #1363, opened 2016-11-25
[NCL-2373] Make all threadpool configurable

thescouser89:
ThreadPool now configurable in class:

- OpenshiftEnvironmentDriver
- PullingMonitor

ThreadPool not made configurable in class:

- JenkinsBuildMonitor (module not built)
- BuildExecutorMock (Used for testing only)

The `OpenshiftEnvironmentDriver` is configured via the
`openshift-environment-driver` config section, with key
`executorThreadPoolSize`.

The `PullingMonitor` is configured via system properties, with key
`pulling-monitor-threadpool`. For example, you can pass:

```
-Dpulling-monitor-threadpool=15
```

parameter.

### DIFF
--- a/common/src/main/java/org/jboss/pnc/common/monitor/PullingMonitor.java
+++ b/common/src/main/java/org/jboss/pnc/common/monitor/PullingMonitor.java
@@ -21,6 +21,8 @@ package org.jboss.pnc.common.monitor;
 import org.jboss.pnc.common.util.ObjectWrapper;
 import org.jboss.pnc.common.util.TimeUtils;
 import org.jboss.util.collection.ConcurrentSet;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.PreDestroy;
 import javax.enterprise.context.ApplicationScoped;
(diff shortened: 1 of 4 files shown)


thescouser89:
This PR includes suggestions in #1334

common/src/main/java/org/jboss/pnc/common/monitor/PullingMonitor.java
```diff
@@ -49,15 +52,32 @@
...
     private ScheduledExecutorService executorService;
     private ScheduledExecutorService timeOutVerifierService;
 
     private ConcurrentSet<RunningTask> runningTasks;
 
     public PullingMonitor() {
+
+        int threadSize = DEFAULT_EXECUTOR_THREADPOOL_SIZE;
+        String threadSizeEnv = System.getProperty(PULLING_MONITOR_THREADPOOL_KEY);
+
```
matejonnet: Please add an option to read the value also from environment variable "PNC_PULLING_MONITOR_THREADPOOL".
Where system property overrides env when both are set.

thescouser89: I have updated the commit to read from both env variable and system property.

PR #1410, opened 2017-01-04
fixed validation for empty internal repo

michalszynkiewicz:
When someone enters internal repo url, and then erases it, the UI sends an empty string to the backend.

This commit fixes validation so that the empty url is not checked against the required scm prefix.


jbartece:
I think we don't want to allow erasing of the internal SCM URL - this shouldn't pass the validation.

jbartece:
retest this please

michalszynkiewicz:
@jbartece this is only in the case of creating build config. 
Assume you have a GitHub URL and you put it into internal instead of external by mistake. Then you cut it from there and put it to external.
In this case, the UI sends an empty internal URL (not null, but an empty string) to the backend. This change fixes accepting the empty string (I coded it earlier but there was a bug in the code)

jbartece:
@michalszynkiewicz  It looks like you deleted your changes by rebase - I see only one line change in persistance.xml

michalszynkiewicz:
@jbartece thanks for pointing this out. Fixed.

PR #1423, opened 2017-01-26
Fix issue with transitive dependencies failure.

matejonnet:


matejonnet:
@michalszynkiewicz Please take a look if it is ok to add this extra condition [1]. It is required to prevent infinite loop which occurred in case of failed transitive dependencies.

1: https://github.com/project-ncl/pnc/pull/1423/commits/f00aabe17eeb52d355735069f232d2426c54447c#diff-b094550a9072e565e980eab02bab41cbR453

build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/DefaultBuildCoordinator.java
```diff
@@ -447,8 +447,10 @@ private boolean isFinished(BuildSetTask buildSetTask) {
     private void handleErroneousFinish(BuildTask failedTask) {
         BuildSetTask taskSet = failedTask.getBuildSetTask();
         if (taskSet != null) {
+            log.debug("Finishing tasks in set {}, after failedTask {}.", taskSet, failedTask);
             taskSet.getBuildTasks().stream()
                     .filter(t -> isDependentOn(failedTask, t))
+                    .filter(t -> !t.getStatus().isCompleted())
```
michalszynkiewicz: is this line the real change?

Do you know how it happen that a build task was completed and had a failed dependency?

matejonnet: Test scenario I added has the dependency chain.
A->B->C, where C is failing and B fall into infinite loop.

Yes that extra filter is a real change.

michalszynkiewicz: I get it now, thanks

michalszynkiewicz:
:+1:

matejonnet:
Retest this please.

PR #1436, opened 2017-03-02
[NCL-2814]: Make the target group for promotion configurable

jbartece:
Defaults to builds-from-pnc


maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/MavenRepositorySession.java
```diff
@@ -73,7 +73,6 @@
  * group, to expose them for use in successive builds in the chain.
  */
 public class MavenRepositorySession implements RepositorySession {
-
```
matejonnet: Please add empty line back as there is always an empty line after the class line.

jbartece: DONE

pkocandr:
I don't like the field name promoteBuildToGroup, it is not very clear what that is. I think buildPromotionGroup would be more understandable. Apart from that it looks ok.

jbartece:
OK. I'll rename it

pkocandr:
OK, now it's perfect ;-)

alexcreasy:
@jbartece UI changes work fine :)

PR #1441, opened 2017-03-22
[NCL-2909] Use latest builder image on Openshift

thescouser89:
Issue
-----

PNC creates new pods used to run builds. Those pods use as image, builder
images with a tag associated with it (<builder image>:<tag, e.g latest>). I'll
use the tag 'latest' as example from now on.

When Openshift starts the pod in one of its Openshift slaves, there are some
cases when the Openshift slave chosen already has the builder image with tag
'latest' in its local docker storage.

There is an issue that arises whenever we rebuild a new builder image with tag
latest on Openshift, and try to start a new pod with image <builder>:latest on
that Openshift slave.

The Openshift slave doesn't attempt to check with docker registry if there is a
newer builder image with tag 'latest', but instead reuses the one in its local
docker storage. This causes us lots of issues because we can never be sure
which builder image we are actually using.

Fix
---
Once PNC wants to start a new build with image <builder>:latest, we ask the
Openshift instance for the latest image id of <builder>:latest instead.

Now having the latest image id, instead of creating the pod with image
<builder>:latest, I'll instead create the pod with image <builder>@<imageid>
(from: https://docs.openshift.com/container-platform/3.3/dev_guide/managing_images.html#referencing-images-in-image-streams)

This should in theory force the Openshift slave to fetch that particular
imageid from docker registry, which means that it'll force the Openshift slave
to always get the latest image for <builder>:latest and we won't suffer from
using the older builder image cached in the local docker storage anymore.

Notice
------
This commit will remove the use of the default build image id specified in
the pnc-config.

### DIFF
--- a/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftEnvironmentDriver.java
+++ b/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftEnvironmentDriver.java
@@ -94,8 +94,7 @@ public class OpenshiftEnvironmentDriver implements EnvironmentDriver {
 
         if (!canRunImageType(systemImageType))
             throw new UnsupportedOperationException("OpenshiftEnvironmentDriver currently provides support only for the following system image types:" + compatibleImageTypes);
-        String buildImageId = StringUtils.addEndingSlash(systemImageRepositoryUrl) + systemImageId;
-        return new OpenshiftStartedEnvironment(executor, configuration, config, pullingMonitor, repositorySession, buildImageId, debugData, accessToken);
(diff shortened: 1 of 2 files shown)


pweil-:
Is this the right pod definition?  https://github.com/project-ncl/pnc/blob/73cc8c858d9e5cf09a13e5653e8790b393a3102e/openshift-environment-driver/src/main/resources/openshift.configurations/pnc-builder-pod.json#L86

Looks like you're setting the `imagePullPolicy` such that it won't pull if it already has the image on the node.  See the [image pull policy docs](https://docs.openshift.org/latest/dev_guide/managing_images.html#image-pull-policy) for more info.

thescouser89:
@pweil- Yup looks like we were using the wrong pull policy. Will close this PR.

PR #1448, opened 2017-03-28
Make PNC work with OSE 3.4 (v1.0 branch)

thescouser89:
See #1442


matejonnet:
The fix is required for the next release only, not required on a branch.

PR #1450, opened 2017-04-03
NCL-2834 Require auth configurations fixed

matedo1:


matedo1:
Yes, this is what I've found out

PR #1453, opened 2017-04-05
NCL-2692 Fix build status on product version page

matedo1:


ui/app/common/restclient/services/BuildsDAO.js
```diff
@@ -50,6 +50,8 @@
         }
       });
 
+      PageFactory.decorateNonPaged(resource, '_getLastByConfiguration', 'getLastByConfigurationNonPaged');
```
alexcreasy: Is the use of "decorateNonPaged" essential here? I need to put a note in the DAOs but these methods should be considered deprecated. They introduce potentially serious performance problems for PNC. If it's a lot of work to change this we can let it go in, once we've removed all usages from the codebase though I'd take a stronger line on it.

ui/app/product/directives/pncProductVersionBCs/pncProductVersionBCs.js
```diff
@@ -48,7 +48,7 @@
           scope.page.onUpdate(function(page) {
             _.forEach(page.data, function(bc) {
               if(!_(scope.latestBuildRecords).has(bc.id)) { // avoid unnecessary requests
-                BuildRecordDAO.getLatestForConfiguration({ configurationId: bc.id }).then(function (data) {
+                BuildsDAO.getLastByConfigurationNonPaged({ id: bc.id }).then(function (data) {
```
alexcreasy: Likewise here, do we need a the non-paged method?

PR #1458, opened 2017-04-19
[NCL-2898]: Fix NPE in /build-records/###/built-artifacts

jbartece:


rest/src/main/java/org/jboss/pnc/rest/provider/ArtifactProvider.java
```diff
@@ -89,12 +91,20 @@ public ArtifactProvider(ArtifactRepository artifactRepository, RSQLPredicateProd
...
+     */
     public CollectionInfo<ArtifactRest> getBuiltArtifactsForBuildRecord(int pageIndex, int pageSize, String sortingRsql, String query,
             int buildRecordId) {
         BuildRecord buildRecord = buildRecordRepository.queryById(buildRecordId);
-
-        return filterAndSort(pageIndex, pageSize, sortingRsql, query,
-                ArtifactRest.class, buildRecord.getBuiltArtifacts());
+        Set<Artifact> builtArtifacts = Collections.emptySet();
+        if (buildRecord != null)
+        	builtArtifacts = buildRecord.getBuiltArtifacts();
```
michalszynkiewicz: my personal feelings are that a 404 in the case of missing buildRecord would say more than 204.
But I think Matej was in favor of 204 in both cases, so let's do it this way.

jbartece: I'd also go for the 404, but we should do the change in some more complex refactoring as I found out it's there multiple times, so I wanted not to break the current convention.

PR #1466, opened 2017-04-28
Ncl 2972 update build result

matejonnet:


rest/src/main/java/org/jboss/pnc/rest/provider/BuildRecordProvider.java
```diff
@@ -229,6 +229,15 @@ public String getBuildRecordLog(Integer id) {
             return null;
     }
 
+    public String getBuildRecordRepourLog(Integer id) {
+        BuildRecord buildRecord = ((BuildRecordRepository) repository).findByIdFetchAllProperties(id);
```
jbartece: When we do a request to build/repour log also the second log is obtained from the database it affects the performance of these requests as that are LOBs - NCL-2989 created

model/src/main/java/org/jboss/pnc/model/BuildRecord.java
```diff
@@ -242,6 +242,8 @@
     @Column(name="value")
     private Map<String, String> attributes = new HashMap<>();
 
+    private String repourLog;
```
jbartece: It will create a varchar(256) in database, which is not enough. We should store it as a LOB as in case of build log, so to add annotations:
@Lob
@Type(type = "org.hibernate.type.StringClobType")
@Basic(fetch = FetchType.LAZY)

PR #1517, opened 2017-06-12
NCL-3008 Add new build status icon with additional warning for corrup…

alexcreasy:
…ted builds


alexcreasy:
@matedo1 I needed the build-status-icon to take the BuildRecord itself as a param rather than the status. I took the opportunity to refactor that out into a new component. It works more or less the same.

alexcreasy:
Retest this please

PR #1520, opened 2017-06-13
Update BPM reposnose logging.

matejonnet:


rest/src/main/java/org/jboss/pnc/rest/utils/BpmNotifier.java
```diff
@@ -69,7 +69,11 @@ public void sendBuildExecutionCompleted(String uri, BuildResult buildResult) {
         String errMessage = "";
         try {
             buildResultRest = new BuildResultRest(buildResult);
-            log.debug("Sending build result to BPM " + buildResultRest.toString() + ".");
+            if (log.isTraceEnabled()) {
+                log.trace("Sending build result to BPM " + buildResultRest.toString() + ".");
+            } else {
+                log.debug("Sending build result to BPM " + buildResultRest.toLogString(100) + ".");
```
jbartece: Shouldn't we rather print the tail of the log? And maybe smaller part?
I'd propose to do smth like log.tail(10)

matejonnet: I think few times x 100 chars is not too much, now we had few MB of them.
Tail: if you want a complete log you can switch logging to trace.

jbartece: OK.

PR #1539, opened 2017-06-16
Rename module definition files to angular standard convention of

alexcreasy:
*.module.js


alexcreasy:
Hi Martin, take a look a this when you get a moment, I've just renamed the module files from _$modulename to $modulename.module.js in line with angular js standard conventions.

matedo1:
Hi, it looks great! The last thing that could be renamed is restclient -> rest-client

PR #1540, opened 2017-06-19
NCL-2425 Label link component added

matedo1:


ui/app/product/directives/pnc-label-link/pnc-label-link.html
```diff
@@ -0,0 +1,48 @@
...
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<script type="text/ng-template" id="tooltipTemplateMilestone.html">
+  <strong>Start Date: </strong>{{$ctrl.item.startingDate | date:'mediumDate'}}<br>
```
alexcreasy: Could some of these expressions use one time binding to minimise performance impact? I can't see them changing after the page loads. see https://docs.angularjs.org/guide/expression (scroll down to One-time binding).

Just to add I realise we're not doing this elsewhere in the UI, mostly because it didn't exist when we started the project (it came in in 1.3). I'd like to start doing this where possible with new code and refactors though. This is kind of premature optimisation since we don't have any problems caused by excessive watchers, but there's also no point in setting excessive watches when we know the values won't change.

ui/app/product/directives/pnc-label-link/pncLabelLink.js
```diff
@@ -0,0 +1,67 @@
...
+  function Controller(ProductMilestoneDAO) {
+    var $ctrl = this;
+
+    // -- Controller API --
+    
+    $ctrl.isMilestone = _.has($ctrl.item, 'productReleaseId');
+    $ctrl.isRelease   = _.has($ctrl.item, 'productMilestoneId');
+    $ctrl.isPrimary   = $ctrl.item.id === $ctrl.currentId;
+
+    if ($ctrl.isRelease) {
```
alexcreasy: I think this should probably go in the $onInit lifecycle function if it's supposed to run at the controller start. See: https://docs.angularjs.org/guide/component (there's a section on lifecycle events part way down)

PR #1544, opened 2017-06-23
RepositoryConfiguration introduced - model & rest

jbartece:
Not ready to be merged yet.


build-coordinator/src/test/java/org/jboss/pnc/coordinator/test/AbstractDependentBuildTest.java
```diff
@@ -34,14 +34,7 @@
...
 import org.jboss.pnc.mock.repository.UserRepositoryMock;
-import org.jboss.pnc.model.Artifact;
-import org.jboss.pnc.model.BuildConfiguration;
-import org.jboss.pnc.model.BuildConfigurationAudited;
-import org.jboss.pnc.model.BuildConfigurationSet;
-import org.jboss.pnc.model.BuildRecord;
-import org.jboss.pnc.model.BuildStatus;
-import org.jboss.pnc.model.IdRev;
-import org.jboss.pnc.model.Project;
+import org.jboss.pnc.model.*;
```
matejonnet: Please reconfigure your IDE to not use the wild-card imports.

jbartece: OK. Will do

datastore/src/main/java/org/jboss/pnc/datastore/repositories/RepositoryConfigurationRepositoryImpl.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+@Stateless
+public class RepositoryConfigurationRepositoryImpl extends AbstractRepository<RepositoryConfiguration, Integer> implements RepositoryConfigurationRepository {
+
+    /**
+     * @deprecated Created for CDI.
+     */
+    @Deprecated
+    public RepositoryConfigurationRepositoryImpl() {
+    }
```
janinko: Why to add this constructor? When the class has interface, we should be injecting the interface.

model/src/main/java/org/jboss/pnc/model/RepositoryConfiguration.java
```diff
@@ -0,0 +1,157 @@
...
+    private String internalScmRepoUrl;
+
+    /**
+     * URL of an upstream SCM repository.
+     * This URL SHOULD be read-only, since push access is not needed
+     */
+    @Size(max = 255)
+    @Getter
+    @Setter
+    private String externalScmRepoUrl;
```
janinko: Do we need such long name there? The fact that it's scm repo is implied from the fact that the class is RepositoryConfiguration, no? I would go for externalUrl and internalUrl.

jbartece: Works for me.

datastore/src/main/java/org/jboss/pnc/datastore/repositories/RepositoryConfigurationRepositoryImpl.java
jbartece: As it's required by CDI spec and if it's not there it's not working. It's in all the *RepositoryImpl beans

janinko: No it's not. If the injection point is using the interface, you don't have to have non-parametric constructor on the implementing class. For more info: http://docs.jboss.org/weld/reference/2.4.0.CR1/en-US/html/injection.html#_client_proxies

jbartece:
@matejonnet Ready for review. 
The only thing missing: I had issues with making the relation BC -> RC mandatory on the entity level. For some reason I cannot use JoinColumn annotation and I'm forced to use the ForeignKey annotation. I need to investigate that deeper, but we can merge it also without this thing fixed. I will solve it in the different PR.

jbartece:
@jdcasey @pkocandr Do you have any idea what's the repo-driver fail about? It looks to me like the latest Indy snapshot has some bug.

jbartece:
retest this please

datastore/src/main/java/org/jboss/pnc/datastore/repositories/RepositoryConfigurationRepositoryImpl.java
matejonnet: Currently used CDI version has a bug, that's why we need no-arg constructors.

pom.xml
```diff
@@ -1479,16 +1479,6 @@
...
-      <id>jboss-snapshots</id>
-      <url>https://repository.jboss.org/nexus/content/repositories/snapshots</url>
-      <snapshots>
-        <enabled>true</enabled>
-      </snapshots>
-      <releases>
-        <enabled>false</enabled>
-      </releases>
-    </repository>
-    <repository>
```
matejonnet: Why removing the JBoss snapshots repo and keep Sonatype snapshots ?
PNC shapshots are published at JBoss one.

model/src/main/java/org/jboss/pnc/model/RepositoryConfiguration.java
```diff
@@ -0,0 +1,157 @@
...
+    @SequenceGenerator(name = SEQ_NAME, sequenceName = SEQ_NAME, allocationSize = 1)
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = SEQ_NAME)
+    private Integer id;
+
+    /**
+     * Repository URL containing project to be build.
+     * This URL *MUST* be read/write.
+     */
+    @Size(max = 255)
+    @Column(unique = true, nullable = false, updatable = false)
```
jbartece: The RW is meant to the repository itself not the DB column. TBH I copied the comment from the BC entity, but I think it makes sense to be there.

pom.xml
jbartece: We need Sonatype repo for the Indy snapshots. We don't need the JBoss one as we build all our components. Sometimes it was happening to me, that I wanted to run tests for e.g. rest and Maven started downloading PNC snapshots from the repo instead of using the local one.
Do you think we need the JBoss snapshots repo?

jbartece:
Indy test disabled because of a bug in Indy snapshot. NCL-3104 created to re-enable it again

rest-model/src/main/java/org/jboss/pnc/rest/restmodel/RepositoryConfigurationRest.java
```diff
@@ -0,0 +1,101 @@
...
+    private String internalUrl;
+
+    @Getter
+    @Setter
+    @ScmUrl(groups = {WhenUpdating.class, WhenCreatingNew.class} )
+    private String externalUrl;
+
+    @Getter
+    @Setter
+    private boolean preBuildSyncEnabled;
```
janinko: I would prefer there Boolean or defalt value, because when creating "default" repository, I want to keep the decision of what  the default values are on PNC.

rest-model/src/main/java/org/jboss/pnc/rest/restmodel/RepositoryConfigurationRest.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @NotNull(groups = WhenUpdating.class)
+    @Null(groups = WhenCreatingNew.class)
+    private Integer id;
+
+    @Getter
+    @Setter
+    @NotBlank(groups = {WhenUpdating.class, WhenCreatingNew.class})
+    @ScmUrl(groups = {WhenUpdating.class, WhenCreatingNew.class} )
+    private String internalUrl;
```
janinko: What if I want to create repository just from external url? This requires that the internal url is always set, no?

model/src/main/java/org/jboss/pnc/model/RepositoryConfiguration.java
matejonnet: Current combination is a bit confusing, it looks like one side is wrong. What about moving the comment to the RC or re-format it a bit.

pom.xml
matejonnet: Having it there it allows someone to build one module only, but that is probably not required.
It might be useful in case we depend on some other JBoss snapshots.
Yours case that the build started to download the dependencies was probably due to outdated local snapshots.
I would leave it in.

model/src/main/java/org/jboss/pnc/model/RepositoryConfiguration.java
```diff
@@ -0,0 +1,157 @@
...
+    @Getter
+    @Setter
+    private String externalUrl;
+
+    /**
+     * Declares whether the pre-build repository synchronization should happen or not.
+     */
+    @Getter
+    @Setter
+    private boolean preBuildSyncEnabled = false;
```
jbartece: I think it was meant from the UI side. From the backend side I wanted to keep the defensive approach not to enable anything until it's explicitly set.

rest-model/src/main/java/org/jboss/pnc/rest/restmodel/RepositoryConfigurationRest.java
jbartece: Then you will use the async process, which will create the RC entity and the end, and not the basic CRU methods.

rest-model/src/main/java/org/jboss/pnc/rest/restmodel/RepositoryConfigurationRest.java
jbartece: You can set the value without any issue there. Why to have there Boolean?
The default value is set on the DB level. No need to have it in both locations IMHO

pom.xml
jbartece: As it's not required, I don't see a point in leaving it in. We typically don't rely on snapshot dependencies. 
If you think, it's important I can add it in.

model/src/main/java/org/jboss/pnc/model/RepositoryConfiguration.java
jbartece: Yes, I agree. It's not the best.

So, what about:
/**
URL to the internal Git repository, which is the main repository used for the builds. New commits are being added to this repository, during the builds
*/
internal

/**
URL to the upstream repository.
*/
external

model/src/main/java/org/jboss/pnc/model/RepositoryConfiguration.java
```diff
@@ -0,0 +1,157 @@
...
+     */
+    @Size(max = 255)
+    @Column(unique = true, nullable = false, updatable = false)
+    @Getter
+    @Setter
+    private String internalUrl;
+
+    /**
+     * URL of an upstream SCM repository.
+     * This URL SHOULD be read-only, since push access is not needed
```
jbartece: Yeah, I agree. Let's change. Feel free to provide your opinion above

jbartece:
@janinko Yes, there are some TODOs, which I will resolve in the next PR, but I don't want to unblock others to work on other parts of this epic

pom.xml
janinko: > Having it there it allows someone to build one module only, but that is probably not required.
As I believe the coupling between modules is pretty tight, I wouldn't recommend building one module without building the dependent modules first.

rest-model/src/main/java/org/jboss/pnc/rest/restmodel/RepositoryConfigurationRest.java
janinko: Because I want to send just { "externalUrl": "foo.bar/baz.git" } and let PNC decide the defaults.

model/src/main/java/org/jboss/pnc/model/RepositoryConfiguration.java
janinko: It seems wrong to me to have default true somewhere and default false somewhere else.

rest-model/src/main/java/org/jboss/pnc/rest/restmodel/RepositoryConfigurationRest.java
jbartece: OK, I'll set it here too

model/src/main/java/org/jboss/pnc/model/RepositoryConfiguration.java
jbartece: OK, I'll change it to default true

model/src/main/java/org/jboss/pnc/model/RepositoryConfiguration.java
matejonnet: Look ok, maybe with small change:
"URL to the internal Git repository, which is the main repository used for the builds. New commits can be added to this repository, during the pre-build steps of the build process."

pom.xml
matejonnet: ok, remove it.

model/src/main/java/org/jboss/pnc/model/RepositoryConfiguration.java
jbartece: OK

PR #1549, opened 2017-07-10
[NCL-3028] Submit preBuildSyncEnabled param to bpm

thescouser89:
This param will be used on the BPM/Maitai side to decide which REST
endpoint / parameter to use on the Repour side to run the adjust + sync
on the repositories.


matejonnet:
Thanks for the update.

PR #1573, opened 2017-07-27
NCL-3194 Add filtering of artifacts by quality

alexcreasy:


alexcreasy:
@matejonnet this PR fixes the UI filtering system to use "==" comparator instead of "like" for enum types.

PR #1576, opened 2017-07-31
Remove periodical sync and update RC creation validation.

matejonnet:


rest/src/main/java/org/jboss/pnc/rest/endpoint/BpmEndpoint.java
```diff
@@ -382,17 +380,29 @@ private void addBuildConfigurationToSet(BuildConfiguration buildConfiguration, S
...
+     */
     private void validate(RepositoryCreationRest repositoryCreationRest)
             throws EmptyEntityException, InvalidEntityException {
 
         RepositoryConfigurationRest repositoryConfiguration = repositoryCreationRest.getRepositoryConfigurationRest();
 
-        ValidationBuilder.validateObject(repositoryConfiguration, WhenCreatingNew.class)
-                .validateNotEmptyArgument()
-                .validateAnnotations();
+        if (repositoryCreationRest == null) {
```
janinko: You check null after dereferencing it on previous line

matejonnet: Well spotted, fixing it.

PR #1586, opened 2017-08-07
NCL-3032 Use only single SCM Url and Revision when importing new Product

matedo1:


janinko:
Be aware that ``externalScmUrl`` and ``externalScmRevision`` fields were removed from the json payload.

PR #1588, opened 2017-08-08
Query for RC against normalized scm url.

matejonnet:


model/src/main/java/org/jboss/pnc/model/RepositoryConfiguration.java
```diff
@@ -61,6 +79,15 @@
     private String internalUrl;
 
     /**
+     * Normalized version of scm url to query against. Normalized version is without the protocol and .git extension.
+     */
+    @Size(max = 255)
+    @Column(unique = true, updatable = false)
+    @Getter
+    @Setter
+    private String internalUrlNormalized;
```
jbartece: Why do you want to create another field just to strip out the suffix?
I think it's unnecessary data duplication.

I'd just update the URL to strip it out and store it normalized

matejonnet: The original one is required for operations, without it we don't know what the protocol is and what the extension is.

jbartece: OK. Let's keep it in

PR #1592, opened 2017-08-11
[NCL-3027] Capture removedRepositories information

thescouser89:
The information will be sent from repour to PNC via the BPM process

### DIFF
--- a/spi/src/main/java/org/jboss/pnc/spi/repour/RemovedRepository.java
+++ b/spi/src/main/java/org/jboss/pnc/spi/repour/RemovedRepository.java
@@ -15,27 +15,24 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jboss.pnc.mock.repour;
+package org.jboss.pnc.spi.repour;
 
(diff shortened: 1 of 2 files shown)


thescouser89:
As far as data from repour is concerned, I believe `RepourResult` is the place to pass data from repour to PNC via BPM process. The BPM process also has a `RepourResult` class, very similar (if not identical) to the one of PNC. The BPM process gets the data from repour, puts the relevant data into the RepourResult class, then sends it back to PNC.

I've checked on the BPM process, and yes you are right. The BuildExecutionConfigration is sent from Maitai when the alignment is done and the build started, and RepourResult is sent at the end of the build.

So I guess if you want the removed repositories list when the alignment is done, your PR makes more sense than mine.

pkocandr:
Yes, we need it before the maven build starts. The info in it is used to create the removed repositories in Indy and make them part of the build group.

thescouser89:
@pkocandr Gotcha! I'll close the ticket and wait for your PR to be merged before doing modifications to the BPM process. Thanks for the review! :)

PR #1600, opened 2017-08-25
[NCL-3277] Use correct class to deserialize ArtifactRepository

thescouser89:


rest-model/src/main/java/org/jboss/pnc/rest/restmodel/BuildExecutionConfigurationRest.java
```diff
@@ -272,7 +272,14 @@ public void setSystemImageType(SystemImageType systemImageType) {
         return artifactRepositories;
     }
 
-    public void setArtifactRepositories(List<ArtifactRepository> artifactRepositories) {
+    public void setArtifactRepositories(List<ArtifactRepositoryRest> artifactRepositoriesRest) {
+
+        List<ArtifactRepository> artifactRepositories = new ArrayList<>();
```
pkocandr: Do we need a list copy here?
If so, I would simply use
```
        this.artifactRepositories = new ArrayList<>(artifactRepositoriesRest);
```
if not, I would just cast the parameter to a List with no type to be able to assign it to the field
```
        this.artifactRepositories = (List) artifactRepositoriesRest;
```
This manual cloning seems a bit awkward to me.

Other than that I agree with this fix. I didn't realize the setter is not part of the interface before. Thanks!

PR #1603, opened 2017-08-28
[NCL-3111] archived buildconfig removed in buildconfigset

thescouser89:
This commit attempts to solve NCL-3111, where triggering a build group
will also start the 'deleted' build configurations.

If the user 'deletes' a build configuration, the build configuration is
actually archived, but still present in the database. However, the build
configuration set still points to the build configuration archived.

When the build configuration set build is triggered, Orch gets all the
build configurations for this build configuration set and builds it even
though some of those build configurations are archived.

So the possible solutions are:
- alter the build configurations built for a build configuration set to
  ones that are not archived
- delete the link between the build configuration and its build
  configuration sets when it gets archived

This commit chooses the second solution as the least instrusive
solution.

The first solution would require us to carefully choose whether to
return the non-archived build configurations or *all* the build
configurations, depending on the situation, and this could get really
messy.

The second solution is simpler; simply remove the link. The link
information for previous build configuration set builds is still
preserved in the build records and buildconfig set records.

This commit will also require a cleanup to be run on existing PNC
databases to unlink all archived build configurations with their build
configuration sets.

### DIFF
--- a/rest/src/main/java/org/jboss/pnc/rest/provider/BuildConfigurationProvider.java
+++ b/rest/src/main/java/org/jboss/pnc/rest/provider/BuildConfigurationProvider.java
@@ -24,6 +24,7 @@ import org.jboss.pnc.common.json.moduleconfig.ScmModuleConfig;
 import org.jboss.pnc.common.json.moduleprovider.PncConfigProvider;
 import org.jboss.pnc.model.BuildConfiguration;
 import org.jboss.pnc.model.BuildConfigurationAudited;
+import org.jboss.pnc.model.BuildConfigurationSet;
 import org.jboss.pnc.model.IdRev;
 import org.jboss.pnc.model.ProductVersion;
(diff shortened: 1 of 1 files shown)


thescouser89:
I'm expecting this PR to perhaps not get merged because the solution is not good enough. I tried the first solution, but there were some places where it was inconsistent; some places would get the non-archived build configurations, and other places would get all the build configurations, even though it should have got the non-archived build configurations.

Instead of trying to untangling this, I thought that simply deleting the link would be a cleaner solution. But again, I don't know if maintaining the link when the build configuration is archived was actually intentional.

thescouser89:
retest this please

michalszynkiewicz:
I think we're OK removing the BC from BCS.
After a group build is done, we store BuidlRecords in BuildConfigSetRecord, so I don't think we lose any data by removing BuildConfig from BuildConfigSet.

:+1: from me

PR #1605, opened 2017-08-30
[NCl-3281] Remove uniqueness constraint on externalurlnormalized

thescouser89:
Given that `externalurl` is used to build `externalurlnormalized`, and
that `externalurl` doesn't have any uniqueness constraint,
`externalurlnormalized` should also not have any uniqueness constraint

### DIFF
--- a/model/src/main/java/org/jboss/pnc/model/RepositoryConfiguration.java
+++ b/model/src/main/java/org/jboss/pnc/model/RepositoryConfiguration.java
@@ -93,7 +93,6 @@ public class RepositoryConfiguration implements GenericEntity<Integer> {
     @Size(max = 255)
     @Getter
     @Setter
-    @Column(unique = true)
     private String externalUrl;
 
     /**


janinko:
The external url should be unique, but allow multiple null fields.

If I look correctly, I see that both ``externalUrl`` and ``externalUrlNormalized`` have ``@Column(unique = true)``.

janinko:
Also [Postgresql doc](https://www.postgresql.org/docs/9.0/static/indexes-unique.html) says:
> When an index is declared unique, multiple table rows with equal indexed values are not allowed. **Null values are not considered equal.**

So that should work.

thescouser89:
@janinko Yeah I think you're right. I also removed the @Unique annotation in the wrong location (should have removed it in externalUrlNormalized; stupid tired brain). I see that the Unique thing got added recently in `externalUrl`, and the migration script assumed that it was not unique (it was written before that change was made).

Thanks for the review! I'll close that PR.

PR #1609, opened 2017-09-01
[NCL-3304] Create BC button disabled when user not logged in

thescouser89:


thescouser89:
I'm not sure what's up with the app.html changes... if it's expected or no. @alexcreasy ?

thescouser89:
ncl-3304

matedo1:
@thescouser89 Thanks for this change! It would be a little bit cleaner to commit app.html changes separately, but anyway it's not a big problem as I also wanted to push them soon.

alexcreasy:
+1 thanks Dustin! re: the random crap in index.html, ideally we would only include dependencies in our minified distributions and use a temp index.html for development, I spent 20 minutes trying to implement it once and it wasn't as simple as it should have been so I had to give up.

PR #1624, opened 2017-09-07
[NCL-3326] Set release status to in-progress

thescouser89:
When the milestone release process is started, its status is not set to
IN_PROGRESS. As such, the status is set to null while the release
process is being performed.

This commit puts the status to IN_PROGRESS just before the BPM process
is triggered to rectify this.


thescouser89:
reminder to @thescouser89, backport to 1.2 if merged

thescouser89:
wow @thescouser what an amazing PR, I'll merge it right away.

PR #1625, opened 2017-09-07
[NCL-2999] Once a milestone closed, no more mods

thescouser89:
Once a milestone is closed, we shouldn't allow the user to do any
further modifications to the closed milestone.

This commit adds a check on the backend repository for product
milestone (which the product milestone endpoint uses) to check if the
existing milestone in the database already has an end date set. If so,
that means that the milestone is 'released', and no more modifications
should be allowed.

The web UI is also updated to:
- disable actions if the end date is set for released milestones

### DIFF
--- a/rest/src/main/java/org/jboss/pnc/rest/provider/ProductMilestoneProvider.java
+++ b/rest/src/main/java/org/jboss/pnc/rest/provider/ProductMilestoneProvider.java
@@ -26,6 +26,7 @@ import org.jboss.pnc.rest.validation.ConflictedEntryValidator;
 import org.jboss.pnc.rest.validation.ValidationBuilder;
 import org.jboss.pnc.rest.validation.exceptions.ConflictedEntryException;
 import org.jboss.pnc.rest.validation.exceptions.InvalidEntityException;
+import org.jboss.pnc.rest.validation.exceptions.RepositoryViolationException;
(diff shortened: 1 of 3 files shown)


thescouser89:
milestone_release

Screenshot above shows behaviour when end date (aka release date) is set, and the behaviour when it's not.

thescouser89:
Note to @dcheung: backport to v1.2 branch

thescouser89:
@matejonnet I'm throwing RepositoryViolationException if a milestone is already released and we are trying to modify the milestone, but I'm unsure if that's the right Exception class to use. Let me know if I should create another Exception class for this, or use another existing Exception class.

thescouser89:
retest this please

janinko:
What if the import to brew fails? Can we still retry (by hitting the close button againt)?

thescouser89:
@janinko No retries by hitting the close button again.

janinko:
Unless the milstone status is SUCCESS the milestone shouln't be considered done

thescouser89:
@janinko Alright I'll see what I can do.

thescouser89:
Would it be better to have a dedicated 'Retry to import to brew' button instead of using the Close button to get the retry?

janinko:
Dunno. Wasn't there an idea that the close process would do more things in the future then just import to brew

PR #1644, opened 2017-09-18
[NCL-3112][NCL-2999] Close milestone process

thescouser89:
This commit applies changes to the close milestone process related to
NCL-3112 and NCL-2999.

A milestone is considered closed when the release process is successful
and the end date for this milestone is set.

We can initiate the close milestone process by hitting the
`/close-milestone` endpoint. Optionally, we can specify the
`downloadUrl` in that endpoint.

When the `/close-milestone` endpoint is hit, it will start the release
process. If the release process for the milestone is successful, the
following happens:

- the `endDate` of the milestone is set to the date the process release
  succeeded
- if the milestone is set as 'current' in the product version, it is
  then unset as 'current'

On the Web UI, if the end date of a milestone is set, then the options
to:

- set as current
- edit milestone
- close milestone

are disabled.

### DIFF
--- a/rest/src/main/java/org/jboss/pnc/rest/endpoint/ProductMilestoneEndpoint.java
+++ b/rest/src/main/java/org/jboss/pnc/rest/endpoint/ProductMilestoneEndpoint.java
@@ -203,6 +203,30 @@ public class ProductMilestoneEndpoint extends AbstractEndpoint<ProductMilestone,
         return Response.ok().build();
     }
 
+    @ApiOperation(value = "Close/Release a Product Milestone")
+    @ApiResponses(value = {
+            @ApiResponse(code = SUCCESS_CODE, message = SUCCESS_DESCRIPTION),
+            @ApiResponse(code = INVALID_CODE, message = INVALID_DESCRIPTION, response = ErrorResponseRest.class),
+            @ApiResponse(code = CONFLICTED_CODE, message = CONFLICTED_DESCRIPTION, response = ErrorResponseRest.class),
+            @ApiResponse(code = SERVER_ERROR_CODE, message = SERVER_ERROR_DESCRIPTION, response = ErrorResponseRest.class)
+    })
+    @PUT
+    @Path("/{id}/close-milestone")
+    public Response closeMilestone(
+            @ApiParam(value = "Product Milestone id", required = true) @PathParam("id") Integer id,
+            ProductMilestoneRest productMilestoneRest,
+            @Context HttpServletRequest httpServletRequest) throws ValidationException {
+
+        if (httpServletRequest != null) {
+            LoggedInUser loginInUser = authenticationProvider.getLoggedInUser(httpServletRequest);
+            productMilestoneProvider.closeMilestone(id, productMilestoneRest, loginInUser.getTokenString());
+        } else {
+            productMilestoneProvider.closeMilestone(id, productMilestoneRest);
+        }
+
+        return Response.ok().build();
+    }
+
     @ApiOperation(value = "Get the artifacts distributed in this milestone")
     @ApiResponses(value = {
             @ApiResponse(code = SUCCESS_CODE, message = SUCCESS_DESCRIPTION, response = ArtifactPage.class),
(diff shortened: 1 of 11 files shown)


thescouser89:
retest this please

thescouser89:
retest this please

thescouser89:
Also related: NCL-3362

thescouser89:
Please review @jbartece  :)

rest/src/main/java/org/jboss/pnc/rest/provider/ProductMilestoneProvider.java
```diff
@@ -137,30 +135,52 @@ public void update(Integer id, ProductMilestoneRest restEntity) throws Validatio
...
-            if (productVersion.getCurrentProductMilestone() != null &&
-                    productVersion.getCurrentProductMilestone().getId().equals(milestoneInDb.getId())) {
+        ProductMilestone milestoneInDb = repository.queryById(id);
 
-                productVersion.setCurrentProductMilestone(null);
-                productVersionRepository.save(productVersion);
-            }
+        // we can't modify milestone if it's already released
+        if (milestoneInDb.getEndDate() != null) {
+            log.error("Milestone is already closed: no more modifications allowed");
```
jbartece: The log level error doesn't look appropriate to me. If the user tries to close already closed milestone, it's not a PNC issue. Info level should be fine here.

rest/src/main/java/org/jboss/pnc/rest/endpoint/ProductMilestoneEndpoint.java
```diff
@@ -203,6 +203,30 @@ public Response update(
...
 
+    @ApiOperation(value = "Close/Release a Product Milestone")
+    @ApiResponses(value = {
+            @ApiResponse(code = SUCCESS_CODE, message = SUCCESS_DESCRIPTION),
+            @ApiResponse(code = INVALID_CODE, message = INVALID_DESCRIPTION, response = ErrorResponseRest.class),
+            @ApiResponse(code = CONFLICTED_CODE, message = CONFLICTED_DESCRIPTION, response = ErrorResponseRest.class),
+            @ApiResponse(code = SERVER_ERROR_CODE, message = SERVER_ERROR_DESCRIPTION, response = ErrorResponseRest.class)
+    })
+    @PUT
+    @Path("/{id}/close-milestone")
```
jbartece: What endpoint was used before to close the milestone and start the milestone release process?

rest/src/main/java/org/jboss/pnc/rest/provider/ProductMilestoneProvider.java
```diff
@@ -137,30 +135,52 @@ public void update(Integer id, ProductMilestoneRest restEntity) throws Validatio
...
+    }
+
+    public void closeMilestone(Integer id, ProductMilestoneRest restEntity, String accessToken) throws ValidationException {
+        ProductMilestone milestoneInDb = repository.queryById(id);
+
+        // If we want to close a milestone, make sure it's not already released (by checking end date)
+        // and there are no release in progress
+        if (milestoneInDb.getEndDate() != null) {
+
+            log.error("Milestone is already closed: no more modifications allowed");
```
jbartece: Also here info level looks enough to me.

rest/src/main/java/org/jboss/pnc/rest/endpoint/ProductMilestoneEndpoint.java
thescouser89: Before we used the `update` endpoint. We had to set the end date to start the milestone release process.

Since we now set the end date to the date when the release is done, I thought it'd be better to create a specific endpoint to just close the milestone.

It'll check that:
- the end date is not already set (which would mean the release is already done)
- there are no release in progress (there would be a release in progress if we hit the 'Close Milestone' button 2 times in quick succession). If there is a release in progress, do nothing.
- If there are no release in progress, trigger the release
- In the release process code, if the process is successful, it'll set the end date

thescouser89:
@jbartece Noted for the logging level. I have changed my commit so that it's at info level now. Thanks for the review! 😃

rest/src/main/java/org/jboss/pnc/rest/endpoint/ProductMilestoneEndpoint.java
jbartece: Great! That's a good way of doing that.

thescouser89:
Merged to v1.2 branch also

PR #1654, opened 2017-09-21
[NCL-3364] Remove /bpm/tasks/start-repository-configuration-creation endpoint (master)

janinko:


janinko:
Note that in this PR there is one commit more (then in the PR to branch)

janinko:
fixed tests, @matejonnet you can teake a look now

PR #1678, opened 2017-10-18
Fix bower

thescouser89:
It looks like since Oct 11, the old bower registry got deprecated.
https://twitter.com/bower/status/918073147789889536

This causes the PNC build to fails because the ui/bower component cannot
talk to the mothership.

This PR updates our .bowerrc to specify the new registry to use.

### DIFF
--- a/ui/.bowerrc
+++ b/ui/.bowerrc
@@ -4,5 +4,6 @@
     "packages": ".build-tmp/bower-cache",
     "registry": ".build-tmp/bower-registry"
   },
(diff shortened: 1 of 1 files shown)


alexcreasy:
@thescouser89 technically bower itself is deprecated, at some point we need to look to moving to NPM exclusively, but it's a bit of work retrofitting a module system in!

thescouser89:
Thanks for the review @alexcreasy :)

PR #1693, opened 2017-11-02
[NCL-3442]:  Build record rest endpoint supporting OR based filtering

jbartece:


jbartece:
Btw. The ordering of commits shown here is misleading and a bug on GitHub. The 1fade68 commit is a HEAD of the branch, but is shown as second, because I did an interactive rebase :-)

PR #1702, opened 2017-11-20
[NCL-3504] Add build options for temp build

thescouser89:
The new parameters 'tempBuild' and 'tempBuildTimestamp' are added to
`BuildExecutionConfiguration` and `BuildExecutionConfigurationRest`.

It is assumed that the 2 parameters are ultimately stored inside
`BuildConfiguration`.


thescouser89:
I created this JIRA as part of NCL-3507. I think the most controversial part about this will be the storing of the 2 new parameters in BuildConfiguration class.

This commit does not add logic on how to populate `tempBuildTimestamp`. It also does not provide any option to set those 2 parameters from UI.

thescouser89:
#1706 is a better implementation of this PR. Closing this one.

PR #1703, opened 2017-11-22
Ncl 3508 storing of the temporary builds

matejonnet:


matejonnet:
Re-test this please.

model/src/main/java/org/jboss/pnc/model/TargetRepository.java
```diff
@@ -0,0 +1,160 @@
...
+     * Types of artifact repositories
+     *
+     * Types *_TEMPORAL are repository groups for temporal build (Snapshots / Pull Request)
+     */
+    public enum Type {
+        /**
+         * Maven artifact repository such as Maven central (http://central.maven.org/maven2/)
+         */
+        MAVEN,
+        MAVEN_TEMPORAL,
```
jbartece: I think that this name is not ideal as it means "related to time" :-). Temporal databases are used to store time related data. The work temporary is not ideal, but I'd say better.

PR #1706, opened 2017-11-24
[NCL-3504]: Add build option "temporary" and "add-timestamp" - part 1

jbartece:
Part 1 - refactor to the object BuildOptions. UI part unblocked


matejonnet:
Functional it looks good.
Please update the formatting to follow the convention: when method parameters and in multiple lines, there have to be each parameter in its own line.

jbartece:
OK, I'll update it.

PR #1718, opened 2017-11-30
[NCL-3506] Configure build repos for temporary builds

pkocandr:


pkocandr:
retest this please

spi/src/main/java/org/jboss/pnc/spi/executor/BuildExecutionConfiguration.java
```diff
@@ -58,9 +58,11 @@
 
     Map<String, String> getGenericParameters();
 
-    boolean isTempBuild();
+    @Override
+	boolean isTempBuild();
 
-    String getTempBuildTimestamp();
+    @Override
+	String getTempBuildTimestamp();
```
jbartece: ehm, broken formatting, but thanks for adding the Override. I forgot obviously.

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
```diff
@@ -245,9 +242,11 @@ private void setupBuildRepos(BuildExecution execution, Indy indy)
         if (!indy.stores().exists(groupKey)) {
             // if the product-level storage repo (for in-progress product builds) doesn't exist, create it.
             StoreKey hostedKey = new StoreKey(MAVEN_PKG_KEY, StoreType.hosted, buildContentId);
+            boolean tempBuild = execution.isTempBuild();
+			boolean snapshotBuild = tempBuild && StringUtils.isEmpty(execution.getTempBuildTimestamp());
```
jbartece: Not all temporary builds are builds with timestamp. It's just one of the options.
So, it should be:
boolean snapshotBuild = tempBuild

And ehm, formatting.

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
```diff
@@ -257,13 +256,19 @@ private void setupBuildRepos(BuildExecution execution, Indy indy)
             }
 
             Group buildGroup = new Group(MAVEN_PKG_KEY, buildContentId);
-            buildGroup.setDescription(String.format("Aggregation group for PNC build #%s", id));
+            String adjective = "";
+            if (snapshotBuild) {
+                adjective = "snapshot ";
+            } else if (tempBuild) {
+                adjective = "temporary ";
+            }
```
jbartece: We don't have anything like snapshotBuild, we just have a temporary builds with, which has less restrictions as for e.g. snapshots are allowed. So, I'd not distinguish it in the name as it will apply to all temporary builds

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
pkocandr: Argh, sorry, stupid new eclipse, that tabs vs spaces irritates me.

Regarding snapshot build isn't it a temporary build which has an empty timestamp?

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
pkocandr: This is just to make the repo description more specific. Why not to do so?

spi/src/main/java/org/jboss/pnc/spi/executor/BuildExecutionConfiguration.java
pkocandr: Override is added automatically by eclipse, you should migrate back ;-) But the automatic addition used again a tab instead of 4 spaces :-/

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
jbartece: Temporary build is a build with the flag tempBuild.

But it doesn't say anything about the alignment.
It can be normal snapshot, it can be with timestamp, it can have standard rh suffix and nothing else, there can be mix of it. We basically don't care and it's up to the user to configure it. So, we're just saying, we will also allow snapshots as it doesn't hurt us.
And for the timestamp build it's expected that there will be no snapshots as we change it in the alignment (if it's not affected somehow by PME params), but for the builds without timestamp, it's more probable the snapshots could theoretically be there.
Basically, we don't know it on this level and we don't have to, so for all the temp builds, we will just allow the snapshots.

And btw. the timestamp is not empty, it's null.

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
jbartece: OK, it doesn't hurt, so whatever, but we don't have the differentiation of snapshot and tempBuild, so tempBuild = snapshotBuild in this case, once the changes above are applied

PR #1737, opened 2017-12-06
Make timestamp generation prettier

thescouser89:
Currently, if the month, day, or time is a single digit, the timestamp
generated will be:

```
t<yyyy>-<m>-<d>-<h>-<M>-<s>-<f>
```

This commit changes this so that we add extra '0' where necessary. The
timestamp then looks like this:

```
t<yyyy>-0<m>-0<d>-0<h>-0<M>-0<s>-00<f>
```

E.g 't20170803-010305-009'

### DIFF
--- a/common/src/test/java/org/jboss/pnc/common/test/util/TimeUtilsTest.java
+++ b/common/src/test/java/org/jboss/pnc/common/test/util/TimeUtilsTest.java
@@ -38,12 +38,27 @@ public class TimeUtilsTest {
     @Test
     public void generateTimestampValidTest() {
         // given
-        Date date = new Date(1512050025464L); //Nov 30, 2017 2:53:45 PM (2017-11-30-14:53:45)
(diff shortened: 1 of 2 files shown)


common/src/test/java/org/jboss/pnc/common/test/util/TimeUtilsTest.java
```diff
@@ -38,12 +38,19 @@ public void generateTimestampDisabledTest() {
...
         String timestamp = TimeUtils.generateTimestamp(true, date);
 
         //then
         assertEquals("t20171130-135345-464", timestamp);
+
+
+        // test that TimeUtils appends an extra '0' when month, day, or time is a single digit
+        Date date2 = new Date(1501722185009L); //Aug 03, 2017 1:03:05 AM (2017-08-03-01:03:05) UTC (9 milliseconds)
+        String timestamp2 = TimeUtils.generateTimestamp(true, date2);
+        assertEquals("t20170803-010305-009", timestamp2);
```
jbartece: Could you create a new test for it, please? It shouldn't be in one ;-).

PR #1756, opened 2017-12-15
[NCL-3514] Don't show deprecated environments

thescouser89:
This commit adds the field 'deprecated' to the 'BuildEnvironment' model
to indicate which build environments are deprecated and shouldn't be
used in current/future build environments.

Note that this commit doesn't add a restriction on the backend on
assigning a deprecated build environment to a build configuration. This
is done so that you can still modify an existing build configuration
using deprecated build environment.


thescouser89:
JIRA has been created to use the 'deprecated' value to show/hide the build environment: NCL-3565

PR #1780, opened 2018-01-26
NCL-3485 Brew push log is now fetched each time user clicks on the br…

alexcreasy:
…ew push tab


alexcreasy:
@janinko as per your request.

janinko:
Nice!

PR #1793, opened 2018-02-05
NCL-3428 - NPM support in repository manager

pkocandr:


thescouser89:
retest this please

pkocandr:
retest this please

pkocandr:
It is ready now. I've created the configuration classes in the MavenRepoDriverModuleConfig because they are related to the repo driver module config. They have shared parent PackageTypeSpecificStringLists (couldn't come up with a better name) and the IgnoredPathSuffixes one is adding the shared bits.
It might be also good to rename the module config class to IndyRepoDriverModuleConfig but then it would also need to update the MODULE_NAME from "maven-repo-driver" to "indy-repo-driver". That would require updating all config files and maybe a bit overkill now, so I left it like this for now.

PR #1843, opened 2018-03-12
Update BpmManager not to block read opperation on BuildTasks#2

jbartece:


jbartece:
Updated. Once I did the update, I realized, that with the change, it should be pretty easy to change it to ConcurrentHashMap and get rid of the synchronized blocks as they were used only for inserting and deleting entities from the Map. I added another commit with that.

PR #1861, opened 2018-03-20
NCL-3413 View Build Configuration button fixed

matedo1:


alexcreasy:
@matedo1 great, can you backport this to 1.3 also... it would be good to get this simple fix on 1.2 as well if it's not too late <-- @jbartece wdyt?

jbartece:
1.2.2 was released and we do not plan 1.2.3, so it's not important to backport it there.

PR #1874, opened 2018-03-26
NCL-3657 Mark persistent repositories as ReadOnly

pkocandr:


indy-repository-manager/src/main/java/org/jboss/pnc/indyrepositorymanager/IndyRepositorySession.java
```diff
@@ -525,7 +528,26 @@ public void promoteToBuildContentSet() throws RepositoryManagerException {
         GroupPromoteRequest request = new GroupPromoteRequest(hostedKey, buildPromotionGroup);
         try {
             GroupPromoteResult result = promoter.promoteToGroup(request);
-            if (!result.succeeded()) {
+            if (result.succeeded()) {
+                if (isTempBuild) {
+                    HostedRepository hosted = indy.stores().load(hostedKey, HostedRepository.class);
+                    hosted.setReadonly(true);
```
matejonnet: This seems to be the wrong way around: isTempBuild -> hosted.setReadonly(true)

pkocandr: Argh, I was watching carefully not to invert the condition above checking build success and missed this one :-) Thanks! Will fix and then merge if no other issues found.

PR #1911, opened 2018-04-19
[NCL-3781]: Correct Swagger return type for  /build-configuration-set…

jbartece:
…s/{id}/build


janinko:
not sure if the new type is really correct

jbartece:
Ups. you are right. My bad, I missed the record string in the line, where the response is constructed. Thanks

jbartece:
retest this please

PR #1924, opened 2018-04-23
NCL-3429 NCL-3803

alexcreasy:


alexcreasy:
@matedo1 yes, you're right we need this on the BC detail page and for it to be editable, I made that into a separate issue since being unable to create a BC was holding up Indy testing. Thanks for the review!

PR #1925, opened 2018-04-23
NCL-3810 Fixes: When BC wizard is re-opened after an error the SCM UR…

alexcreasy:
…L is not checked to see if an RC already exists causing a 409 on submit


alexcreasy:
retest this please

PR #1939, opened 2018-05-03
[NCL-3816] Fix predicates for buildconfigset

thescouser89:
The predicate fix is for getting a list of BuildRecord from a build
config set id.

### DIFF
--- a/spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/BuildRecordPredicates.java
+++ b/spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/BuildRecordPredicates.java
@@ -21,6 +21,8 @@ import org.jboss.pnc.model.Artifact;
 import org.jboss.pnc.model.Artifact_;
 import org.jboss.pnc.model.BuildConfigSetRecord;
 import org.jboss.pnc.model.BuildConfigSetRecord_;
+import org.jboss.pnc.model.BuildConfigurationSet;
+import org.jboss.pnc.model.BuildConfigurationSet_;
 import org.jboss.pnc.model.BuildRecord;
 import org.jboss.pnc.model.BuildRecord_;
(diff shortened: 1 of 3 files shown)


thescouser89:
@matejonnet  @janinko @jbartece Because it *seems* I found an old bug from 2015, and I'm not even sure of myself, could you please verify this PR?

It seems there was confusion on build configuration set id and build configuration set record id for NCL-3816, but the commit was done in 2015 and I'm unsure how that bug would have been unnoticed for so long.

Could you also verify my predicate for: `withBuildConfigSetId`? I think I got it right but heh it's Dustin!

The Buildconfigsetrecord has a foreign key on buildconfigurationset. Once we obtain the buildconfigsetrecord, we can then get all the build records linking to those particular buildconfigsetrecord.

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/BuildRecordPredicates.java
```diff
@@ -73,8 +75,12 @@
 
     public static Predicate<BuildRecord> withBuildConfigSetId(Integer buildConfigSetId) {
         return (root, query, cb) -> {
-            Join<BuildRecord, BuildConfigSetRecord> joinedConfiguSet = root.join(BuildRecord_.buildConfigSetRecord);
-            return cb.equal(joinedConfiguSet.get(org.jboss.pnc.model.BuildConfigSetRecord_.id), buildConfigSetId);
```
thescouser89: Here we're confusing build config set id and build config set record id

janinko:
Yeah, it seems weird that there were two methods doing the same. Looked at your change and couldn't find any obvious issue.

thescouser89:
Merging it for further testing in our Devel environment

PR #1940, opened 2018-05-03
[v1.2][NCl-3816] Fix predicates for buildconfigset

thescouser89:
The predicate fix is for getting a list of BuildRecord from a build
config set id.


thescouser89:
retest this please

PR #1953, opened 2018-05-09
Adding metrics reporting using dropwizard metrics

thescouser89:
ose_pnc

### DIFF
--- a/metrics/src/main/java/org/jboss/pnc/metrics/MetricsConfiguration.java
+++ b/metrics/src/main/java/org/jboss/pnc/metrics/MetricsConfiguration.java
@@ -1,0 +1,163 @@
+/**
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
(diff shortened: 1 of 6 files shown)


metrics/src/main/java/org/jboss/pnc/metrics/MetricsConfiguration.java
```diff
@@ -0,0 +1,163 @@
...
+        Graphite graphite = new Graphite(new InetSocketAddress(host, port));
+
+        GraphiteReporter reporter = GraphiteReporter.forRegistry(metricRegistry)
+                .prefixedWith(prefix)
+                .convertRatesTo(TimeUnit.SECONDS)
+                .convertDurationsTo(TimeUnit.MILLISECONDS)
+                .filter(MetricFilter.ALL)
+                .build(graphite);
+
+        reporter.start(1, TimeUnit.MINUTES);
```
matejonnet: I would be useful to have a configurable metrics interval. Maybe in seconds.

thescouser89: I will add it in my next PR for v1.3 and master!

PR #1965, opened 2018-05-21
Remove build logs from buildRecord entity.

matejonnet:


jbartece:
I see this as really problematic change. We should fix the lazy loading properly. It's increasing the complexity and won'ŧ solve all the problems. The lazy loading is used through the application and not only in BuildRecord.
If it doesn't work, it affects the perfomance of the whole application.

PR #1978, opened 2018-05-31
[NCL-3738]: Remove tokens from logs - master

jbartece:


thescouser89:
retest this please

pom.xml
```diff
@@ -782,6 +782,12 @@
         <version>${version.catch-exception}</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-log4j12</artifactId>
```
matejonnet: Any particular reason to use log4j and not slf4j-simple which is used elsewhere?

jbartece: Yes, the log4j is needed by the OpenShift client library and without this dependency the tests were failing with ClassNotFoundError. Btw. The scope for it is only test.

jbartece:
retest this please

PR #1997, opened 2018-06-21
Force cancel in the driver if remote cancellation fails.

matejonnet:


jbartece:
@matejonnet  Is there any PR for master?

PR #2023, opened 2018-07-09
[v1.3] Update arquillian versions to latest

thescouser89:


thescouser89:
Closing because I thought this might have improved the flaky tests. But given that it doesn't help and the upgrade might be risky for the maintenance branch v1.3, I'm closing it.

PR #2031, opened 2018-07-18
Rename module config maven-repo-driver to indy-repo-driver

pkocandr:
to match the maven module name / class name.


pkocandr:
retest this please

pkocandr:
Hmm, it fails consistently while it succeeds locally and it seems there is no reason why it should get broken by changed config module ID.

pkocandr:
retest this please

PR #2037, opened 2018-07-23
[Not merge] Fix indy-weft injection problem during IT

ligangty:
Seems when weft wrapped in the ear, this problem will happen. Not sure why jboss CDI environment can not find weft config @Default bean. But after remove it, I can run the it smoothly on my local env.
@pkocandr you can have this fix in and try on you local (Please clear all cached dependency and built target then try it)


pkocandr:
Was made part of #2035.

PR #2042, opened 2018-07-24
Update version to 2.0.0-SNAPSHOT

pkocandr:
The version was 1.4.0-SNAPSHOT, which could cause issues, because the same version is in branch v1.4.


pkocandr:
retest this please

PR #2072, opened 2018-08-13
[NCL-3702] Cannot delete BuildConfigSet

michalovjan:
Implemented by archival of entity

### DIFF
--- a/rest/src/main/java/org/jboss/pnc/rest/provider/BuildConfigurationSetProvider.java
+++ b/rest/src/main/java/org/jboss/pnc/rest/provider/BuildConfigurationSetProvider.java
@@ -60,6 +60,8 @@ public class BuildConfigurationSetProvider extends AbstractProvider<BuildConfigu
 
     private BuildConfigurationAuditedRepository buildConfigurationAuditedRepository;
 
+    private BuildConfigSetRecordRepository buildConfigSetRecordRepository;
+
     private BuildConfigurationRepository buildConfigurationRepository;
 
     public BuildConfigurationSetProvider() {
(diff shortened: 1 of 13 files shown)


michalovjan:
@thescouser89 Hi Dustin, database migration would be needed. I just added field "active" in BuildConfigurationSet entity.

michalovjan:
retest this please

rest/src/main/java/org/jboss/pnc/rest/provider/BuildConfigurationSetProvider.java
```diff
@@ -110,6 +111,24 @@ protected void validateBeforeSaving(BuildConfigurationSetRest buildConfiguration
...
+        ValidationBuilder.validateObject(WhenUpdating.class).validateAgainstRepository(repository, buildConfigurationSetId,
+            true);
+        BuildConfigurationSet buildConfigurationSet = repository.queryById(buildConfigurationSetId);
+        buildConfigurationSet.setArchived(true);
+
+        // if a build group is archived, unlink the build group from the build configurations is associated with
+        for (BuildConfiguration bc: buildConfigurationSet.getBuildConfigurations()) {
+            bc.removeBuildConfigurationSet(buildConfigurationSet);
+            buildConfigurationRepository.save(bc);
+        }
```
matejonnet: Why removing? Isn't the point of archiving to keep the history record?

thescouser89:
@michalovjan Noted! Thanks for letting me know! :)

rest/src/main/java/org/jboss/pnc/rest/provider/BuildConfigurationSetProvider.java
michalovjan: I discussed this with @jbartece today. My inspiration was the archive method from BCProvider.java ([link](https://github.com/project-ncl/pnc/blob/master/rest/src/main/java/org/jboss/pnc/rest/provider/BuildConfigurationProvider.java#L217)). 

The intended behaviour was the same as in the BCProvider, to delete links from BCS to BC so that the UI would not display archived entity. Kuba noted that direction matters and if I would do the same thing from the other direction here, then it would delete both directions (because of the owning side of relationship). That would be problem because if someone wanted to request archived BCS he would see only name and not BCs which are the most important information that user is looking for. 

We realized that if BCSRecord is attached to BCS, then BCs could be reconstructed if necessary, but if no BCSRecord is attached, we would have basically blank BCS archived in the database. 
We agreed to this solution: 
* If BCS has no records attached, then it would be entirely deleted from DB.
* If BCS has record/s attached, then it would be archived and if someone wanted to see BCs it had, then it would be reconstructed from the last record.

Do you agree with this solution @matejonnet?

matejonnet: Sounds good.

michalovjan:
will add tests after vacation :)

michalovjan:
retest this please

michalovjan:
@thescouser89 Hi Dustin, could you check the migration script? :)

PR #2076, opened 2018-08-13
[v1.4] Use pncmetrics 1.0.1 to enable metrics

thescouser89:
This commit removes the existing Metrics classes that got duplicated
in our codebase and other PNC applications


thescouser89:
retest this please

PR #2125, opened 2018-09-10
[NCL-4089] Add EXECUTION_ROOT_NAME generic param

thescouser89:
Specify the execution root name of the build configuration. This is used
to override the default value, and can be useful for builds that disable
PME. Format is '<groupid>:<artifactid>'


thescouser89:
retest this please

PR #2155, opened 2018-09-25
Fix new atomic builds endpoints not following REST guideliness - 1.4

jbartece:


jbartece:
@alexcreasy For the endpoint "update-and-get-audited" I let the path the same as it returns different entity than POST /{id}. That would not be great to return different entity based on a query parameter

alexcreasy:
@jbartece +1 to just changing the method on update-and-get-audited, it's more explicit in it's intent than using the query param and, you're right multiple return types for one endpoint is confusing.

jbartece:
retest this please

jbartece:
retest this please

PR #2163, opened 2018-10-04
[NCL-3881] Introduce new modules for REST Refactor

janinko:
I created new branch **rest-refactor** based from master, so the PRs can be merged into it and the master branch will keep working without breakage from the refactor. This PR is for that branch.

I included also a few example files, just to be sure I did it correctly.

### DIFF
--- a/facade/pom.xml
+++ b/facade/pom.xml
@@ -1,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    JBoss, Home of Professional Open Source.
+    Copyright 2014-2018 Red Hat, Inc., and individual contributors
+    as indicated by the @author tags.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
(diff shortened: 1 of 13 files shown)


rest/nb-configuration.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project-shared-configuration>
+    <!--
+This file contains additional configuration written by modules in the NetBeans IDE.
+The configuration is intended to be shared among all the users of project and
+therefore it is assumed to be part of version control checkout.
+Without this configuration present, some functionality in the IDE may be limited or fail altogether.
+-->
+    <properties xmlns="http://www.netbeans.org/ns/maven-properties-data/1">
```
jbartece: You should ignore such settings and keep it locally IMHO

facade/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>parent</artifactId>
+        <groupId>org.jboss.pnc</groupId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>facade</artifactId>
```
jbartece: Haven't you agreed on a better name with Matej?

notifications/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+    <description>Websocket notification service.</description>
+  
+  
+    <dependencies>
+      
+        <!-- Project dependencies -->
+        <dependency>
+            <groupId>org.jboss.pnc</groupId>
+            <artifactId>spi</artifactId>
+        </dependency>
```
jbartece: It should communicate using DTOs, so the dependency should be added here.

pom.xml
```diff
@@ -499,6 +541,24 @@
...
+      <dependency>
+        <groupId>io.swagger.core.v3</groupId>
+        <artifactId>swagger-jaxrs2-servlet-initializer</artifactId>
+        <version>2.0.5</version>
+      </dependency>
+
+      <dependency>
+        <groupId>io.swagger.core.v3</groupId>
+        <artifactId>swagger-annotations</artifactId>
+        <version>2.0.5</version>
```
jbartece: A variable would be sensible here.

facade/pom.xml
janinko: No

rest/nb-configuration.xml
janinko: Sorry, that was mistake

jbartece:
@janinko The build failed because of missing header. 
[WARNING] Missing header in: /home/jenkins/workspace/PNC_Pull_Requests/constants/pom.xml

PR #2169, opened 2018-10-10
Add GenericRestEntity

jbartece:


dto/src/main/java/org/jboss/dto/GenericRestEntity.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.dto;
+
+/**
+ * A base interface for all REST entities
+ * to allow generic implementation of some standard operations
+ */
+public interface GenericRestEntity {
```
janinko: just RestEntity

dto/src/main/java/org/jboss/dto/DtoEntity.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.dto;
+
+/**
+ * A base interface for all DTO entities
+ * to allow generic implementation of some standard operations
+ */
+public interface DtoEntity {
```
janinko: DTOEntity :exclamation:

PR #2190, opened 2018-10-19
[NCL-3955] add support to lookup artifact based on checksums

michalovjan:


rest/src/main/java/org/jboss/pnc/rest/provider/ArtifactProvider.java
```diff
@@ -181,6 +185,13 @@ private String getPublicUrl(Artifact artifact) {
         return queryForCollection(pageIndex, pageSize, sortingRsql, query, withDependantBuildRecordId(buildRecordId));
     }
 
+    public CollectionInfo<ArtifactRest> getAll(int pageIndex, int pageSize, String sortingRsql, String query,
+            Optional<String> sha256, Optional<String> md5, Optional<String> sha1) {
+        logger.debug("HELLO + sha1: " + sha1 + " sha256:" + sha256 + " md5: " + md5);
```
jbartece: Better debug message would be nice ;-)

michalovjan: oh, I was supposed to delete that :D you've seen nothing!

PR #2205, opened 2018-10-26
[NCL-4214]: Temporary builds won't be assigned to a milestone anymore - 1.4

jbartece:
### Checklist:

### DIFF
--- a/build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/datastore/DatastoreAdapter.java
+++ b/build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/datastore/DatastoreAdapter.java
@@ -294,11 +294,8 @@ public class DatastoreAdapter {
                 .user(buildTask.getUser())
                 .submitTime(buildTask.getSubmitTime())
                 .startTime(buildTask.getStartTime())
+                .productMilestone(buildTask.getProductMilestone())
                 .temporaryBuild(buildOptions.isTemporaryBuild());
-
-        if (! buildOptions.isTemporaryBuild()) {
-            builder.productMilestone(buildTask.getProductMilestone());
-        }
 
         if (buildTask.getEndTime() == null) {
             buildTask.setEndTime(Date.from(Instant.now()));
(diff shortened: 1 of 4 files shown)


build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/datastore/DatastoreAdapter.java
```diff
@@ -294,9 +294,12 @@ public void storeRejected(BuildTask buildTask) throws DatastoreException {
                 .user(buildTask.getUser())
                 .submitTime(buildTask.getSubmitTime())
                 .startTime(buildTask.getStartTime())
-                .productMilestone(buildTask.getProductMilestone())
                 .temporaryBuild(buildOptions.isTemporaryBuild());
 
+        if (! buildOptions.isTemporaryBuild()) {
```
pkocandr: Not sure if this is the best place to put the logic. Shouldn't this be a concern of the Builder.build() implementation? I.e. if the temporary flag is true on the builder it will ignore the milestone value. That would ensure that temporary builds' records will always be consistent even when initialized anywhere else than in the DatastoreAdapter. OTOH not sure if the logic will ever be implemented elsewhere so not a big deal.

And why is the space after the exclamation mark? It is not supposed to be there! :-)

jbartece: Moved to the model and added also PrePersist check

model/src/main/java/org/jboss/pnc/model/BuildRecord.java
```diff
@@ -338,6 +339,14 @@ public void preRemove() {
             throw new PersistenceException("The non-temporary builds cannot be deleted! Only deletion of temporary builds is supported");
     }
 
+    @PrePersist
+    public void prePersist() {
+        if(this.temporaryBuild && this.productMilestone != null) {
```
pkocandr: Missing space after "if" :-)

model/src/main/java/org/jboss/pnc/model/BuildRecord.java
```diff
@@ -888,6 +896,10 @@ public BuildRecord build() {
             }
             buildRecord.setTemporaryBuild(temporaryBuild);
 
+            if(!temporaryBuild) {
```
pkocandr: And another space nissing after "if" :-)

jbartece: We should define the code style more clearly and use some autoformatter. It's written the same way several lines above.

PR #2207, opened 2018-10-26
[NCL-4117] restricting of editing Product Version name due to having closed milestone

michalovjan:
### Checklist:

* [NCL-4117] added restriction on a user when trying to update product version name in the scenario of version having a closed milestone (both REST and UI)
* [NCL-4117] added integration test

### DIFF
--- a/rest/src/main/java/org/jboss/pnc/rest/endpoint/ProductVersionEndpoint.java
+++ b/rest/src/main/java/org/jboss/pnc/rest/endpoint/ProductVersionEndpoint.java
@@ -134,7 +134,8 @@ public class ProductVersionEndpoint extends AbstractEndpoint<ProductVersion, Pro
     @Path("/{id}")
     public Response update(@ApiParam(value = "Product Version id", required = true) @PathParam("id") Integer id,
             ProductVersionRest productVersionRest) throws RestValidationException {
-        return super.update(id, productVersionRest);
+        productVersionProvider.update(id, productVersionRest);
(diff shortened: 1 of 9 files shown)


thescouser89:
retest this please

michalovjan:
retest this please

rest/src/main/java/org/jboss/pnc/rest/endpoint/ProductVersionEndpoint.java
```diff
@@ -134,7 +134,8 @@ public Response getSpecific(
     @Path("/{id}")
     public Response update(@ApiParam(value = "Product Version id", required = true) @PathParam("id") Integer id,
             ProductVersionRest productVersionRest) throws RestValidationException {
-        return super.update(id, productVersionRest);
+        productVersionProvider.update(id, productVersionRest);
```
jbartece: You should not update the endpoint, but rather create a validator and bind it to the Rest entity.
See RepositoryConfigurationEndpoint#update RepositoryConfigurationRest.internalUrl ScmUrl ScmUrlValidator

michalovjan:
retest this please

michalovjan:
After implementing the issue through BeanValidation i've found out that this is not high enough level to validate, because we need access to database which is not available on the bean level. -> leaving this on rest level

michalovjan:
retest this please

michalovjan:
retest this please

thescouser89:
retest this please

PR #2214, opened 2018-11-01
Add converted Endpoint APIs

janinko:


janinko:
It won't compile because of missing DTOs

PR #2237, opened 2018-11-20
[NCL-4235]: Fix build content ID is not unique

jbartece:
### Checklist:

### DIFF
--- a/bpm/src/main/java/org/jboss/pnc/bpm/task/BpmBuildTask.java
+++ b/bpm/src/main/java/org/jboss/pnc/bpm/task/BpmBuildTask.java
@@ -73,7 +73,7 @@ public class BpmBuildTask extends BpmTask {
     private BuildExecutionConfigurationRest getBuildExecutionConfiguration(BuildTask buildTask) {
 
         BuildConfigurationAudited buildConfigurationAudited = buildTask.getBuildConfigurationAudited();
-        String contentId = ContentIdentityManager.getBuildContentId(buildConfigurationAudited.getName());
+        String contentId = ContentIdentityManager.getBuildContentId(buildTask.getId());
 
         BuildExecutionConfiguration buildExecutionConfiguration = BuildExecutionConfiguration.build(
                 buildTask.getId(),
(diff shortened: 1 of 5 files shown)


bpm/src/main/java/org/jboss/pnc/bpm/task/BpmBuildTask.java
```diff
@@ -73,7 +73,7 @@ protected Serializable getProcessParameters() throws CoreException {
     private BuildExecutionConfigurationRest getBuildExecutionConfiguration(BuildTask buildTask) {
 
         BuildConfigurationAudited buildConfigurationAudited = buildTask.getBuildConfigurationAudited();
-        String contentId = ContentIdentityManager.getBuildContentId(buildConfigurationAudited.getName());
+        String contentId = ContentIdentityManager.getBuildContentId(buildConfigurationAudited);
```
matejonnet: Since ContentIdentityManager is about strings, I would rather pass in additional primitive parameter then a model object.

model/src/main/java/org/jboss/pnc/model/utils/ContentIdentityManager.java
```diff
@@ -45,9 +46,10 @@ public static String getProductContentId(ProductVersion productVersion) {
                 safeIdPart(productVersion.getVersion()));
     }
 
-    public static String getBuildContentId(String identifier) {
+    public static String getBuildContentId(BuildConfigurationAudited buildConfigurationAudited) {
         String timestamp = generateTimestamp();
-        return String.format(BUILD_CONTENT_ID_FORMAT, safeIdPart(identifier), timestamp);
+        return String.format(BUILD_CONTENT_ID_FORMAT, safeIdPart(buildConfigurationAudited.getName()), timestamp,
+                buildConfigurationAudited.getId(), buildConfigurationAudited.getRev());
```
pkocandr: I think adding ID into the generated string is not necessary since name is unique (in combination with active). This makes the generated build content id unnecessarily longer, not by much, but still...

bpm/src/main/java/org/jboss/pnc/bpm/task/BpmBuildTask.java
jbartece: I had the same thought, but later I changed my mind as it's in the model module, but I have no problem with changing it.

model/src/main/java/org/jboss/pnc/model/utils/ContentIdentityManager.java
jbartece: Right, I'll change it.

PR #2239, opened 2018-11-20
[NCL-3518] update demo data to contain valid SCM info

michalovjan:
### Checklist:

* [NCL-3518] implemented loading internal repositories from pnc-config files


michalovjan:
The idea is to have the valid internal SCMs to be loaded from pnc-config.json. These files are unique for every environment devel/stage/prod so we can specify the SCMs or leave them with default values (these are located in DemoDataConfig.java and used if the JsonProperties are not specified in config). 

This is the first part, the second one is to modify configs for each environment to use valid repos which are located in private repository.

WDYT? @jbartece @matejonnet

PR #2253, opened 2018-12-03
Add generic internal-repo pattern list, support for generic-http in isExternalOrigin()

jdcasey:
I'm adding handling logic for generic-http to deal with IllegalArgumentException when trying to sort out whether tracking record entries are from an external origin. I've also added better support for generic-http in the ignored path suffixes, to be consistent.

I couldn't find a place to add unit tests for these. Please advise if you need more WRT that.

### Checklist:

* [ ] Have you added a note in the CHANGELOG.md for your change if user-facing? (N/A)
* [ ] Have you added unit tests for your change? (No)

### DIFF
--- a/indy-repository-manager/src/main/java/org/jboss/pnc/indyrepositorymanager/IndyRepositorySession.java
+++ b/indy-repository-manager/src/main/java/org/jboss/pnc/indyrepositorymanager/IndyRepositorySession.java
@@ -669,7 +669,7 @@ public class IndyRepositorySession implements RepositorySession {
                 suffixes = ignoredPathSuffixes.getNpmWithShared();
                 break;
             case GENERIC_PKG_KEY:
-                suffixes = ignoredPathSuffixes.getGenericWithShared();
+                suffixes = ignoredPathSuffixes.getShared();
                 break;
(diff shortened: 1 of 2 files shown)


jdcasey:
@pkocandr I think this will help with the PNC error that's triggered by the SSL support we're trying to add to Indy.

indy-repository-manager/src/main/java/org/jboss/pnc/indyrepositorymanager/IndyRepositorySession.java
```diff
@@ -666,7 +669,7 @@ private boolean ignoreContent(String packageType, String path) {
                 suffixes = ignoredPathSuffixes.getNpmWithShared();
                 break;
             case GENERIC_PKG_KEY:
-                suffixes = ignoredPathSuffixes.getShared();
+                suffixes = ignoredPathSuffixes.getGenericWithShared();
```
pkocandr: I would stick to Shared only, because if we have a missing package type with some specific metadata path patterns we want to ignore, we shouldn't put them here, because it could cause issue with other unsupported package types.

PR #2265, opened 2018-12-10
Implement remaining DTOs

janinko:
The BuildResult DTO is empty ATM. It was too complex for me to solve it quickly.


janinko:
Don't take the integration test failures into account

PR #2287, opened 2019-01-04
Add @JsonDeserialize to DTOs

thescouser89:
To let Jackson know which Builder class to use to deserialize JSON back to
its DTO, we need to add `@JsonDeserialize` annotation to the class and
specify which Builder class to use

The existing `@JsonPOJOBuilder` annotation is used to tell Jackson how
to call the Builder class (which methods, setter method prefix to use)

Reference: https://stackoverflow.com/a/21624629/2907906

### Checklist:


thescouser89:
No need for changelog or unit tests (I think)

janinko:
Huh, how could I have forgot this?

PR #2289, opened 2019-01-08
Change visibility of DTO all args constructors to private

janinko:


janinko:
@matejonnet Are you OK with this after our discussion?

PR #2314, opened 2019-02-20
Add retries for build env creation

thescouser89:
This commit adds retries for build env creation. The number of retries
is configurable by setting 'creationPodRetry' in pnc-config.json

We do the retry by identifying when the pod creation failed by checking
its status periodically in monitors (or RunningTasks). The
monitors are setup in `monitorInitialization`. Once the monitors are
setup, they run in the background and `monitorInitialization` returns.

If the pod creation failed, we throw an exception that is ultimately
captured in `retryPod`. The latter will check the current number of
retries allowed. If 0: we call the `onError` consumer to end the
creation attempt. If greater than 0, we cancel and clear the existing
monitors, destroy any existing build environment Openshift objects,
recreate new build environment Openshift objects, and call
`monitorInitialization` again, with the number of retries decremented.

If the pod creation succeeded, we just call the `onComplete` consumer to
signal that pod creation is done.

This loop will continue until either the environment is created
successfully, or there is a failure and the number of retries is 0.

The side benefit of this commit is that instead of waiting for the pod to be
in state 'Running', we now identify if the pod failed and take action
based on this instead. This reduces the time we have to wait to take
alternate actions.

Note that in the case where pod creation fails with 'grpc connection
unavailable', we cannot capture that error on the pod status (it remains
on 'ContainerCreating' for some reason). However after \<timeout\> seconds
the creation will timeout, and another retry will be attempted.

### Checklist:

### DIFF
--- a/common/src/main/java/org/jboss/pnc/common/monitor/PullingMonitor.java
+++ b/common/src/main/java/org/jboss/pnc/common/monitor/PullingMonitor.java
@@ -20,6 +20,7 @@ package org.jboss.pnc.common.monitor;
 
 import org.jboss.pnc.common.concurrent.MDCExecutors;
 import org.jboss.pnc.common.util.ObjectWrapper;
+import org.jboss.pnc.common.util.ReadEnvProperty;
 import org.jboss.pnc.common.util.TimeUtils;
 import org.jboss.util.collection.ConcurrentSet;
 import org.slf4j.Logger;
@@ -30,7 +31,6 @@ import javax.enterprise.context.ApplicationScoped;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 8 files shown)


common/src/main/java/org/jboss/pnc/common/monitor/PullingMonitor.java
```diff
@@ -67,22 +67,20 @@
 
     public PullingMonitor() {
```
jbartece: I know, it is not part of this PR, but shouldn't it be called PollingMonitor?

openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
```diff
@@ -115,8 +139,12 @@ public OpenshiftStartedEnvironment(
             boolean tempBuild,
             Date temporaryBuildExpireDate) {
 
+        ReadEnvProperty reader = new ReadEnvProperty();
+        CREATION_POD_RETRY = reader.getIntValueFromPropertyOrDefault(CREATION_POD_RETRY_PROPERTY, DEFAULT_CREATION_POD_RETRY);
```
jbartece: This approach doesn't follow the standard way, how we configure orchestrator - using config file. We should avoid mixing the approaches and follow one.

common/src/main/java/org/jboss/pnc/common/util/ReadEnvProperty.java
```diff
@@ -0,0 +1,57 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.pnc.common.util;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ReadEnvProperty {
```
jbartece: I see that we read env properties in PullingMonitor, but IMHO it is wrong as it should use the config file as the rest of the application. Therefore I'd go with this deviation further and do not create a new class for it. We should refactor PullingMonitor instead and use config file approach and set default values in the config entity.

jbartece:
Some tests would be nice as well ;-).

common/src/main/java/org/jboss/pnc/common/util/ReadEnvProperty.java
matejonnet: I wonder what was the reason to not use the default configs in the PullingMonitor.

thescouser89: It was just so as to make it optional. And not change configs. But if you want it in pnc-config.json, we can do this.

common/src/main/java/org/jboss/pnc/common/monitor/PullingMonitor.java
thescouser89: This can be renamed in another PR :)

openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
thescouser89: Ok will use config file

thescouser89:
I have no idea how to test the changes in OpenshiftStartedEnvironment

thescouser89:
retest this please

thescouser89:
@jbartece @matejonnet This new commit incorporates changes regarding where to set creationPodRetry value in the config

PR #2315, opened 2019-02-21
[v1.4] Add retries for build env creation

thescouser89:
This commit adds retries for build env creation. The number of retries
is configurable by setting 'creationPodRetry' value in the
pnc-config.json

We do the retry by identifying when the pod creation failed by checking
its status periodically in monitors (or RunningTasks). The
monitors are setup in `monitorInitialization`. Once the monitors are
setup, they run in the background and `monitorInitialization` returns.

If the pod creation failed, we throw an exception that is ultimately
captured in `retryPod`. The latter will check the current number of
retries allowed. If 0: we call the `onError` consumer to end the
creation attempt. If greater than 0, we cancel and clear the existing
monitors, destroy any existing build environment Openshift objects,
recreate new build environment Openshift objects, and call
`monitorInitialization` again, with the number of retries decremented.

If the pod creation succeeded, we just call the `onComplete` consumer to
signal that pod creation is done.

This loop will continue until either the environment is created
successfully, or there is a failure and the number of retries is 0.

The side benefit of this commit is that instead of waiting for the pod to be
in state 'Running', we now identify if the pod failed and take action
based on this instead. This reduces the time we have to wait to take
alternate actions.

Note that in the case where pod creation fails with 'grpc connection
unavailable', we cannot capture that error on the pod status (it remains
on 'ContainerCreating' for some reason). However after <timeout> seconds
the creation will timeout, and another retry will be attempted.

Special thanks to @matejonnet for his review!

### Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - [NCL-4393] Adds pnc service status visualisations to the UI dashboard
 - [NCL-4392] Expose info about number of building and enqueued builds via REST
-
+- Add retries for build environment creation in case of pod creation failures
 
 ### Fixed
 - [NCL-4333] A closed milestone now cannot be set as current via REST API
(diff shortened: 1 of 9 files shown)


thescouser89:
@jbartece @matejonnet This new commit incorporates changes regarding where to set creationPodRetry value in the config

PR #2330, opened 2019-03-04
[NCL-4439]: Add a new generic parameter for extra Repos

jbartece:
…fy public repository required by the build

### Checklist:

### DIFF
--- a/maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
+++ b/maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
@@ -56,12 +56,14 @@ import org.slf4j.LoggerFactory;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import static org.commonjava.indy.pkg.maven.model.MavenPackageTypeDescriptor.MAVEN_PKG_KEY;
 import static org.jboss.pnc.mavenrepositorymanager.MavenRepositoryConstants.COMMON_BUILD_GROUP_CONSTITUENTS_GROUP;
@@ -78,6 +80,8 @@ import static org.jboss.pnc.mavenrepositorymanager.MavenRepositoryConstants.TEMP
  */
 @ApplicationScoped
 public class RepositoryManagerDriver implements RepositoryManager {
+
+    private static final String EXTRA_PUBLIC_REPOSITORIES_KEY = "EXTRA_PUBLIC_REPOSITORIES";
 
     private final Logger logger = LoggerFactory.getLogger(getClass());
 
@@ -188,14 +192,16 @@ public class RepositoryManagerDriver implements RepositoryManager {
      *                                    (or product, or shared-releases).
      */
     @Override
-    public RepositorySession createBuildRepository(BuildExecution buildExecution, String accessToken,
-            String serviceAccountToken) throws RepositoryManagerException {
+    public RepositorySession createBuildRepository(BuildExecution buildExecution,
+                                                   String accessToken,
+                                                   String serviceAccountToken,
+                                                   Map<String, String> genericParameters) throws RepositoryManagerException {
         Indy indy = init(accessToken);
         Indy serviceAccountIndy = init(serviceAccountToken);
-
         String buildId = buildExecution.getBuildContentId();
+
         try {
-            setupBuildRepos(buildExecution, serviceAccountIndy);
+            setupBuildRepos(buildExecution, serviceAccountIndy, genericParameters);
         } catch (IndyClientException e) {
             throw new RepositoryManagerException("Failed to setup repository or repository group for this build: %s", e,
                     e.getMessage());
@@ -236,7 +242,7 @@ public class RepositoryManagerDriver implements RepositoryManager {
      * @param execution The execution object, which contains the content id for creating the repo, and the build id.
      * @throws IndyClientException
      */
-    private void setupBuildRepos(BuildExecution execution, Indy indy)
+    private void setupBuildRepos(BuildExecution execution, Indy indy, Map<String, String> genericParameters)
             throws IndyClientException {
 
         String buildContentId = execution.getBuildContentId();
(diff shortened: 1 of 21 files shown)


jbartece:
I'm about to push some tests as well

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
```diff
@@ -270,14 +276,27 @@ private void setupBuildRepos(BuildExecution execution, Indy indy)
...
     }
 
+    private List<ArtifactRepository> extractExtraRepositoriesFromGenericParameters(Map<String, String> genericParameters) {
+        String extraReposString = genericParameters.get(EXTRA_PUBLIC_REPOSITORIES_KEY);
+        if (extraReposString == null)
+            return Collections.emptyList();
+
+        return Arrays.stream(extraReposString.split("\n"))
+                .map(repoString ->
+                        ArtifactRepository.build(repoString.trim(), repoString.trim(), repoString.trim(), true, false))
```
pkocandr: Could we introduce some format here including also an id? Something like
```central http://repo1.maven.org/maven2/```
If id not specified just convert the hostname by replacing dots for hyphens, e.g.
```repo1-maven-org```
The value can be used for both id and name to not complicate it any more.

Comment (jbartece) [reply maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java]: I'll go for the generation as you proposed. IMHO less input from users better UX

File: maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java [hunk 2]
```diff
@@ -270,14 +278,39 @@ private void setupBuildRepos(BuildExecution execution, Indy indy)
...
+
+        return Arrays.stream(extraReposString.split("\n"))
+                .map((repoString) -> {
+                    try {
+                        String id = new URL(repoString)
+                                .getHost()
+                                .replaceAll("\\.", "-");
+                        return ArtifactRepository.build(id, id, repoString.trim(), true, false);
+                    } catch (MalformedURLException e) {
+                        logger.info("Malformed repository URL entered: " + repoString + ". Skipping.");
```
matejonnet: For the master PR it would be useful to use the user log (warning) here, see
https://github.com/project-ncl/pnc/blob/master/indy-repository-manager/src/main/java/org/jboss/pnc/indyrepositorymanager/IndyRepositorySession.java#L91

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
```diff
@@ -270,14 +276,27 @@ private void setupBuildRepos(BuildExecution execution, Indy indy)
...
 
             indy.stores().create(buildGroup, "Creating repository group for resolving artifacts in build: " + id
                     + " (repo: " + buildContentId + ")", Group.class);
         }
     }
 
+    private List<ArtifactRepository> extractExtraRepositoriesFromGenericParameters(Map<String, String> genericParameters) {
+        String extraReposString = genericParameters.get(EXTRA_PUBLIC_REPOSITORIES_KEY);
+        if (extraReposString == null)
+            return Collections.emptyList();
```
pkocandr: Just noticed that this will cause issues on line 261 - EmptyList class does not support add and addAll methods, it throws UnsupportedOperationException when you try to use them. So you should return something like "new ArrayList<>()" instead.

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
```diff
@@ -79,6 +81,8 @@
 @ApplicationScoped
 public class RepositoryManagerDriver implements RepositoryManager {
 
+    private static final String EXTRA_PUBLIC_REPOSITORIES_KEY = "EXTRA_PUBLIC_REPOSITORIES";
```
pkocandr: Do we need to use the word PUBLIC here? I think the repos might be inside VPN also (typically a Brew repo), Indy is able to access them.

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
jbartece: You are commenting on an outdated version ;-). It was fixed in the next commits as it was discovered by failing test ;-)

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
jbartece: OK, I can drop the word public

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
```diff
@@ -270,14 +280,39 @@ private void setupBuildRepos(BuildExecution execution, Indy indy)
             // Global-level repos, for captured/shared artifacts and access to the outside world
             addGlobalConstituents(buildGroup, tempBuild);
 
-            // add extra repositories removed from poms by the adjust process
-            addExtraConstituents(execution.getArtifactRepositories(), id, buildContentId, indy, buildGroup);
+            // add extra repositories removed from poms by the adjust process and set in BC by user
+            List<ArtifactRepository> extraDependencyRepositories = extractExtraRepositoriesFromGenericParameters(genericParameters);
+            if(execution.getArtifactRepositories() != null) {
```
pkocandr: Missing space after "if"

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
```diff
@@ -270,14 +280,39 @@ private void setupBuildRepos(BuildExecution execution, Indy indy)
...
 
             indy.stores().create(buildGroup, "Creating repository group for resolving artifacts in build: " + id
                     + " (repo: " + buildContentId + ")", Group.class);
         }
     }
 
+    List<ArtifactRepository> extractExtraRepositoriesFromGenericParameters(Map<String, String> genericParameters) {
+        String extraReposString = genericParameters.get(EXTRA_PUBLIC_REPOSITORIES_KEY);
+        if (extraReposString == null)
+            return new ArrayList<>();
```
pkocandr: Missing curly brackets around this.

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
jbartece: Ah, I keep forgetting about it. I was used to do it without space. Updated :-)

maven-repository-manager/src/main/java/org/jboss/pnc/mavenrepositorymanager/RepositoryManagerDriver.java
jbartece: The same as above. I cannot get used to it. Updated :-)

PR #2335, opened 2019-03-11
NCL-3308 Add reminder to submit generic parameter

matedo1:


alexcreasy:
Hi Martin, I'm planning a much better fix to this, this only mitigates the UX problem here... there should be no need to submit the parameter at all. feel free to merge in the mean time., just FYI.

matedo1:
Hi Alex, thanks for letting me know and I agree some auto-saving concept should be implemented, hopefully this fix will be still useful for users until we are able to raise better solution

PR #2344, opened 2019-03-14
[NCL-4398] Implement BuildConfig/Project REST endpoints

thescouser89:
This commit does not contain the implementation of some endpoints for
triggering a build. This will be provided in a future commit.

Part of the reason for not implementing some endpoints is because it'll take time to get it ready to commit and during that time this might create more merge conflicts.

It'll also be nice for REST client users to be able to play with more available REST endpoints in the meantime.

### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationProviderImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationProviderImpl.java
@@ -19,26 +19,127 @@ package org.jboss.pnc.facade.providers;
 
 import org.jboss.pnc.dto.BuildConfiguration;
 import org.jboss.pnc.dto.BuildConfigurationRef;
+import org.jboss.pnc.dto.BuildConfigurationRevision;
+import org.jboss.pnc.dto.DTOEntity;
 import org.jboss.pnc.dto.response.Page;
+import org.jboss.pnc.dto.validation.groups.WhenCreatingNew;
+import org.jboss.pnc.dto.validation.groups.WhenUpdating;
 import org.jboss.pnc.facade.mapper.api.BuildConfigurationMapper;
+import org.jboss.pnc.facade.mapper.api.BuildConfigurationRevisionMapper;
 import org.jboss.pnc.facade.providers.api.BuildConfigurationProvider;
+import org.jboss.pnc.facade.validation.ConflictedEntryException;
+import org.jboss.pnc.facade.validation.ConflictedEntryValidator;
+import org.jboss.pnc.facade.validation.InvalidEntityException;
+import org.jboss.pnc.facade.validation.ValidationBuilder;
+import org.jboss.pnc.model.BuildConfigurationAudited;
+import org.jboss.pnc.model.GenericEntity;
+import org.jboss.pnc.model.IdRev;
+import org.jboss.pnc.spi.datastore.repositories.BuildConfigurationAuditedRepository;
 import org.jboss.pnc.spi.datastore.repositories.BuildConfigurationRepository;
+import org.jboss.pnc.spi.datastore.repositories.ProductVersionRepository;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.ejb.Stateless;
 import javax.inject.Inject;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
 
+import static org.jboss.pnc.spi.datastore.predicates.BuildConfigurationPredicates.isNotArchived;
+import static org.jboss.pnc.spi.datastore.predicates.BuildConfigurationPredicates.withDependantConfiguration;
+import static org.jboss.pnc.spi.datastore.predicates.BuildConfigurationPredicates.withName;
 import static org.jboss.pnc.spi.datastore.predicates.BuildConfigurationPredicates.withProductVersionId;
+import static org.jboss.pnc.spi.datastore.predicates.BuildConfigurationPredicates.withProjectId;
+import static org.jboss.pnc.common.util.StreamHelper.nullableStreamOf;
 
 @Stateless
 public class BuildConfigurationProviderImpl
         extends AbstractProvider<org.jboss.pnc.model.BuildConfiguration, BuildConfiguration, BuildConfigurationRef> implements BuildConfigurationProvider {
 
+    private final Logger logger = LoggerFactory.getLogger(BuildConfigurationProviderImpl.class);
+
+    private ProductVersionRepository productVersionRepository;
+    private BuildConfigurationRevisionMapper buildConfigurationRevisionMapper;
+    private BuildConfigurationAuditedRepository buildConfigurationAuditedRepository;
+
     @Inject
     public BuildConfigurationProviderImpl(BuildConfigurationRepository repository,
-                                          BuildConfigurationMapper mapper) {
+                                          BuildConfigurationMapper mapper,
+                                          ProductVersionRepository productVersionRepository,
+                                          BuildConfigurationRevisionMapper buildConfigRevisionMapper,
+                                          BuildConfigurationAuditedRepository buildConfigurationAuditedRepository) {
         super(repository, mapper, org.jboss.pnc.model.BuildConfiguration.class);
+
+        this.productVersionRepository = productVersionRepository;
+        this.buildConfigurationRevisionMapper = buildConfigRevisionMapper;
+        this.buildConfigurationAuditedRepository = buildConfigurationAuditedRepository;
     }
 
+
+    @Override
+    protected void validateBeforeSaving(BuildConfiguration buildConfigurationRest) {
+
+        super.validateBeforeSaving(buildConfigurationRest);
+
+        validateIfItsNotConflicted(buildConfigurationRest);
+    }
+
+    @Override
+    protected void validateBeforeUpdating(Integer id, BuildConfiguration buildConfigurationRest) {
+
+        super.validateBeforeUpdating(id, buildConfigurationRest);
+
+        validateIfItsNotConflicted(buildConfigurationRest);
+        validateDependencies(buildConfigurationRest.getId(), buildConfigurationRest.getDependencyIds());
+    }
+
+    private void validateDependencies(Integer buildConfigId, Set<Integer> dependenciesIds) throws InvalidEntityException {
+
+        if (dependenciesIds == null || dependenciesIds.isEmpty()) {
+            return;
+        }
+
+        org.jboss.pnc.model.BuildConfiguration buildConfig = repository.queryById(buildConfigId);
+
+        for (Integer dependencyId : dependenciesIds) {
+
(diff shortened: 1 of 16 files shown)


facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationProviderImpl.java
```diff
@@ -51,4 +160,165 @@ public BuildConfigurationProviderImpl(BuildConfigurationRepository repository,
...
+    @Override
+    public BuildConfigurationRevision getRevision(Integer id, Integer rev) {
+
+        IdRev idRev = new IdRev(id, rev);
+
+        BuildConfigurationAudited auditedBuildConfig = buildConfigurationAuditedRepository.queryById(idRev);
+
+        if (auditedBuildConfig == null) {
+            return null;
+        }
```
janinko: DTO mappers are safe for nulls

facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationProviderImpl.java
```diff
@@ -51,4 +160,165 @@ public BuildConfigurationProviderImpl(BuildConfigurationRepository repository,
...
+                equalsWithNull(audited.getBuildScript(), rest.getBuildScript()) &&
+                equalsId(audited.getRepositoryConfiguration(), rest.getRepository()) &&
+                equalsWithNull(audited.getScmRevision(), rest.getScmRevision()) &&
+                equalsWithNull(audited.getDescription(), rest.getDescription()) &&
+                equalsId(audited.getProject(), rest.getProject()) &&
+                equalsId(audited.getBuildEnvironment(), rest.getEnvironment()) &&
+                audited.getGenericParameters().equals(rest.getGenericParameters());
+    }
+
+    private boolean equalsWithNull(Object o1, Object o2) {
```
janinko: Objects#equals can be used instead.

facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
```diff
@@ -48,4 +67,65 @@ public BuildProviderImpl(BuildRecordRepository repository, BuildMapper mapper, B
...
+                                           int pageSize,
+                                           String sortingRsql,
+                                           String query,
+                                           Integer projectId) {
+
+        List<Object[]> buildConfigurationRevisions = AuditReaderFactory
+                .get(entityManager)
+                .createQuery()
+                .forRevisionsOfEntity(BuildConfiguration.class, false, false)
+                .add(AuditEntity.relatedId("project").eq(projectId))
```
janinko: Relationship to project is not audited so this whole thing is pointless and wastes resources. Instead just do normal query with `queryForCollection`.

facade/src/main/java/org/jboss/pnc/facade/providers/ProjectProviderImpl.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Inject
+    public ProjectProviderImpl(ProjectRepository repository, ProjectMapper mapper, ProjectRSQLMapper rsql) {
+        super(repository, mapper, rsql);
+    }
+
+
+    @Override
+    protected void validateBeforeSaving(Project projectRest) {
+
```
janinko: This should not call super?

model/src/main/java/org/jboss/pnc/model/BuildConfigurationAudited.java
```diff
@@ -33,7 +33,7 @@
  * Keep in mind that it is not managed by JPA and needs to be filled manually.
  *
  */
-public class BuildConfigurationAudited {
+public class BuildConfigurationAudited implements GenericEntity<Integer> {
```
janinko: Not sure if we can do this. What do you think @matejonnet @jbartece ?

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/BuildConfigurationEndpointImpl.java
```diff
@@ -0,0 +1,198 @@
...
+        this.buildConfigurationProvider = buildConfigurationProvider;
+        this.buildProvider = buildProvider;
+        this.bcSupportedGenericParametersProvider = bcSupportedGenericParametersProvider;
+        this.groupConfigurationProvider = groupConfigurationProvider;
+    }
+
+    @Override
+    public Page<BuildConfiguration> getAll(PageParameters pageParams) {
+        return super.getAll(pageParams);
+    }
```
janinko: I wonder, would it work if we skip this implementation entirely? And the following that just call their super version too?

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/BuildConfigurationEndpointImpl.java
```diff
@@ -0,0 +1,198 @@
...
+
+    @Override
+    public BuildConfigurationRevision createRevision(int id, BuildConfiguration buildConfiguration) {
+        buildConfigurationProvider.update(id, buildConfiguration);
+
+        Optional<BuildConfigurationRevision> buildConfigurationAuditedRestOptional =
+                buildConfigurationProvider.getLatestAuditedMatchingBCRest(buildConfiguration);
+
+        if (buildConfigurationAuditedRestOptional.isPresent()) {
+            return buildConfigurationAuditedRestOptional.get();
```
janinko: You can use `.orElseThrow` on optionals.
Maybe using IllegalStateException would make more sense?
And `buildConfigurationAuditedRestOptional` matches the old type name :P

facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationProviderImpl.java
```diff
@@ -51,4 +160,165 @@ public BuildConfigurationProviderImpl(BuildConfigurationRepository repository,
...
+
+        org.jboss.pnc.model.BuildConfiguration buildConfig = repository.queryById(configId);
+
+        ValidationBuilder.validateObject(buildConfig, WhenUpdating.class)
+                .validateCondition(buildConfig!=null, "No build config exists with id: " + configId);
+
+        ProductVersion productVersion = null;
+
+        if (productVersionId != null) {
+            productVersion = productVersionRepository.queryById(productVersionId);
```
janinko: What happens if the product version with such ID doesn't exists? Shouldn't there be a validation?

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/BuildConfigurationEndpointImpl.java
thescouser89: This is doable if we use the `public` access for all the methods in `AbstractEndpoint`. Do you think if that's ok? I don't know the original reason why it was set to `protected`.

thescouser89:
Thanks for all the comments. Will go through the rest on Wednesday and hopefully I'll get something ready for review again :)

facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationProviderImpl.java
thescouser89: Will add!

facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
thescouser89: Replaced with:

```
        @SuppressWarnings("unchecked")
        Set<Integer> buildConfigIds = buildConfigurationRepository
                .queryWithPredicates(withProjectId(projectId))
                .stream()
                .map(BuildConfiguration::getId)
                .collect(Collectors.toSet());

        return queryForCollection(pageIndex, pageSize, sortingRsql, query, withBuildConfigurationIds(buildConfigIds));
```

thescouser89: BuildRecord has no projectId field, but has a relationship to project via it's associated BuildConfiguration

facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationProviderImpl.java
```diff
@@ -51,4 +160,165 @@ public BuildConfigurationProviderImpl(BuildConfigurationRepository repository,
...
+    public Page<BuildConfigurationRevision> getRevisions(int pageIndex, int pageSize, Integer id) {
+
+        List<BuildConfigurationAudited> auditedBuildConfigs =
+                buildConfigurationAuditedRepository.findAllByIdOrderByRevDesc(id);
+
+        List<BuildConfigurationRevision> toReturn =  nullableStreamOf(auditedBuildConfigs)
+                .map(buildConfigurationRevisionMapper::toDTO)
+                .skip(pageIndex * pageSize)
+                .limit(pageSize)
+                .collect(Collectors.toList());
```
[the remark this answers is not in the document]
thescouser89: While we can do the pagination via the `AuditQuery` (setFirstResult, setMaxResults), I think it's not worth it because when creating the Page object, we need to provide the 'totalHits'. Using AuditQuery, there's no way to tell it that we only want a count of all the audited objects. Instead we can only get the full result list.

thescouser89:
@janinko  Pushed resolved comments + fix merge conflict

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/BuildConfigurationEndpointImpl.java
matejonnet: The reason is probably that some endpoints don't have getAll. Eg. UserEndpoint.

facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationProviderImpl.java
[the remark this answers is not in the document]
janinko: ok, let's keep it that way and if we decide we don't need total hits/pages we can delete it then

facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationProviderImpl.java
```diff
@@ -51,4 +153,154 @@ public BuildConfigurationProviderImpl(BuildConfigurationRepository repository,
...
+        ValidationBuilder.validateObject(buildConfig, WhenUpdating.class)
+                .validateCondition(buildConfig != null, "No build config exists with id: " + configId)
+                .validateCondition(dependency != null, "No dependency build config exists with id: " + dependencyId);
+
+        buildConfig.removeDependency(dependency);
+        repository.save(buildConfig);
+    }
+
+    @Override
+    public void setProductVersion(Integer configId, Integer productVersionId) {
```
janinko: This method is not used at all! Please don't just copy all the code from previous producers but take this oportunity to do a litlle cleanup.

thescouser89:
retest this please

PR #2352, opened 2019-03-21
Replace Instant objects with Date objects

thescouser89:
The Instant objects are somewhat challenging to serialize via Jackson
and it would be better to use the regular Date objects for serialization
and deserialization

### Checklist:


matejonnet:
As discussed we'll keep the Instant.

PR #2361, opened 2019-04-01
Some fixes and improvements for RR

janinko:


rest/src/main/java/org/jboss/pnc/rest/notifications/websockets/TypedMessage.java
```diff
@@ -30,7 +30,6 @@
  * @author <a href="mailto:matejonnet@gmail.com">Matej Lazar</a>
  */
 @JsonDeserialize(builder = TypedMessage.TypedMessageBuilder.class)
-@Builder
```
janinko: The class already has builder written by hand.

PR #2362, opened 2019-04-01
Add Group* endpoints

janinko:


janinko:
retest this please

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/GroupBuildEndpointImpl.java
```diff
@@ -0,0 +1,114 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.ws.rs.NotFoundException;
+
+/**
+ *
+ * @author Honza Brázdil &lt;jbrazdil@redhat.com&gt;
+ */
+@ApplicationScoped
```
thescouser89: `@Stateless` maybe?

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/GroupConfigurationEndpointImpl.java
```diff
@@ -0,0 +1,140 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.enterprise.context.ApplicationScoped;
+
+/**
+ *
+ * @author Honza Brázdil &lt;jbrazdil@redhat.com&gt;
+ */
+@ApplicationScoped
```
thescouser89: @Stateless maybe?

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/BuildEndpointImpl.java
```diff
@@ -0,0 +1,152 @@
...
+import org.jboss.pnc.rest.api.parameters.BuildsFilterParameters;
+import org.jboss.pnc.rest.api.parameters.PageParameters;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ *
+ * @author Honza Brázdil &lt;jbrazdil@redhat.com&gt;
+ */
+public class BuildEndpointImpl extends AbstractEndpoint<Build, BuildRef> implements BuildEndpoint {
```
thescouser89: Missing EJB annotation!

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/GroupBuildEndpointImpl.java
janinko: Why? It doesn't have to be in transactions.

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/BuildEndpointImpl.java
janinko: Yeah, missing annotation but why EJB and not CDI?

Also from [Resteasy doc](https://docs.jboss.org/resteasy/docs/3.6.1.Final/userguide/html/CDI.html):
> If a JAX-RS root resource does not define a scope explicitly, it is bound to the Request scope.

thescouser89: Sure go with CDI then!

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/GroupBuildEndpointImpl.java
thescouser89: Wouldn't @ApplicationScoped create a singleton basically?

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/BuildEndpointImpl.java
thescouser89: We'll have to sync all the other EndpointImpl to use the same CDI annotation/EJB though

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/GroupBuildEndpointImpl.java
janinko: Yes it would. That's why I'm using it.

thescouser89: Does that affect the throughput when requests go in? If I remember correctly, its 1 thread per request? But yeah I dunno how this behaves with CDI and no concurrency limit.

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/BuildEndpointImpl.java
janinko: Yeah, I have it planned

PR #2376, opened 2019-04-08
[NCL-4538] Add new BUILDER_POD_MEMORY parameter

janinko:
### Checklist:


janinko:
Waiting for 1.5 branch for this PR

spi/src/main/java/org/jboss/pnc/spi/environment/EnvironmentDriver.java
```diff
@@ -48,7 +51,8 @@ StartedEnvironment startEnvironment(
             SystemImageType systemImageType,
             RepositorySession repositorySession,
             DebugData debugData,
-            String accessToken) throws EnvironmentDriverException;
+            String accessToken,
+            Map<String, String> parameters) throws EnvironmentDriverException;
```
matejonnet: It would be better to use specific parameter instead of the map.

janinko: I understand (generic) parameters to be a map that is passed to various services and they rocognize their parameters and use them: CUSTOM_PME_PARAMETERS is consumed by PME, EXECUTION_ROOT_NAME is consumed by Repour and BUILDER_POD_MEMORY would be consumed by environment drivers.

You think it should be some other service/part that would consume the BUILDER_POD_MEMORY parameter?

matejonnet: The API that defines what is required rather than taking the map of Strings is more readable.
The other thing is OpenshiftStartedEnvironment it is an implementation details that does not need to know about the Keys in the generic parametrs map.
I see the use case you described more on the REST api level. On the other side I see that repository driver has been done in the same way.

jbartece:
@janinko Don't wait for branch 1.5. There are also commits in the 1.4, which will be released as 1.5. Once the PR is ready, merge it. We will sort out the branches when releasing.

jbartece:
Sorry, missclick

PR #2390, opened 2019-05-01
NCL-4435

alexcreasy:
### Checklist:

### DIFF
--- a/ui/app/artifacts/artifacts.module.js
+++ b/ui/app/artifacts/artifacts.module.js
@@ -1,0 +1,61 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014-2019 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


alexcreasy:
@thescouser89 this wasn't ready to go, it's broken :( I should have put some note on the PR that I was holding off as everything was ticked and it looked ready to go. I had to context switch onto sprint tasks and was coming back to this after they were done.

This probably is a good point we need to figure out, what do we do with PRs that are opened for people to discuss and collaborate on but aren't quite ready to go? We don't want them piling up, but likewise, if they are going to run a long time and aren't ready to merge, do we close them? Do we tag them as not ready or something? 

Ultimately the lack of adequate tests is the real culprit here I suppose, but beyond that, is there some process we can adopt?

Let's not revert this as it breaks nothing else, I'll try and come up with a minimum fix today.

matedo1:
I think there could be two solutions, use title prefix like [don't merge] or there is draft pull request feature https://github.blog/2019-02-14-introducing-draft-pull-requests/

thescouser89:
Draft pull request works!

alexcreasy:
I think draft pull requests seem like exactly what we need 👍

PR #2402, opened 2019-05-07
[NCL-4624] [RR] product milestone version validation endpoint

michalovjan:
### Checklist:

* [ ] Update documentation.

### DIFF
--- a/dto/src/main/java/org/jboss/pnc/dto/requests/validation/VersionValidationRequest.java
+++ b/dto/src/main/java/org/jboss/pnc/dto/requests/validation/VersionValidationRequest.java
@@ -1,0 +1,42 @@
+/**
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014-2019 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.pnc.dto.requests.validation;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import lombok.Builder;
+import lombok.Data;
+
+import javax.validation.constraints.NotBlank;
+
+/**
+ *
+ * @author jmichalo <jmichalo@redhat.com>
+ */
+@Data
+@Builder(builderClassName = "Builder")
+@JsonDeserialize(builder = VersionValidationRequest.Builder.class)
+public class VersionValidationRequest {
+
+    @NotBlank
+    public final String version;
+
+    @JsonPOJOBuilder(withPrefix = "")
+    public static final class Builder {
+    }
+}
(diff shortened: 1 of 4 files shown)


dto/src/main/java/org/jboss/pnc/dto/requests/validation/VersionValidationRequest.java
```diff
@@ -0,0 +1,21 @@
...
+
+import javax.validation.constraints.NotBlank;
+
+@Data
+@Builder(builderClassName = "Builder")
+@JsonDeserialize(builder = VersionValidationRequest.Builder.class)
+public class VersionValidationRequest {
+
+    @NotBlank
+    public final String version;
```
matedo1: I think the final decision was to use "value" attribute?

michalovjan: I've discussed with @janinko whether to have a RequestDTO that is done in a general way and takes f.e. one string. It was decided that it could cause confusion because some DTO's could take a List of strings, number or List of numbers. For each, there would have to be a separate class with names like StringRequest, StringListRequest... and in the code, it could potentially affect readability. To be sure we went with concrete naming and attributes.

janinko: As it is `/product-milestones/validate-version` it make sense to me that the field is named `version`. What's your reasonto nameit `value`, @matedo1?

matedo1: my first proposal was to name it `version`, we considered making it more generic while using `value` (as it could be reused for other string validations too), but as API consumer I totally agree also with `version` that is actually slightly better and easier to read, so if it's ok for you guys, let's use it

janinko:
@michalovjan @matedo1 should the `ValidationResponse` class have only `isValid` field, or also (optional?) message?

michalovjan:
I am all for it! Message sound good.

matedo1:
It's actually good idea, one step forward could have to make it similar with Field validation format, moving our discussion to chat.

dto/src/main/java/org/jboss/pnc/dto/response/ValidationResponse.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @author jmichalo <jmichalo@redhat.com>
+ */
+@Data
+@Builder(builderClassName = "Builder")
+@JsonDeserialize(builder = ValidationResponse.Builder.class)
+public class ValidationResponse {
+
+    @NotNull
+    public final Boolean isValid;
```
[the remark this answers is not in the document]
michalovjan: I remember a discussion with you why we shouldn't use primitive types in DTOs. Is this different case?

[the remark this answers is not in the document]
janinko: I don't remember the discussion, but you are right it could theoreticly cause problems in requests. But as this is only response object, I think it should be safe.

michalovjan:
As per @matedo1 's suggestion, I've added a hint for ProductMilestone version.

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/ProductMilestoneEndpointImpl.java
```diff
@@ -96,4 +102,16 @@ public void closeMilestone(String id, ProductMilestone productMilestone) {
...
         productMilestoneProvider.cancelMilestoneCloseProcess(id);
     }
+
+    @Override
+    public ValidationResponse validateVersion(VersionValidationRequest versionRequest) {
+        return ValidationResponse.builder()
+                .isValid(Pattern.matches(Patterns.PRODUCT_MILESTONE_VERSION, versionRequest.getVersion()))
+                .hints(
+                        Lists.newArrayList(
+                                "Product Milestone version consists of 2 or 3 numeric components followed by a string"
```
matedo1: I would try to make it a little bit shorter, something like:
`Allowed format consists of 2 or 3 numeric components (separated by a dot) followed by a string qualifier starting with a character, eg. 3.0.0.GA, 1.0.11.CR2.ER1, 3.0.CR2`

michalovjan: Will change!

PR #2411, opened 2019-05-13
[v1.6][NCL-4568] Stop env monitor on error

thescouser89:
Sometimes object pod creation fails because of quota limits. In those
cases, an exception is thrown on pod creation indicating we are above
quota. Unfortunately we do not capture the exception and deal with it
properly.

This commit attempts to capture the exception and calls the 'onError'
callback when an exception is thrown. This is done by:

- Using CompletableFuture instead of Future

  This is done to:
  * be able to capture any exceptions raised in the runnables
  * combine CompletableFutures together

- Adding an additional monitor to check if the Openshift
  object creation CompletableFutures are finished and without
  exceptions. If with exceptions, it is wrapped as a
  PodFailedStartedException and thrown (and caught by the
  PollingMonitor).

  If the CompletableFutures are not finished yet, then we just return
  false, to indicate that it's not ready yet.

- Add all the monitors into the monitors list to be able to cancel them
  on failure

### Checklist:

### DIFF
--- a/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
+++ b/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
@@ -53,16 +53,19 @@ import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


thescouser89:
retest this please

thescouser89:
If I understand correctly, `startMonitor` will add the periodic monitors to check if the service, route, and pod are running?

I think we could use the 'super' CompleteableFuture (that groups all the creation completable future) and when complete, it calls that `startMonitor` above. Then yeah we could remove those booleans :)

thescouser89:
closing this as it didn't make it to 1.6

PR #2430, opened 2019-05-17
[v1.7][NCL-4568] Stop env monitor on error

thescouser89:
Sometimes object pod creation fails because of quota limits. In those
cases, an exception is thrown on pod creation indicating we are above
quota. Unfortunately we do not capture the exception and deal with it
properly.

This commit attempts to capture the exception and calls the 'onError'
callback when an exception is thrown. This is done by:

- Using CompletableFuture instead of Future

  This is done to:
  * be able to capture any exceptions raised in the runnables
  * combine CompletableFutures together

- Adding an additional monitor to check if the Openshift
  object creation CompletableFutures are finished and without
  exceptions. If with exceptions, it is wrapped as a
  PodFailedStartedException and thrown (and caught by the
  PollingMonitor).

  If the CompletableFutures are not finished yet, then we just return
  false, to indicate that it's not ready yet.

- Add all the monitors into the monitors list to be able to cancel them
  on failure

### Checklist:

### DIFF
--- a/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
+++ b/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
@@ -106,10 +106,6 @@ public class OpenshiftStartedEnvironment implements StartedEnvironment {
      */
     private static final String BUILDER_POD_MEMORY = "BUILDER_POD_MEMORY";
 
-    private boolean serviceCreated = false;
-    private boolean podCreated = false;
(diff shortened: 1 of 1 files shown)


thescouser89:
@matejonnet  You asked on https://github.com/project-ncl/pnc/pull/2411 : 

> Is it still required to have podCreated (serviceCreated, routeCreated) flags?
> What about starting a monitor once the object(pod) is created?
> eg.: creatingPod.thenApplyAsync(startMonitor)

I can't do the example above because at that point (new OpenshiftStartedEnvironment) I don't have the callbacks for the monitors. They are provided on 'monitorInitization' method call.

PR #2442, opened 2019-05-30
[NCL-4701] Store scm info on build failure if possible

thescouser89:
This commit adds the scm information of a build in case of failure. This is useful when a build has already finished alignment, and is either in the process of starting the build or building, and a failure occurs.

ncl-4701-before
From the picture above, you can see that the scm revision and scm tag are not stored. This is useful information that the builder might want to see to debug her builds locally (to see for e.g the alignment done)

With this commit, the scm revision and scm tag are stored (if available).
ncl-4701-after

### Checklist:


thescouser89:
retest this please

PR #2450, opened 2019-06-05
[NCL-4796] Implement getSpecific for builds

thescouser89:
### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/mapper/api/BuildMapper.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/mapper/api/BuildMapper.java
@@ -46,7 +46,7 @@ import java.util.stream.Collectors;
 public interface BuildMapper extends EntityMapper<BuildRecord, Build, BuildRef>{
 
     @Override
-    @Mapping(target = "environment", source = "buildEnvironment", qualifiedBy = Reference.class)
+    @Mapping(target = "environment", source = "buildConfigurationAudited.buildEnvironment", qualifiedBy = Reference.class)
     @Mapping(target = "dependentBuildIds", source = "dependentBuildRecordIds")
     @Mapping(target = "dependencyBuildIds", source = "dependencyBuildRecordIds")
     @Mapping(target = "buildConfigurationRevision", source = "buildConfigurationAudited", resultType = BuildConfigurationRevisionRef.class)
@@ -57,7 +57,7 @@ public interface BuildMapper extends EntityMapper<BuildRecord, Build, BuildRef>{
             "buildLogSize", "sshCommand", "sshPassword", "executionRootName", "executionRootVersion", "builtArtifacts",
             "dependencies", "productMilestone", "buildConfigSetRecord", "repourLog", "repourLogMd5", "repourLogSha256",
             "repourLogSize", "buildRecordPushResults", "buildConfigurationId", "buildConfigurationRev",
-            "buildConfigurationAuditedIdRev"
+            "buildConfigurationAuditedIdRev", "buildEnvironment"
     })
     Build toDTO(BuildRecord dbEntity);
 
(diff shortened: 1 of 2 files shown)


facade/src/main/java/org/jboss/pnc/facade/mapper/api/BuildMapper.java
```diff
@@ -101,6 +109,21 @@ default BuildRecord toIDEntity(BuildRef dtoEntity) {
     @BeanMapping(ignoreUnmappedSourceProperties = {"project", "repository"})
     BuildRecord toEntity(Build dtoEntity);
 
+
+
+    @Mapping(target = "project", source = "buildConfigurationAudited.project", resultType = ProjectRef.class)
+    @Mapping(target = "repository", source = "buildConfigurationAudited.repositoryConfiguration")
```
michalovjan:
```suggestion
    @Mapping(target = "repository", source = "buildConfigurationAudited.repositoryConfiguration", qualifiedBy = Reference.class)
```
without Reference.class mapstruct uses toDTO instead of toRef method for mapping the SCMRepository, it can cause confusion and it's good practice to keep the mappers clean :)

facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
```diff
@@ -67,20 +69,27 @@
     private BuildConfigurationAuditedRepository buildConfigurationAuditedRepository;
     private Gerrit gerrit;
     private BuildConfigurationRevisionMapper buildConfigurationRevisionMapper;
+    private BuildMapper buildMapper;
```
janinko: The mapper is already included in the superclass. I think it's named `mapper`.

facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
```diff
@@ -67,20 +69,27 @@
...
 
     @Inject
     public BuildProviderImpl(BuildRecordRepository repository,
                              BuildMapper mapper,
                              BuildConfigurationRepository buildConfigurationRepository,
                              BuildConfigurationAuditedRepository buildConfigurationAuditedRepository,
                              Gerrit gerrit,
-                             BuildConfigurationRevisionMapper buildConfigurationRevisionMapper) {
+                             BuildConfigurationRevisionMapper buildConfigurationRevisionMapper,
+                             BuildMapper buildMapper,
```
janinko: `BuildMapper` is here twice.

facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
```diff
@@ -236,6 +245,27 @@ private BuildRecord getBuildRecord(int id) {
...
+        } else {
+
+            List<BuildTask> runningBuilds = buildCoordinator.getSubmittedBuildTasks();
+
+            return runningBuilds.stream()
+                    .filter(buildTask -> id.equals(buildTask.getId()))
+                    .findAny()
+                    .map(buildMapper::fromBuildTask)
+                    .orElse(null);
+        }
```
janinko: Can it happen that build is both running and in DB? I would think that trying to find a build among running builds would be faster than in DB, so I would swap the order.

matejonnet: It cannot happen that the running build is also in the DB. The Build record in DB is created when the build completes.

facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
thescouser89: I remove the second argument. Good catch! :)

facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
thescouser89: I cannot use the one in the superclass since it's an EntityMapper type, with `toDTO` and other entity + DTO conversions.

facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
thescouser89: I've changed the order anyway. But yeah as Matej said, it can only be present in the running or in the DB, but not both.

PR #2461, opened 2019-06-20
[NCL-4797]: Add implementation of an endpoint /builds/cancel

jbartece:
### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/BuildTriggerer.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/BuildTriggerer.java
@@ -20,6 +20,7 @@ package org.jboss.pnc.facade;
 import java.util.Optional;
 import java.util.OptionalInt;
 
+import org.jboss.pnc.common.logging.BuildTaskContext;
 import org.jboss.pnc.dto.requests.GroupBuildRequest;
 import org.jboss.pnc.spi.BuildOptions;
 import org.jboss.pnc.spi.exception.BuildConflictException;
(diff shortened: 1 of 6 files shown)


facade/src/main/java/org/jboss/pnc/facade/BuildTriggerer.java
```diff
@@ -36,5 +38,16 @@ int triggerBuild(int buildConfigId, OptionalInt rev, BuildOptions buildOptions)
 
     int triggerGroupBuild(int groupConfigId, Optional<GroupBuildRequest> revs, BuildOptions buildOptions)
             throws BuildConflictException, CoreException;
-    
+
+    /**
+     * Cancels a running build
+     *
+     * @param buildId ID of a running build
+     * @return True if the cancel request is successfully accepted, otherwise false.
```
janinko: `false` means that the build doesn't exist or can it mean something else too?

jbartece: It only means, that it doesn't exist. Correcting the javadoc.

facade/src/main/java/org/jboss/pnc/facade/impl/BuildTriggererImpl.java
```diff
@@ -108,6 +110,16 @@ public int triggerGroupBuild(int groupConfigId,
         return result.getId();
     }
 
+    @Override
+    public boolean cancelBuild(int buildId) throws CoreException {
+        return buildCoordinator.cancel(buildId);
+    }
+
+    @Override
+    public Optional<BuildTaskContext> getMdcMeta(Integer buildId) {
```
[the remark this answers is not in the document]
jbartece: OK

PR #2465, opened 2019-06-20
[NCL-3800] Unification of status work

thescouser89:
### Checklist:


constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
```diff
@@ -55,51 +55,58 @@
...
-    UNKNOWN,
+    NEW;
 
-    /**
-     * There have not been any builds of this configuration, essentially the same meaning
-     * as a null status
-     */
-    NONE;
+    // isFinal means it's in the final state, it could also have failed
+    private final boolean isFinal;
```
jbartece: Why do we need this boolean?

thescouser89: To know if the build is in a final state. Aka if it's done (whether successful or not)

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
```diff
@@ -20,21 +20,21 @@
...
- * Status of a running or completed build.
+ * Status of a running or isFinal build.
  *
  * @author Jakub Bartecek &lt;jbartece@redhat.com&gt;
  *
  */
 public enum BuildStatus {
     /**
-     * Build completed successfully
+     * Build isFinal successfully
```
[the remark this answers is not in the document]
thescouser89: That's intellij being too aggressive. Will fix it. Thanks for reporting it!

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
jbartece: Do we have any use-case when we need it?

thescouser89: Yes: https://github.com/project-ncl/pnc/blob/master/build-coordinator/src/main/java/org/jboss/pnc/coordinator/notifications/buildTask/BuildStatusNotifications.java#L66 (this is using BuildCoordinatorStatus, so we need an equivalent for BuildStatus)

thescouser89:
@jbartece Unfortunately I won't be able to fix the test failures before the end of the sprint. Sorry about that :(

thescouser89:
retest this please

thescouser89:
retest this please

PR #2468, opened 2019-06-21
Add Json Patch support.

matejonnet:


rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/BuildConfigurationEndpoint.java
```diff
@@ -123,6 +122,7 @@
     })
     @GET
     @Path("/{id}")
+    @Consumes(MediaType.APPLICATION_JSON_PATCH_JSON) //workaround for PATCH support
```
janinko: What is this doing? And why there is `@Consumes` on a getter method?

rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ProductVersionEndpoint.java
```diff
@@ -114,6 +114,23 @@ void update(
             @Parameter(description = PV_ID) @PathParam("id") int id,
             @NotNull ProductVersion productVersion);
 
+    @Operation(summary = "Patch an existing product version.",
+            responses = {
+                    @ApiResponse(responseCode = SUCCESS_CODE, description = SUCCESS_DESCRIPTION,
```
janinko: Should we use here (and on other patch specific) this?
```suggestion
                    @ApiResponse(responseCode = ENTITY_UPDATED_CODE, description = ENTITY_UPDATED_DESCRIPTION,
```

rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ProductVersionEndpoint.java
```diff
@@ -114,6 +114,23 @@ void update(
...
+                            content = @Content(schema = @Schema(implementation = ProductVersion.class))),
+                    @ApiResponse(responseCode = INVALID_CODE, description = INVALID_DESCRIPTION,
+                            content = @Content(schema = @Schema(implementation = ErrorResponse.class))),
+                    @ApiResponse(responseCode = NOT_FOUND_CODE, description = NOT_FOUND_DESCRIPTION),
```
janinko: We use INVALID_CODE instead of NOT_FOUND for methods that take body as a parameter (like update).
Also you should add description to the `/Newcastle+Rest+API+Guidelines#NewcastleRestAPIGuidelines-CommonCRUDentityoperations` confluence doc.
And there should also probably be the CONFLICT code response.

dto/src/main/java/org/jboss/pnc/dto/BuildConfiguration.java
```diff
@@ -74,6 +84,6 @@ private BuildConfiguration(SCMRepository repository, ProjectRef project, Environ
     }
 
     @JsonPOJOBuilder(withPrefix = "")
-    public static final class Builder {
+    public static class Builder {
```
janinko: Why have you removed the final from the Builder?

rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/BuildConfigurationEndpoint.java
matejonnet: RestEasy creates a full entity from the JsonPatch request, it does it by internally invoking the GET method mapped on the same Path. The additional media type is required to workaround a RestEasy bug.

dto/src/main/java/org/jboss/pnc/dto/BuildConfiguration.java
matejonnet: Probably by mistake, adding it back...

rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ProductVersionEndpoint.java
matejonnet: According to the [guidelines](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH), if there is a response body, the code should be 200.

rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ProductVersionEndpoint.java
matejonnet: 404 NOT_FOUND is returned by the GET method that is invoked before the PATCH one. To change the behaviour it requires to create a custom PatchMethodFilter which unfortunately cannot be overridden in current RestEasy version but requires to patch EAP distribution. I would leave it as is.

PR #2472, opened 2019-06-25
[NCL-4801]: Use flushAndRefresh method to resolve issues with storing artifacts

jbartece:
### Checklist:

### DIFF
--- a/datastore/src/main/java/org/jboss/pnc/datastore/repositories/internal/AbstractRepository.java
+++ b/datastore/src/main/java/org/jboss/pnc/datastore/repositories/internal/AbstractRepository.java
@@ -25,14 +25,21 @@ import org.jboss.pnc.spi.datastore.repositories.api.SortInfo;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
 
+import javax.inject.Inject;
+import javax.persistence.EntityManager;
 import java.io.Serializable;
 import java.util.Collection;
 import java.util.List;
 
 public class AbstractRepository<T extends GenericEntity<ID>, ID extends Serializable> implements Repository<T, ID> {
 
+    @Inject
+    protected EntityManager entityManager;
+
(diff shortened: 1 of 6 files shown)


jbartece:
retest this please

janinko:
Why haven't you create the flushAndRefresh method in the Repository? I think it would make more sense there.

jbartece:
@janinko I don't follow. It is in the repository. See
https://github.com/project-ncl/pnc/pull/2472/files#diff-091bbe31bae658a95985f3d59cc72c6aL49
https://github.com/project-ncl/pnc/pull/2472/files#diff-92da85adf5c0bee7c108138d96d3aee9R27

janinko:
sorry, brainfart

datastore/src/main/java/org/jboss/pnc/datastore/repositories/internal/AbstractRepository.java
```diff
@@ -33,20 +34,30 @@
 
     protected JpaRepository<T, ID> springRepository;
     protected JpaSpecificationExecutor<T> springSpecificationsExecutor;
+    protected EntityManager entityManager;
```
janinko: Why not use just this instead of passing it through constructor?
```suggestion
    @Inject
    protected EntityManager entityManager;
```

jbartece: Can it be used in the abstract class? I was not sure about it. I thought it is not possible. If it is possible, then yes, it is much better.

janinko: it is possible. I use it like this in abstratProvider I think

jbartece: Done

jbartece:
@janinko edited based on your comment. Yes, it simplified it a lot. I force updated it to make it more readable.

PR #2473, opened 2019-06-25
[NCL-4771] implement build list returning endpoints

janinko:
### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
@@ -19,6 +19,8 @@ package org.jboss.pnc.facade.providers;
 
 import org.jboss.pnc.common.gerrit.Gerrit;
 import org.jboss.pnc.common.gerrit.GerritException;
+import static org.jboss.pnc.common.util.StreamHelper.nullableStreamOf;
+import org.jboss.pnc.common.util.StringUtils;
 import org.jboss.pnc.dto.Build;
 import org.jboss.pnc.dto.BuildConfigurationRevision;
 import org.jboss.pnc.dto.BuildRef;
@@ -48,6 +50,7 @@ import javax.annotation.security.PermitAll;
 import javax.annotation.security.RolesAllowed;
 import javax.ejb.Stateless;
 import javax.inject.Inject;
+
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.List;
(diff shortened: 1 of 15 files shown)


rest-api/src/main/resources/openapi.yml
```diff
@@ -0,0 +1,18 @@
+resourcePackages:
```
thescouser89: Do we need to write our own openapi.yaml?

facade/src/main/java/org/jboss/pnc/facade/rsql/RSQLProducerImpl.java
```diff
@@ -145,4 +146,9 @@ public Path apply(From<?, DB> from, RSQLSelectorPath selector) {
         };
     }
 
+    @Override
+    public <DTO> Comparator<DTO> getComparator(String rsql) {
+        throw new UnsupportedOperationException("Not supported yet."); //To change body of generated methods, choose Tools | Templates.
```
thescouser89: Is there a JIRA to implement it in the future?

rest-api/src/main/resources/openapi.yml
janinko: This is for general information. But it was supposed to be .json... :thinking:

facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
```diff
@@ -302,4 +324,192 @@ public void setDependentArtifacts(int id, List<Integer> artifactIds) {
...
+
+    /**
+     * Returns the page of Latest build (running or finished) filtered by given predicate.
+     */
+    private Page<Build> getLatestBuild(java.util.function.Predicate<BuildTask> predicate, Predicate<BuildRecord> dbPredicate) {
+        TreeSet<Build> sorted = new TreeSet<>(Comparator.comparing(Build::getSubmitTime).reversed());
+        readLatestRunningBuld(predicate).ifPresent(sorted::add);
+        readLatestFinishedBuild(dbPredicate).ifPresent(sorted::add);
+        if (sorted.size() > 1) {
+            sorted.pollLast();
```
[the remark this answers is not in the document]
janinko: I need a collection containing the first element. It could be done without the if by
```
Collections.singleton(sorted.first());
```
Do you find this more readable?

facade/src/main/java/org/jboss/pnc/facade/rsql/RSQLProducerImpl.java
janinko: yup

rest-api/src/main/resources/openapi.yml
janinko: and it was already added in b519b3da19e68006c08fcddc4a99113559f9df24. This file should not be here.

facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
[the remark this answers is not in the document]
janinko: ok, that's not possible, because there may not be any present and first() would throw exception in that case.

janinko:
@alexcreasy Just that you implemented the running/finished build mixing before, so to check how I did it now.

janinko:
Retest this please

janinko:
Retest this please

thescouser89:
retest this please

PR #2474, opened 2019-06-26
[NCL-4254] added field for DTO and moved a const

michalovjan:
### Checklist:

This change is needed to be merged before the Causeway part!

### DIFF
--- a/constants/src/main/java/org/jboss/pnc/constants/Patterns.java
+++ b/constants/src/main/java/org/jboss/pnc/constants/Patterns.java
@@ -53,5 +53,5 @@ public class Patterns {
      * the repository URL.
      */
     public static final String INTERNAL_REPOSITORY_NAME = "(\\/[\\w\\.:\\~_-]+)+(\\.git)(?:\\/?|\\#[\\d\\w\\.\\-_]+?)$";
-    
+
 }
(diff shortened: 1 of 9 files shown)


michalovjan:
@alexcreasy could you look at the naming of the new field in Build DTO. Is this ok or should we use better naming? That field signifies the URL used to clone git repo when building.

constants/src/main/java/org/jboss/pnc/constants/Patterns.java
```diff
@@ -53,5 +53,10 @@
      * the repository URL.
      */
     public static final String INTERNAL_REPOSITORY_NAME = "(\\/[\\w\\.:\\~_-]+)+(\\.git)(?:\\/?|\\#[\\d\\w\\.\\-_]+?)$";
-    
+
+    /**
+     * Represents an key in attribute of {@link org.jboss.pnc.model.ProductVersion}. It's value is
+     * tag prefix in Brew for a Version.
+     */
+    public static final String ATTRIBUTE_KEY_BREW_TAG_PREFIX = "BREW_TAG_PREFIX";
```
janinko: This is not a regex pattern, move it to an appropriate class.

michalovjan: Nice catch! Will do.

michalovjan:
retest this please

PR #2491, opened 2019-07-04
[NCL-4398] Fix create BC revision endpoint

janinko:
The problem here is that the new revision is created only when transaction is
commited (because "Envers flushes it's audit action queue in the transaction's
before-completion callback, which happens just before the commit occurs.").

I fixed it by creating new class with
  `@Transactional(Transactional.TxType.REQUIRES_NEW)`
to force new transactions.

### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigRevisionHelper.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigRevisionHelper.java
@@ -1,0 +1,89 @@
+/**
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014-2019 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.pnc.facade.providers;
+
+import org.jboss.pnc.dto.BuildConfigurationRevision;
+import org.jboss.pnc.facade.mapper.api.BuildConfigurationRevisionMapper;
+import org.jboss.pnc.model.BuildConfiguration;
+import org.jboss.pnc.model.BuildConfigurationAudited;
+import org.jboss.pnc.model.GenericEntity;
+import org.jboss.pnc.spi.datastore.repositories.BuildConfigurationAuditedRepository;
+import org.jboss.pnc.spi.datastore.repositories.BuildConfigurationRepository;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.security.PermitAll;
+import javax.ejb.Stateless;
+import javax.inject.Inject;
+
+import java.util.Objects;
+
+import javax.transaction.Transactional;
+
+@PermitAll
+@Stateless
+public class BuildConfigRevisionHelper {
+
+    private final Logger logger = LoggerFactory.getLogger(BuildConfigRevisionHelper.class);
+
+    @Inject
+    private BuildConfigurationAuditedRepository buildConfigurationAuditedRepository;
+
+    @Inject
+    private BuildConfigurationRepository buildConfigurationRepository;
+
+    @Inject
+    private BuildConfigurationRevisionMapper buildConfigurationRevisionMapper;
+    
+    @Transactional(Transactional.TxType.REQUIRES_NEW)
+    public void updateBuildConfiguration(org.jboss.pnc.model.BuildConfiguration bcEntity) {
+        buildConfigurationRepository.save(bcEntity);
+    }
+
+    public BuildConfigurationRevision findRevision(int id, BuildConfiguration bcEntity) {
+        return buildConfigurationAuditedRepository
+                .findAllByIdOrderByRevDesc(id)
+                .stream()
+                .peek(p-> logger.warn("going through: " + p))
+                .filter(bca -> equalValues(bca, bcEntity))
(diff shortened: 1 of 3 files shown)


facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigRevisionHelper.java
```diff
@@ -0,0 +1,89 @@
...
+public class BuildConfigRevisionHelper {
+
+    private final Logger logger = LoggerFactory.getLogger(BuildConfigRevisionHelper.class);
+
+    @Inject
+    private BuildConfigurationAuditedRepository buildConfigurationAuditedRepository;
+    @Inject
+    private BuildConfigurationRepository buildConfigurationRepository;
+    @Inject
+    private BuildConfigurationRevisionMapper buildConfigurationRevisionMapper;
```
jbartece: Code-style: Add empty line between 2 declarations

facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigRevisionHelper.java
```diff
@@ -0,0 +1,89 @@
...
+    
+    @Transactional(Transactional.TxType.REQUIRES_NEW)
+    public void updateBuildConfiguration(org.jboss.pnc.model.BuildConfiguration bcEntity) {
+        buildConfigurationRepository.save(bcEntity);
+    }
+
+    @Transactional(Transactional.TxType.REQUIRES_NEW)
+    public BuildConfigurationRevision findRevision(int id, BuildConfiguration bcEntity) {
```
jbartece: Why do we need a new transaction for a lookup?

facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationProviderImpl.java
```diff
@@ -94,6 +94,9 @@
     @Inject
     private SCMRepositoryProvider scmRepositoryProvider;
 
+    @Inject
+    private BuildConfigRevisionHelper buildConfigRevisionHelper;
+
```
jbartece: IMHO we should either inject the constructor or inject the fields. Combining both approaches looks confusing to me.

integration-test/src/test/java/org/jboss/pnc/integration_new/BuildConfigurationEndpointTest.java
```diff
@@ -81,4 +84,22 @@ public void shouldPatchBuildConfiguration() throws ClientException, PatchBuilder
         Assert.assertEquals(newDescription2, updated2.getDescription());
     }
 
+    @Test
+    public void shouldCreateBuildConfigRevision() throws ClientException {
+        BuildConfigurationClient client = new BuildConfigurationClient(RestClientConfiguration.getConfiguration(RestClientConfiguration.AuthenticateAs.USER));
+        BuildConfiguration bc = client.getSpecific(100);
+
+        BuildConfiguration newBC1 = bc.toBuilder().description("Updated description.").build();
+        BuildConfiguration newBC2 = bc.toBuilder().buildScript("mvn deploy # Updated script").build();
```
jbartece: I'd rather make both values final local variables.

facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigRevisionHelper.java
janinko: Hmm, what transaction isolation level do we have? Because there is an overlapping transaction in `BuildConfigurationProviderImpl`

facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigRevisionHelper.java
janinko: will fix

integration-test/src/test/java/org/jboss/pnc/integration_new/BuildConfigurationEndpointTest.java
janinko: I have no problem with that, but why?

jbartece: IMHO it is better to use final variables in tests rather then copy-paste magic-value

facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigRevisionHelper.java
jbartece: AFAIK we use the default transaction isolation level, which is Read Commited for Postgres 10: https://www.postgresql.org/docs/10/transaction-iso.html

jbartece: Could you explain more, why we need new transaction here?

As I understand it it is in the same transaction using the same PersistanceContext and therefore should be read from the PersistanceContext itself and won't go to the to read the entity, right?

janinko: Hmmm, yeah. If we have read committed then we probably don't need the transaction there.

integration-test/src/test/java/org/jboss/pnc/integration_new/BuildConfigurationEndpointTest.java
janinko: Hmm, so you mean it like this
```suggestion
        final BuildConfiguration newBC2 = bc.toBuilder().buildScript("mvn deploy # Updated script").build();
```
Or something else?

janinko: Oh, I guess you ment to extract the String parametr to final variable.

facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigRevisionHelper.java
janinko: Tests passed, seems you are right.

integration-test/src/test/java/org/jboss/pnc/integration_new/BuildConfigurationEndpointTest.java
jbartece: Yes, I mean to extract the String value

PR #2493, opened 2019-07-09
[NCL-4952] Map Group Configs DTOs to DB entities as active

janinko:
### Checklist:


michalovjan:
I think this also involves BuildConfiguration as we archive it too. 
@janinko

PR #2497, opened 2019-07-10
[v1.7][NCL-4910] Drop Indy client cache

pkocandr:
### Checklist:

### DIFF
--- a/build-executor/src/main/java/org/jboss/pnc/executor/DefaultBuildExecutor.java
+++ b/build-executor/src/main/java/org/jboss/pnc/executor/DefaultBuildExecutor.java
@@ -202,7 +202,7 @@ public class DefaultBuildExecutor implements BuildExecutor {
         }
         TargetRepository.Type repositoryType = BuildTypeToRepositoryType.getRepositoryType(buildType);
         //TODO use repositoryType
-        
+
         try {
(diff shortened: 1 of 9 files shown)


pkocandr:
Reworking the client takes time, but we need to address the issue now. I'm not saying we shouldn't do it, just probably not for 1.7. In the meantime we need a temporary solution. Not sure if dropping the cache completely is ok to do in respect to the performance impact. It woun't be anything huge of course, the client would be created just a few times for a build. But the resources consumption is probably worse.
It caches the client because it contains the connection pool. It was the only way how to reuse the same connection pool for multiple builds ran in parallel by single user. Without the cache it would multiply the connection pools by the number of builds running in parallel, but OTOH there would be no connections left when the builds finish.

jbartece:
I found out, that we use snapshot version of the client. Please include change of the version as well:     <indyVersion>1.8.0-SNAPSHOT</indyVersion>

jbartece:
Personally as a quick fix I'd drop the cache. It should be simpler and safe. I don't expect the performance penalty to be critical.

In 1.8 or 1.7.X we can reworking the client.
@matejonnet @vibe13 Wdyt?

pkocandr:
retest this please

PR #2503, opened 2019-07-11
[NCL-4965] JAX-RS method to specify alt HTTP code

thescouser89:
With the current strategy of implementing our REST endpoints, if
successful the JAX-RS method will always return with HTTP status 200 (or 204 if
we return void). This is because we now return the DTO in our JAX-RS
method instead of the `Response` object where we can specify the HTTP
status code.

This is problematic because we are expected to return HTTP status 201
(for creation of new stuff), and 202 for async accepted operations.

This commit adds a new annotation, `@RespondWithStatus`, to allow us to
annotate JAX-RS method with the appropriate HTTP status code. A filter
will intercept the annotation and update the HTTP to the one specified
in the annotation. The filter will only activate if a 2xx status is
returned from the JAX-RS method (typically 200 or 204)

The annotation strangely accepts the HTTP status code as a string so as
to be in harmony with the constants used in the OpenAPI documentation.
It is converted behind the scenes into an integer.

### Checklist:

### DIFF
--- a/rest-api/src/main/java/org/jboss/pnc/rest/annotation/RespondWithStatus.java
+++ b/rest-api/src/main/java/org/jboss/pnc/rest/annotation/RespondWithStatus.java
@@ -1,0 +1,29 @@
+/**
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014-2019 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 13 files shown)


rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/GroupBuildEndpoint.java
```diff
@@ -30,6 +30,7 @@
 import org.jboss.pnc.dto.response.Graph;
 import org.jboss.pnc.dto.response.Page;
 import org.jboss.pnc.processor.annotation.Client;
+import org.jboss.pnc.rest.annotation.RespondWithStatus;
```
janinko: not used?

PR #2505, opened 2019-07-15
[RR] fixed REST Exception handling when EJBException is thrown

michalovjan:
### Checklist:

@matedo1 This should resolve your message and status code issues.

### DIFF
--- a/rest-new/src/main/java/org/jboss/pnc/rest/provider/EJBExceptionMapper.java
+++ b/rest-new/src/main/java/org/jboss/pnc/rest/provider/EJBExceptionMapper.java
@@ -1,0 +1,56 @@
+/**
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014-2019 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.pnc.rest.provider;
+
+import org.jboss.pnc.dto.response.ErrorResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.ejb.EJBException;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+import javax.ws.rs.ext.Providers;
+
+import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+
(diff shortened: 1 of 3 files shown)


rest-new/src/main/java/org/jboss/pnc/rest/provider/EJBExceptionMapper.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override public Response toResponse(EJBException exception) {
+        Throwable t = exception.getCause();
+        ExceptionMapper mapper = providers.getExceptionMapper(t.getClass());
+        log.info("Unwrapping " + t.getClass().getSimpleName() + " from EJBException and passing in it to its appropriate ExceptionMapper");
+
+        if( mapper != null ) {
+            return mapper.toResponse(t);
+        }
+        else {
+            log.error("Could not find exception mapper for exception " + t);
```
janinko: Should there be `t.getClass()`?

rest-new/src/main/java/org/jboss/pnc/rest/provider/EJBExceptionMapper.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override public Response toResponse(EJBException exception) {
+        Throwable t = exception.getCause();
+        ExceptionMapper mapper = providers.getExceptionMapper(t.getClass());
+        log.info("Unwrapping " + t.getClass().getSimpleName() + " from EJBException and passing in it to its appropriate ExceptionMapper");
+
+        if( mapper != null ) {
+            return mapper.toResponse(t);
+        }
+        else {
```
janinko: code style
`} else {`

rest-new/src/main/java/org/jboss/pnc/rest/provider/EJBExceptionMapper.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(EJBExceptionMapper.class);
+
+    @Context
+    private Providers providers;
+
+    @Override public Response toResponse(EJBException exception) {
+        Throwable t = exception.getCause();
+        ExceptionMapper mapper = providers.getExceptionMapper(t.getClass());
+        log.info("Unwrapping " + t.getClass().getSimpleName() + " from EJBException and passing in it to its appropriate ExceptionMapper");
```
janinko: I would change the level to debug.

rest-new/src/main/java/org/jboss/pnc/rest/provider/EJBExceptionMapper.java
michalovjan: Yes, the wording is wrong there. Hmm, do we want to log just the Exception name or the whole stacktrace? This is a very specific case that should never happen so logging the whole stacktrace would be better for investigation I guess?

rest-new/src/main/java/org/jboss/pnc/rest/provider/EJBExceptionMapper.java
michalovjan: Why not.

rest-new/src/main/java/org/jboss/pnc/rest/provider/EJBExceptionMapper.java
michalovjan: My bad :D

rest-new/src/main/java/org/jboss/pnc/rest/provider/EJBExceptionMapper.java
janinko: Yeah, you are right, we should log whole stacktrace.

PR #2513, opened 2019-07-16
[NCL-4965] Fix NPE on RespondWithStatusFilter

thescouser89:
### Checklist:

### DIFF
--- a/rest-new/src/main/java/org/jboss/pnc/rest/provider/RespondWithStatusFilter.java
+++ b/rest-new/src/main/java/org/jboss/pnc/rest/provider/RespondWithStatusFilter.java
@@ -33,8 +33,12 @@ public class RespondWithStatusFilter implements ContainerResponseFilter {
     public void filter(ContainerRequestContext containerRequestContext, ContainerResponseContext containerResponseContext) throws IOException {
 
         // for any 2xx status code, this gets activated
-        if (containerResponseContext.getStatus() / 100 == 2) {
+        // Skip filter for Http Method OPTIONS to not break CORS
+        if (containerResponseContext.getStatus() / 100 == 2 &&
+            !containerRequestContext.getRequest().getMethod().equals("OPTIONS") &&
+            containerResponseContext.getEntityAnnotations() != null) {
             for (Annotation annotation : containerResponseContext.getEntityAnnotations()) {
+
                 if (annotation instanceof RespondWithStatus) {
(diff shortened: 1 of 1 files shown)


rest-new/src/main/java/org/jboss/pnc/rest/provider/RespondWithStatusFilter.java
```diff
@@ -33,8 +33,12 @@
     public void filter(ContainerRequestContext containerRequestContext, ContainerResponseContext containerResponseContext) throws IOException {
 
         // for any 2xx status code, this gets activated
-        if (containerResponseContext.getStatus() / 100 == 2) {
+        // Skip filter for Http Method OPTIONS to not break CORS
+        if (containerResponseContext.getStatus() / 100 == 2 &&
+            !containerRequestContext.getRequest().getMethod().equals("OPTIONS") &&
```
jbartece: Why are we skipping OPTIONS annotation?  What about about HEAD? Isn't it possible, that it will suffer with the same issue?

thescouser89: OPTIONS is used for CORS. Why should we skip the check for HEAD?

jbartece: It is used by CORS but also by other applications as it is standard HTTP call. I don't understand, why it should by disabled for this particular method. Is it due to performance or is it fixing the NPE?

thescouser89: It's disabled for OPTIONS as a precaution for CORS. We know it will cause an NPE since it doesn't go through the JAX-RS methods sequence and therefore there are no annotations to capture. If that's a big deal, I can remove the check and the Null check will disable the filter anyways.

jbartece: As we know, that there is a potential place for NPE we should solve it for all possible cases.
Yes, I'd do it by null check and left out the OPTIONS check as I think, it might be the same for HEAD operation as well

thescouser89: OK will create a PR to remove OPTIONS check then. I honestly don't see the point though.

PR #2516, opened 2019-07-17
Rework annotation RespondWithStatus not to parse String in every request

jbartece:
### Checklist:


janinko:
retest this please

thescouser89:
That's exactly what I wanted to do first but that was going to create a split between the OpenAPI documentation and what we're using with respect to variables.

If y'all think that's better, then :man_shrugging:

jbartece:
At the end it is just a constant with the same value and very similar name. I think, it is not an issue.

PR #2522, opened 2019-07-17
[NCL-4960] Appropriate http code for create-and-sync

thescouser89:
For the scm-repositories/create-any-sync endpoint, we should return the
appropriate HTTP status code for the following scenarios:

- scm repository already exists in the PNC database (HTTP code: 200)
- scm repository is being created (HTTP code: 202)

The implementation uses `HttpServletResponse`. The `flushBuffer` method
call is required, based on experiments. Otherwise the desired http code
is never set.

### Checklist:


janinko:
Other option would be throw WebApplicationException with a Response object.

PR #2530, opened 2019-07-25
Create a mapper module

thescouser89:
We should separate the mapper codebase from the facade module to allow
for more modularity.

The mapper code does not depend on the providers and therefore can be
extracted out.

This allows us to more easily use the mapper code in other modules,
while avoiding cyclic dependencies between modules

### Checklist:


mapper/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+                    <source>1.8</source>
+                    <target>1.8</target>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>org.mapstruct</groupId>
+                            <artifactId>mapstruct-processor</artifactId>
+                            <version>${version.mapstruct}</version>
+                        </path>
+                        <path>
+                            <groupId>org.projectlombok</groupId>
```
michalovjan: I don't think lombok-processor is needed here right?

mapper/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+        <dependency>
+            <groupId>org.jboss.pnc</groupId>
+            <artifactId>spi</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.pnc</groupId>
+            <artifactId>model</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.hibernate</groupId>
```
michalovjan: I might be mistaken but I don't think we use hibernate in mappers.

pom.xml
```diff
@@ -95,6 +95,7 @@
     <module>process-managers</module>
     <module>processor</module>
     <module>rest-client</module>
+      <module>mapper</module>
```
michalovjan: Weird indent?

thescouser89: Fixed!

mapper/pom.xml
thescouser89: If I don't have it, the compilation fails with:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.5.1:compile (default-compile) on project mapper: Compilation failure
[ERROR] /home/dcheung/projects/work/pnc/mapper/src/main/java/org/jboss/pnc/mapper/api/BuildConfigurationMapper.java:[37,8] Internal error in the mapping processor: com.sun.tools.javac.code.Symbol$CompletionFailure: class file for org.hibernate.engine.spi.ManagedEntity not found  
[ERROR] 
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <goals> -rf :mapper

```

michalovjan: Huh.. ok then

thescouser89:
:heart:

PR #2553, opened 2019-08-13
[NCL-4973] Add Javadoc for notifications

janinko:


constants/src/main/java/org/jboss/pnc/enums/JobNotificationProgress.java
```diff
@@ -18,12 +18,20 @@
...
+    /**
+     * The job is waiting. For example build waiting for dependencies.
+     */
     PENDING,
+    /**
+     * The job is running. For example build is building.
+     */
     IN_PROGRESS,
+    /**
+     * The job has finished. For example build failed.
```
alexcreasy: You're clearly a glass half full guy :)

janinko: Well, I was thinking that example "build succeeded" was obvious and that it can represent also a failure was less obvious :)

PR #2563, opened 2019-08-21
[NCL-4872] Change REST DTOs to use String ids

thescouser89:
### Checklist:


janinko:
The id in URLs in rest-api should also be changed to String I think. E.g. https://github.com/project-ncl/pnc/blob/master/rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ArtifactEndpoint.java#L95

thescouser89:
@janinko I'd like to do the change in a later commit to not increase the size of this PR :)

janinko:
Ok, let's just merge it and hope for the best.

PR #2573, opened 2019-08-27
Add REST tests

janinko:
### Checklist:

Obviously


facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
```diff
@@ -83,6 +83,8 @@
 import java.util.Spliterators;
 import java.util.TreeSet;
 import java.util.stream.StreamSupport;
+import javax.annotation.security.DenyAll;
```
michalovjan: unused import

PR #2580, opened 2019-08-30
[NCL-5007] Unit tests for UserProvider

thescouser89:
### Checklist:

### DIFF
--- a/facade/src/test/java/org/jboss/pnc/facade/providers/UserProviderImplTest.java
+++ b/facade/src/test/java/org/jboss/pnc/facade/providers/UserProviderImplTest.java
@@ -79,7 +79,7 @@ public class UserProviderImplTest extends AbstractProviderTest<User> {
         when(repository.queryByPredicates(any(Predicate.class))).thenAnswer(inv -> user);
 
         // When
-        org.jboss.pnc.dto.User userDTO = provider.getOrCreateNewUser("test");
+        org.jboss.pnc.dto.User userDTO = provider.getOrCreateNewUser(user.getUsername());
 
         // then
         assertThat(userDTO.getId()).isEqualTo(user.getId().toString());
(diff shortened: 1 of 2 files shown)


janinko:
@thescouser89 Why did you not added unit tests?
image
;-)

facade/src/test/java/org/jboss/pnc/facade/providers/UserProviderImplTest.java
```diff
@@ -0,0 +1,180 @@
...
+        // when
+        org.jboss.pnc.dto.User userDTO = org.jboss.pnc.dto.User.builder()
+                .username("john-cormack")
+                .build();
+
+        org.jboss.pnc.dto.User userDTOSaved = provider.store(userDTO);
+
+        // then
+        assertThat(userDTOSaved.getId()).isNotNull().isNotBlank();
+        assertThat(userDTOSaved.getUsername()).isEqualTo(userDTO.getUsername());
```
janinko: I think we don't support of creating users via REST.

facade/src/test/java/org/jboss/pnc/facade/providers/UserProviderImplTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+    @Test
+    public void testGetExistingUser() {
+        // Prepare
+        User user = repositoryList.get(0);
+
+        when(repository.queryByPredicates(any(Predicate.class))).thenAnswer(inv -> user);
+
+        // When
+        org.jboss.pnc.dto.User userDTO = provider.getOrCreateNewUser("test");
```
janinko: I know when it's so mocked it actually doesn't have an effect, but I think you should still ask for a username that you actually receive.
```suggestion
        org.jboss.pnc.dto.User userDTO = provider.getOrCreateNewUser(user.getUsername());
```

facade/src/test/java/org/jboss/pnc/facade/providers/UserProviderImplTest.java
thescouser89: I believe we do for new users. If a user is not part of the PNC database, and starts to use PNC, it'll hit the REST url '/users/current', which will automatically add it to the database so that we can use it when noting who built a build config. The username is obtained from the Keycloak server.

thescouser89:
Thanks for the reviews @janinko ! Much much much appreciated buddy! :)

PR #2641, opened 2019-09-19
NCL-4200 fixes some links that were missed in initial refactor

alexcreasy:
### Checklist:


alexcreasy:
@matedo1 this should fix the links that were pointing to build-groups.* ... I've updated them so they will work post-refactor of group-builds (i.e. making use of GroupConfigRef for the link), right now they are broken in a different way.

PR #2650, opened 2019-09-26
Removed BuildRecords from BuildConfigurationAudited to simplify model…

vibe13:
… and avoid unnecessary queries

### Checklist:


vibe13:
@alexcreasy @janinko @matejonnet @jbartece I have removed entirely the buildRecords from buildConfigurationAudited. They are expensively retrieved from the db (only their IDs), and never used by any piece of code. Moreover, the buildConfigurationAudited.buildRecords() are ignored in BuildRecordProvider.toRESTModel(), so any endpoint which converts BuildRecord to BuildRecordRest would be more lightweight with this change. All tests passed with no complain whatsoever.

PR #2651, opened 2019-09-26
[v1.7] [NCL-5192] fix temporaryBuilds falsely rebuilding if there is newer persistent record

michalovjan:
### Checklist:

### DIFF
--- a/datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
+++ b/datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
@@ -376,6 +376,7 @@ public class DefaultDatastore implements Datastore {
             return true;
         }
         if (checkImplicitDependencies) {
+            logger.debug("Checking if BCA: {} has implicit dependencies that need rebuild", idRev);
             boolean rebuild = hasARebuiltImplicitDependency(latestSuccessfulBuildRecord, temporaryBuild);
             logger.debug("Implicit dependency check for rebuild of buildConfiguration.idRev: {} required: {}.", idRev, rebuild);
             if (rebuild) {
@@ -422,7 +423,13 @@ public class DefaultDatastore implements Datastore {
     private boolean hasARebuiltImplicitDependency(BuildRecord latestSuccessfulBuildRecord, boolean temporaryBuild) {
         Collection<BuildRecord> lastBuiltFrom = getRecordsUsedFor(latestSuccessfulBuildRecord);
         return lastBuiltFrom.stream()
-                .anyMatch(br -> hasNewerVersion(br, temporaryBuild));
+                .anyMatch(br -> {
+                    if(hasNewerVersion(br, temporaryBuild)) {
+                        logger.debug("Latest successful BuildRecord: {} has implicitly dependent BR: {} that requires rebuild.", latestSuccessfulBuildRecord.getId(), br.getId());
+                        return true;
+                    }
+                    return false;
+                });
     }
 
     /**
--- a/spi/src/main/java/org/jboss/pnc/spi/datastore/repositories/BuildRecordRepository.java
+++ b/spi/src/main/java/org/jboss/pnc/spi/datastore/repositories/BuildRecordRepository.java
@@ -25,6 +25,7 @@ import org.jboss.pnc.spi.datastore.repositories.api.Predicate;
 import org.jboss.pnc.spi.datastore.repositories.api.Repository;
 import org.jboss.pnc.spi.datastore.repositories.api.SortInfo;
 
+import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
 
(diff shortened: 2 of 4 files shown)


spi/src/main/java/org/jboss/pnc/spi/datastore/repositories/BuildRecordRepository.java
```diff
@@ -48,20 +48,22 @@
...
-                 *  |         FALSE          |          TRUE              |        TRUE           |
-                 *  |         FALSE          |          FALSE             |        FALSE          |
-                 */
-                .filter(record -> buildTemporary || !record.isTemporaryBuild() )
-                .sorted((o1, o2) -> -o1.getId().compareTo(o2.getId()))
-                .findFirst().orElse(null);
+                //First part includes temporary BRs and second part includes persitent BRs
+                .filter(record -> (includeTemporary && record.isTemporaryBuild()) ||
+                                  (includePersistent && !record.isTemporaryBuild()))
+                .min((o1, o2) -> -o1.getId().compareTo(o2.getId()))
```
jbartece: I'd change it to max and removed the -.

datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
```diff
@@ -419,10 +419,16 @@ private boolean isLatestSuccessBRFromThisBCA(BuildConfigurationAudited buildConf
      * Check is some of the dependencies from the previous build were rebuild.
      * Checking is done based on captured dependencies which are stored in the Build Record.
      */
-    private boolean hasARebuiltImplicitDependency(BuildRecord latestSuccessfulBuildRecord, boolean temporaryBuild) {
+    private boolean hasARebuiltImplicitDependency(BuildRecord latestSuccessfulBuildRecord, boolean temporaryBuild, IdRev idRev) {
```
matejonnet: Adding a parameter that is used for logging only is not the best practice. If the parameter is not used in the logic, why is it important to log that. What about logging the operation where the method is called ?

spi/src/main/java/org/jboss/pnc/spi/datastore/repositories/BuildRecordRepository.java
```diff
@@ -48,20 +48,22 @@
...
-                 *  |         TRUE           |          FALSE             |        TRUE           |
-                 *  |         FALSE          |          TRUE              |        TRUE           |
-                 *  |         FALSE          |          FALSE             |        FALSE          |
-                 */
-                .filter(record -> buildTemporary || !record.isTemporaryBuild() )
-                .sorted((o1, o2) -> -o1.getId().compareTo(o2.getId()))
-                .findFirst().orElse(null);
+                //First part includes temporary BRs and second part includes persitent BRs
+                .filter(record -> (includeTemporary && record.isTemporaryBuild()) ||
+                                  (includePersistent && !record.isTemporaryBuild()))
```
matejonnet: What about including the one or another filter based on the condition? It would make the filters simpler and more readable.

janinko: On what condition? The `buildTemporary`? I think that would make it simple only in one (persistent) branch.

janinko: Hmm, it would look like this:
```java
Predicate<BuildRecord> filter;
if(buildTemporary){
    final boolean containsTemporary = buildRecords.stream().anyMatch(BuildRecord::isTemporaryBuild);
    final boolean includePersistent = !containsTemporary;
    filter = record -> record.isTemporaryBuild() || includePersistent;
}else{
    filter = record -> !record.isTemporaryBuild();
}
```

datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
michalovjan: Yeah, I get you. I was trying to find out how to log in the cleanest way. I had a problem that this method didn't return any information about which implicit dependency triggered the rebuild.

Solution to this could be changing the signature of this method to return List<BuildRecord> and log it on a higher level.

matejonnet: I see. Returning a list is also not nice. What about adding a log line (eg. checking if idRev hasARebuiltImplicitDependency) before the method call? Or leave it with the new parameter.

michalovjan: I don't have a strong opinion on this. With the first option, there could be confusion in the situation when there is concurrent implicit checking done for 2 builds at the same time. You may run into a situation when you can't differentiate which implicit dependency belongs to which IdRev

michalovjan: Something like this:
'''
line 1: Checking whether IdRev id:1 rev:1 has implicit dependencies that need rebuild.
line 2: Checking whether IdRev id:2 rev:2 has implicit dependencies that need rebuild.
line 3: Latest successful BuildRecord: 5 has has implicitly dependent BR: 4 that requires rebuild. (belongs to id:2 rev:2)
line 4 :Latest successful BuildRecord: 8 has has implicitly dependent BR: 3 that requires rebuild. (belongs to id:1 rev:1)
'''

janinko: Parameter only for logging is the more justifiable the higher the logging level is. I personally would draw the line on the info level. Is this log message important enough to info level? If no I would be fine having it on two lines (thread id is usually part of the logline, so you can distinguish it).

spi/src/main/java/org/jboss/pnc/spi/datastore/repositories/BuildRecordRepository.java
matejonnet: This is even harder to read.

datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
michalovjan: From my personal experience, this logline would save me 2-3 hours of finding out which implicitly-dependent build record caused the rebuild. I think it needs to be in DEBUG log level as we currently don't have any straightforward way of finding these build records.

Good catch, I've forgotten about thread id. With that, it should be perfectly fine having it in more lines. :+1:

michalovjan:
Closing as we are not releasing 1.7.2.

PR #2659, opened 2019-09-27
[v1.8] - NCL-5143 Remove deprecated Hibernate annotations and use JPA annotations; add …

vibe13:
…names for all UK and FK

### Checklist:


vibe13:
I have removed the deprecated Hibernate annotations and add all names to UK and FK to have all corresponding names with db (and not autogenerated ones which are very ugly and confusing).

PR #2676, opened 2019-10-07
NCL-5184 - Optimize queries to minimize CPU load and DB usage when hitting particular endpoints

vibe13:
### Checklist:


vibe13:
Merging as already been reviewed and verified in branch 1.8

janinko:
But this does not modify the behaviour of the new API in rest-new, is that correct?

PR #2682, opened 2019-10-14
[NCL-5257] Map also entities from BCA in BuildBCRevisionFetcher

janinko:

### DIFF
--- a/mapper/src/main/java/org/jboss/pnc/mapper/BuildBCRevisionFetcher.java
+++ b/mapper/src/main/java/org/jboss/pnc/mapper/BuildBCRevisionFetcher.java
@@ -18,8 +18,14 @@
 package org.jboss.pnc.mapper;
 
 import org.jboss.pnc.dto.Build;
-import org.jboss.pnc.dto.BuildConfigurationRevision;
+import org.jboss.pnc.dto.BuildConfigurationRevisionRef;
+import org.jboss.pnc.dto.Environment;
+import org.jboss.pnc.dto.ProjectRef;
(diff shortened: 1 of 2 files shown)


mapper/src/main/java/org/jboss/pnc/mapper/BuildBCRevisionFetcher.java
```diff
@@ -48,10 +66,20 @@ public void mockBrewAttributes(BuildRecord build, @MappingTarget Build.Builder d
         Integer id = build.getBuildConfigurationId();
         Integer revision = build.getBuildConfigurationRev();
 
-        BuildConfigurationAudited buildConfigurationAudited = bcAuditedRepository
-                .queryById(new IdRev(id, revision));
+        // If somebody before us already set the BCA we don't need to query it from DB again
+        BuildConfigurationAudited bca = build.getBuildConfigurationAudited();
+        if(bca == null){
+            bca = bcAuditedRepository.queryById(new IdRev(id, revision));
+        }
```
janinko: @vibe13 I added the null check to skip the DB query if it is already set. I don't think that current implementation retrieves the BCA sometimes before this, but it can be useful if you find a place that can be optimized e.g. by querying a bunch of BCAs in one query or some other optimization.

janinko:
retest this please

janinko:
retest this please

vibe13:
@janinko  I could not spot any obvious place where to apply some optimizations. I only created one in facade package, see https://github.com/project-ncl/pnc/pull/2707

PR #2697, opened 2019-10-21
[v1.8][NCL-5267] Log PNC build related stages begin and end

thescouser89:
The build stages logged are:

- Scheduling (when a build config is being broken down into build tasks)
- Enqueued (build stayed in enqueued state)

- The various stages in `BuildExecutionStatus`
  * NEW,
  * REPO_SETTING_UP,
  * BUILD_ENV_SETTING_UP,
  * BUILD_ENV_WAITING,
  * BUILD_ENV_SETUP_COMPLETE_SUCCESS,
  * BUILD_SETTING_UP,
  * BUILD_WAITING,
  * COLLECTING_RESULTS_FROM_BUILD_DRIVER,
  * COLLECTING_RESULTS_FROM_REPOSITORY_MANAGER,
  * REPOSITORY_MANAGER_SEALING_TRACKING_RECORD,
  * REPOSITORY_MANAGER_DOWNLOADING_TRACKING_REPORT,
  * REPOSITORY_MANAGER_PROCESSING_DEPENDENCIES,
  * REPOSITORY_MANAGER_PROCESSING_BUILT_ARTIFACTS,
  * REPOSITORY_MANAGER_REMOVE_BUILD_AGGREGATION_GROUP,
  * REPOSITORY_MANAGER_PROMOTION_BUILD_CONTENT_SET,
  * COLLECTING_RESULTS_FROM_REPOSITORY_MANAGER_COMPLETED_SUCCESS,
  * BUILD_COMPLETED_SUCCESS,
  * BUILD_ENV_DESTROYING,
  * BUILD_ENV_DESTROYED,
  * FINALIZING_EXECUTION,

While this PR doesn't try to log into Kafka directly, it is useful to identify if everything is being logged properly in Kibana. The next step would be to modify the static method logging the changes to send the data to a particular kafka topic.

### Checklist:

### DIFF
--- a/common/src/main/java/org/jboss/pnc/common/util/ProcessStageUtils.java
+++ b/common/src/main/java/org/jboss/pnc/common/util/ProcessStageUtils.java
@@ -1,0 +1,66 @@
+/**
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.pnc.common.util;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class ProcessStageUtils {
+
+    private static final Logger log = LoggerFactory.getLogger("org.jboss.pnc._userlog_.process-stage-update");
+    public static final String MDC_PROCESS_STAGE_NAME = "process_stage_name";
+    public static final String MDC_PROCESS_STAGE_STEP = "process_stage_step";
+
+    /**
+     * Log the process stage's begin and end stage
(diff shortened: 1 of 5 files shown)


thescouser89:
output

This is the output from Kibana when filtering for 'mdc.build_stage_step' and 'mdc.build_stage_name'.

common/src/main/java/org/jboss/pnc/common/util/ProcessStageUtils.java
```diff
@@ -0,0 +1,71 @@
...
+            mdc = new HashMap<>();
+        }
+
+        try (MDC.MDCCloseable a = MDC.putCloseable(MDC_PROCESS_STAGE_NAME, processStage);
+             MDC.MDCCloseable b = MDC.putCloseable(MDC_PROCESS_STAGE_STEP, step.toString())) {
+
+            log.info("{}: {} ", step, processStage);
+        }
+
+        // cleanup
```
matejonnet: cleanup is not required, closeable takes care of it.

thescouser89: Good catch! I should have removed it from the beginning. My fault

thescouser89:
> What about moving build execution timing DBESession to DefaultBuildExecutor.startBuilding to have start and end around each stage method invocations ?

Sure that works. Will do the change

thescouser89:
> What about moving build execution timing DBESession to DefaultBuildExecutor.startBuilding to have start and end around each stage method invocations ?

I tried to do it, but I don't think I can keep everything in DefaultBuildExecutor. The various stages of 'retrieveRepositoryManagerResults' happen in the MavenRepositorySession, which provides detailed metrics on what's going on with Indy.

The various states that Indy are in are indicated by the changes in status.

Also it becomes a little bit awkward with the CompletableFutures (for e.g method 'runTheBuild'), where I have to print the end state for 'cancel', 'onComplete', and 'onError'.

I'm happy to discuss this further, but I don't think right now that would be a beneficial change.

PR #2704, opened 2019-10-24
[NCL-5216] BuildTest integration tests

michalovjan:
### Checklist:

### DIFF
--- a/integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildEndpointTest.java
+++ b/integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildEndpointTest.java
@@ -239,7 +239,8 @@ public class BuildEndpointTest {
 
     @Test
     public void shouldFailToGetSshCredentialsForUserThatDidntTrigger() {
-        BuildClient client = new BuildClient(RestClientConfiguration.asUser());
+        BuildClient client = new BuildClient(RestClientConfiguration.getConfiguration(
+                RestClientConfiguration.AuthenticateAs.USER2));
 
         assertThatThrownBy(() -> client.getSshCredentials(buildId))
                 .hasCauseInstanceOf(ForbiddenException.class); // 403 means not authorized
--- a/integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildTest.java
+++ b/integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildTest.java
@@ -1,0 +1,323 @@
+/**
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014-2019 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.pnc.integration_new.endpoint;
+
+import org.jboss.arquillian.container.test.api.Deployment;
+import org.jboss.arquillian.container.test.api.RunAsClient;
+import org.jboss.arquillian.junit.Arquillian;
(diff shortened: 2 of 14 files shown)


michalovjan:
PR-tests were successful on the first try.... WAAAAAAAAAT

thescouser89:
** Mind games **

integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildEndpointTest.java
```diff
@@ -239,7 +239,7 @@ public void shouldGetBuildLogs() throws ClientException {
 
     @Test
     public void shouldFailToGetSshCredentialsForUserThatDidntTrigger() {
-        BuildClient client = new BuildClient(RestClientConfiguration.asUser());
+        BuildClient client = new BuildClient(RestClientConfiguration.asSystem());
```
janinko: Why did you cahnge this?

rest-api/src/main/java/org/jboss/pnc/rest/api/parameters/BuildParameters.java
```diff
@@ -36,19 +36,15 @@
...
     @Parameter(description = "What should varant rebuild?")
     @QueryParam("rebuildMode")
+    @DefaultValue("IMPLICIT_DEPENDENCY_CHECK")
     RebuildMode rebuildMode;
 
     @Parameter(description = "Should we build also dependencies of this BuildConfiguration?")
     @QueryParam("buildDependencies")
     @DefaultValue("true")
-    boolean buildDependencies;
+    Boolean buildDependencies;
```
janinko: what if it would be 
```suggestion
    boolean buildDependencies = true;
```
Would that work? Would that break something?

integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildTest.java
```diff
@@ -0,0 +1,323 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(BuildTest.class);
+
+    BuildConfigurationClient buildConfigurationClient = new BuildConfigurationClient(asUser());
+
+    GroupConfigurationClient groupConfigurationClient = new GroupConfigurationClient(asUser());
+
+    GroupBuildClient groupBuildClient = new GroupBuildClient(asUser());
+
+    BuildClient buildClient = new BuildClient(asUser());
```
janinko: private?

integration-test/src/test/java/org/jboss/pnc/integration_new/setup/RestClientConfiguration.java
```diff
@@ -34,7 +34,7 @@ public static Configuration getConfiguration(AuthenticateAs authAs) {
         if (AuthenticateAs.SYSTEM_USER.equals(authAs)) {
             builder.basicAuth(new Configuration.BasicAuth("system", "system.1234"));
         } else if (AuthenticateAs.USER.equals(authAs)) {
-            builder.basicAuth(new Configuration.BasicAuth("user", "pass.1234"));
+            builder.basicAuth(new Configuration.BasicAuth("demo-user", "pass.1234"));
```
janinko: Why did you change this?

rest-api/src/main/java/org/jboss/pnc/rest/api/parameters/BuildParameters.java
michalovjan: Then you would always have buildDependencies=true in a request. I don't have a strong opinion on this.

integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildEndpointTest.java
michalovjan: I had to change authentication to "demo-user" because as old "user" when building inside the code I would have User as null because we don't have user with username:"user" in our demo-data. 

In this case that buildRecord, according to demo-data, was built by demo-user so I had to change the user in request. The "system" was our only other user to choose from.

integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildTest.java
michalovjan: ya, mb

integration-test/src/test/java/org/jboss/pnc/integration_new/setup/RestClientConfiguration.java
michalovjan: I had to change authentication to "demo-user" because as old "user" when building inside the code I would have User as null because we don't have user with username:"user" in our demo-data.

PR #2709, opened 2019-10-29
Ncl 5167

janinko:
### Checklist:


janinko:
retest this please

dto/src/main/java/org/jboss/pnc/dto/TargetRepository.java
```diff
@@ -36,17 +35,16 @@
 @ToString(callSuper = true)
 @EqualsAndHashCode(callSuper = true)
 @JsonDeserialize(builder = TargetRepository.Builder.class)
+@JsonIgnoreProperties(ignoreUnknown = true)
 public class TargetRepository extends TargetRepositoryRef {
```
michalovjan: What is the point of now having TargetRepository full DTO?

janinko: None, it's not even used.

janinko: And never have been.

michalovjan: If we removed it, would it cause issues? It's just for clean code.

janinko:
retest this please

dto/src/main/java/org/jboss/pnc/dto/TargetRepository.java
janinko: Removed

PR #2711, opened 2019-10-29
[NCL-4581] Add health check endpoint for PNC

thescouser89:
Right now the health check done are:

- check that we can read from database
- check that we can write to database

The endpoint will return 500 if *any* of the checks fail, otherwise 200.

More checks can be added in the `HealthCheckProviderImpl` in the future.

We decided to not use the Microprofile health since it is in technology
preview for EAP 7.2.x. At the point of writing, we are using EAP 7.2.x
for PNC.

### Checklist:

### DIFF
--- a/rest-new/src/main/java/org/jboss/pnc/rest/endpoints/internal/HealthCheckEndpointImpl.java
+++ b/rest-new/src/main/java/org/jboss/pnc/rest/endpoints/internal/HealthCheckEndpointImpl.java
@@ -1,0 +1,54 @@
+/**
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014-2019 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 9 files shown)


thescouser89:
If this PR is approved, I'll backport the changes to v1.8 also and add the SQL upgrade.

thescouser89:
retest this please

thescouser89:
Thank you Jenkins

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/internal/HealthCheckEndpointImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    public Response check() {
+        return generateResponseFromResult(healthCheckProvider.check());
+    }
+
+    private Response generateResponseFromResult(Map<String, Boolean> result) {
+        if (result == null) {
+            return Response.serverError().build();
+        } else {
+
+            boolean atLeastOneCheckFailed = result.entrySet().stream().anyMatch(e -> !e.getValue());
```
janinko:
```suggestion
            boolean atLeastOneCheckFailed = result.values().stream().anyMatch(v -> !v);
```

thescouser89:
@vibe13 

1. Yeah pnc-status, and the Openshift health check. If the Openshift health check fails, it'll kill the pod, and that's when we'll know something is very wrong.

2. I was thinking of commiting the SQL script in the 1.8 PR just so that I can create the script with the correct <from>-<to>.sql script. Let me know if that's ok!

3. Yeah! Hopefully EAP 7.3. Shouldn't be too big of a change I think anyways.

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/internal/HealthCheckEndpointImpl.java
thescouser89: Ah that makes sense since I don't care about the keys. Thanks Honza for the suggestion! :)

thescouser89:
Woops actually you're right. I should commit the sql file here because of branches. Me dumb dumb

PR #2712, opened 2019-10-30
fix PR #2704

michalovjan:
Related PR #2704 
### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
@@ -47,6 +47,7 @@ import org.jboss.pnc.model.BuildConfiguration;
 import org.jboss.pnc.model.BuildConfigurationAudited;
 import org.jboss.pnc.model.BuildRecord;
 import org.jboss.pnc.model.IdRev;
+import org.jboss.pnc.model.User;
 import org.jboss.pnc.spi.coordinator.BuildCoordinator;
 import org.jboss.pnc.spi.coordinator.BuildTask;
 import org.jboss.pnc.spi.datastore.predicates.BuildRecordPredicates;
@@ -70,7 +71,6 @@ import javax.annotation.security.RolesAllowed;
 import javax.ejb.EJBAccessException;
 import javax.ejb.Stateless;
 import javax.inject.Inject;
-
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
@@ -227,7 +227,13 @@ public class BuildProviderImpl extends AbstractIntIdProvider<BuildRecord, Build,
     @Override
     public SSHCredentials getSshCredentials(String id) {
         BuildRecord buildRecord = getBuildRecord(id);
-        if(!buildRecord.getUser().equals(userService.currentUser())){
+        User user = null;
+        try {
+            user = userService.currentUser();
+        } catch(IllegalStateException e) {
+            //leaves user null
+        }
+        if(!buildRecord.getUser().equals(user)) {
             throw new EJBAccessException("Only user who executed the build is allowed to get the SSH credentials");
         }
 
--- a/integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildTest.java
+++ b/integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildTest.java
@@ -57,6 +57,7 @@ import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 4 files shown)


michalovjan:
retest this please

facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
```diff
@@ -227,7 +227,13 @@ public String getBuildLog(String id) {
     @Override
     public SSHCredentials getSshCredentials(String id) {
         BuildRecord buildRecord = getBuildRecord(id);
-        if(!buildRecord.getUser().equals(userService.currentUser())){
+        User user = null;
+        try {
+            user = userService.currentUser();
+        } catch(IllegalStateException e) {
```
janinko: I would let it throw.

integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildTest.java
```diff
@@ -284,10 +284,10 @@ private Boolean buildToFinish(String buildId, EnumSet<BuildStatus> isIn, EnumSet
             logger.debug("Gotten build with status: {}", build.getStatus());
             if (!build.getStatus().isFinal()) return false;
         } catch (RemoteResourceNotFoundException e) {
-            //expected
+            logger.error("Group Build with id:{} not present", buildId);
```
janinko: The test should FAIL immediately in this case.

integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildTest.java
```diff
@@ -309,10 +309,10 @@ private Boolean groupBuildToFinish(String groupBuildId, EnumSet<BuildStatus> isI
             logger.debug("Gotten build with status: {}", build.getStatus());
             if (!build.getStatus().isFinal()) return false;
         } catch (RemoteResourceNotFoundException e) {
-            //expected
+            logger.error("Group Build with id:{} not present", groupBuildId);
```
janinko: The test should FAIL immediately in this case.

rest-new/src/main/java/org/jboss/pnc/rest/RequestLoggingFilter.java
```diff
@@ -75,7 +75,12 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
             MDCUtils.addProcessContext(logProcessContext);
         }
 
-        User user = userService.currentUser();
+        User user = null;
+        try {
+            user = userService.currentUser();
+        } catch (IllegalStateException e) {
+            //user not found.. proceed
+        }
```
janinko: I would maybe move the bellow if inside the try?

facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
michalovjan: If I let it throw, the Rest will return 500SystemError and not 403Forbidden as it should be.

integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildTest.java
michalovjan: sure

integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildTest.java
michalovjan: sure

rest-new/src/main/java/org/jboss/pnc/rest/RequestLoggingFilter.java
michalovjan: sure

facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
janinko: It should not throw 403 Forbidden, because the user is authenticated.

michalovjan: It should throw 403 Forbidden because the user is not the one from BR and not because he is not authenticated. In reality, this should never happen because if a user built something, his entity is in the BuildRecord.

michalovjan: still makes problems in integration-tests though

janinko: You are right. And I mistook 401 Unautorized and 403 Forbidden again.

PR #2715, opened 2019-10-31
[NCL-5230] Show ErrorResponse message in client exceptions

janinko:
### Checklist:

### DIFF
--- a/processor/src/main/java/org/jboss/pnc/processor/ClientGenerator.java
+++ b/processor/src/main/java/org/jboss/pnc/processor/ClientGenerator.java
@@ -206,7 +206,6 @@ public class ClientGenerator extends AbstractProcessor {
                         MethodSpec.Builder methodBuilder = beginMethod(restApiMethod);
                         addDefaultParameters(restApiMethod, methodBuilder);
                         methodBuilder
-                                .addException(ClassName.get("org.jboss.pnc.client", "ClientException"))
                                 .returns(TypeName.get(returnType))
                                 .addStatement("return getEndpoint()." + restApiMethod.getSimpleName() + "(" + parametersList + ")")
                                 .nextControlFlow("catch ($T e)", NotFoundException.class)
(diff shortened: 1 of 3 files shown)


rest-api/src/main/java/org/jboss/pnc/client/RemoteResourceException.java
```diff
@@ -17,33 +17,45 @@
...
 /**
  * @author <a href="mailto:matejonnet@gmail.com">Matej Lazar</a>
  */
 public class RemoteResourceException extends ClientException {
 
+    @Getter
     private final int status;
 
+    @Getter
+    private final ErrorResponse response;
```
matejonnet: What about making it Optional, to protect users form NPE?

rest-api/src/main/java/org/jboss/pnc/client/RemoteResourceException.java
```diff
@@ -17,33 +17,45 @@
...
-        status = -1;
+        this.status = -1;
+        this.response = null;
     }
 
     public RemoteResourceException(ClientErrorException cause) {
         super(cause);
-        Response response = cause.getResponse();
-        status = response.getStatus();
+        this.status = cause.getResponse().getStatus();
```
matejonnet: is this NPE prof ? Is getResponse() always not null ?

janinko: I just put it on one line, I didn't change behaviour :)

But it should be always set AFAIK.

processor/src/main/java/org/jboss/pnc/processor/ClientGenerator.java
```diff
@@ -206,7 +206,6 @@ private boolean process0(Set<? extends TypeElement> annotations, RoundEnvironmen
                         MethodSpec.Builder methodBuilder = beginMethod(restApiMethod);
                         addDefaultParameters(restApiMethod, methodBuilder);
                         methodBuilder
-                                .addException(ClassName.get("org.jboss.pnc.client", "ClientException"))
```
matejonnet: Why did you remove the exception from the message signature ?

janinko: Because this exception is not thrown anywhere.

janinko:
retest this please

PR #2716, opened 2019-11-01
[v1.8][NCL-5254] Move the keycloak client secret to a path

thescouser89:
This is done so that we don't leak our secret in our main config file.

As such, the change required in the config file is from:

```
credentials: {
    "secret": "super-secret-token"
}
```

to:

```
credentials: {
    "secretFileLocation": "/path/to/file/that/contains/token"
}
```

### Checklist:


thescouser89:
I've modified tests jsons, if that helps.

PR #2738, opened 2019-11-20
Generate openapi spec at build time

alexcreasy:
### Checklist:


alexcreasy:
@janinko any objections to merging this to generate the open api swagger at build time (it also attaches it to the deploy output so will get published to snapshot repo). I didn't put this in a profile as the effect on build time appears to be microscopic.

PR #2754, opened 2019-12-05
NCL-5353 - Add MDC context passing to headers.

matejonnet:
### Checklist:

### DIFF
--- a/rest-api/src/main/java/org/jboss/pnc/client/Configuration.java
+++ b/rest-api/src/main/java/org/jboss/pnc/client/Configuration.java
@@ -18,16 +18,17 @@
 package org.jboss.pnc.client;
 
 import lombok.AllArgsConstructor;
-import lombok.Builder;
 import lombok.Data;
 import lombok.Getter;
+import org.jboss.pnc.common.logging.MDCUtils;
+
+import java.util.Map;
 
 /**
  * @author <a href="mailto:matejonnet@gmail.com">Matej Lazar</a>
  */
 @Data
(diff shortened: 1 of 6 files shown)


rest-api/src/main/java/org/jboss/pnc/client/Configuration.java
```diff
@@ -55,4 +77,67 @@ public BasicAuth(String username, String password) {
         }
     }
 
+    public static final class ConfigurationBuilder {
```
janinko: Do you know that you can combine custom with lombok generated builder?

rest-api/src/test/java/org/jboss/pnc/client/ClientTest.java
```diff
@@ -34,30 +38,44 @@
...
+        MDC.put(MDCUtils.REQUEST_CONTEXT_KEY, requestContext);
         try {
             projectClient.getSpecific("1");
         } catch (javax.ws.rs.ProcessingException | ClientException e) {
             //expected
         }
 
         Assert.assertTrue(requestsReceived.intValue() > 2);
+        Assert.assertEquals(requestContext, headerReceived.get());
+        System.out.println("Done!");
```
janinko: Done!

rest-api/src/main/java/org/jboss/pnc/client/Configuration.java
matejonnet: Forgot about that, thanks updating it.

PR #2771, opened 2019-12-10
NCL-5351 - Filter subquery for the same BuildConfigurationAudited.idRev

vibe13:
### Checklist:


vibe13:
@thescouser89 @jbartece please review. Thanks!

vibe13:
Thanks @thescouser89 foe the review, cool. If you mean the indexes on the tables used by the queries above, yes we already have all the indexes in DB for those. I double checked with a EXPLAIN ANALYZE query on the generated Hibernate sql (linked in the JIRA), which uses only Index scan (good). It's that what you mean? Thanks!

thescouser89:
@vibe13 yup! COOL COOL THEN! As always you already prepared for everything :)

vibe13:
@jbartece ok thanks! I have updated the javadoc now :-)

PR #2772, opened 2019-12-10
WIP [v1.8] NCL-5276 - implicit rebuild checks optimization

vibe13:
So working on https://github.com/project-ncl/pnc/pull/2771, I noticed that the "record.getDependencies()" call generates one single query, while the ".map(Artifact::getBuildRecords)" generates one query per dependency. So if there is a list of 1000 dependencies in a BuildRecord, 1000 queries are generated. I kept the first query and then attempted to create a single query for the rest, using a JPA query to optimize as much as possible. I would need to test it and turn on the SQL logging though to check that everything is perfectly fine (I am just not 100% on the JPA query). Wdyt? This should improve (I believe a lot) the checks for the implicit rebuilds. Thanks!


thescouser89:
retest this please

vibe13:
retest this please

vibe13:
retest this please

vibe13:
retest this please

vibe13:
retest this please

PR #2766, opened 2019-12-09
WIP - Spike for first configuration of Infinispan caches

vibe13:
### Checklist:

* [ ] Have you added a note in the [CHANGELOG wiki](https://github.com/project-ncl/pnc/wiki/Changelog) for your change if user-facing?
* [ ] Have you added unit tests for your change?

PR #2778, opened 2019-12-13
[v1.8] NCL-5276 implicit rebuild optimization 2

vibe13:
@jbartece this is an extension of the work done in https://github.com/project-ncl/pnc/pull/2777, to create a cache to avoid duplicated checks. As we are processing dependencies, it could be very common that in a set of BuildRecords to be verified during an implicit rebuild, the same dependencies are present (e.g. parent upstream poms). This cache aims to avoid making the same queries for the same dependencies. Thanks for the review!

### DIFF
--- a/build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildTasksInitializer.java
+++ b/build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildTasksInitializer.java
@@ -98,6 +98,8 @@ public class BuildTasksInitializer {
 
         toBuild.add(buildConfigurationAudited);
         if (buildOptions.isBuildDependencies()) {
+
+            Set<Integer> processedDependenciesCache = new HashSet<Integer>();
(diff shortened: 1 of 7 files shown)


thescouser89:
Change makes sense and it's a :+1:  from me :)

thescouser89:
@vibe13 BTW, make sure you add those optimizations to the changelog wiki. I think the users will love this :)

vibe13:
@thescouser89 thanks for the review and the reminder, added the changes in the wiki (and added some others fixes / improvements which were missing from the list)! :-)

vibe13:
@jbartece @thescouser89  ok thanks for the review and comment! I definitely agree! So wrt the hibernate caching, I am making some experiments already in https://github.com/project-ncl/pnc/pull/2766 :-). WIP as I am trying to configure tests atm. So, if you agree, I could merge this one for the moment and then when we decide what to do with Hibernate caching (and if the above spike could look good), I can revert to use it and remove this in memory optimization. Wdyt? Thanks!

jbartece:
Sure. Merging

PR #2793, opened 2020-01-07
[NCL-5012] Remove eager fetch in ProductMilestone

janinko:
### Checklist:


janinko:
The getPerformedBuilds() is used only in old rest and getDistributedArtifacts() is not used at all.

PR #2809, opened 2020-01-24
Remove `<str>.isBlank()` method call

thescouser89:
Replace it with:
```
<str>.trim().isEmpty()
```

Reasoning:

`isBlank` String method is only available in Java 11, and since we're
still targetting Java 1.8, this method shouldn't be used.

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#isBlank()
### Checklist:


janinko:
retest this please

thescouser89:
No idea! The Java 11 compiler is probably a bit permissive when using Java 11 API with source 1.8.

PR #2812, opened 2020-01-27
[NCL-5413] Serialize also null values to JSON

janinko:
In REST we serialize also NULL values, in WS we use object mapper from
common and it was not ommiting fields with NULL values.

### Checklist:


janinko:
retest this please

PR #2823, opened 2020-01-30
Ncl 5400 - Enable the second level cache

vibe13:
### Checklist:

### DIFF
--- a/datastore/src/main/resources/META-INF/persistence.xml
+++ b/datastore/src/main/resources/META-INF/persistence.xml
@@ -26,6 +26,7 @@
    <persistence-unit name="primary">
        <jta-data-source>@persistence.jta-data-source@</jta-data-source>
        <jar-file>model.jar</jar-file>
+       <shared-cache-mode>ENABLE_SELECTIVE</shared-cache-mode>
        <properties>
          <!-- Properties for Hibernate -->
           <property name="hibernate.dialect" value="@persistence.hibernate.dialect@"/>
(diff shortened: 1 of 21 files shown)


vibe13:
retest this please

vibe13:
retest this please

vibe13:
retest this please

vibe13:
@thescouser89 @jbartece This is the first configuration to cache all entities and collections. With some additional optimizations (batching). It would be great if you could do a review, and start seeing some changes in the grafana dashboards :-). Thanks!

thescouser89:
Unfortunately I don't know much about hibernate cache, but I see the changes are mostly configuration and adding cache annotations. So +1 for me.

datastore/src/main/resources/META-INF/persistence.xml
```diff
@@ -37,6 +38,43 @@
...
+          <property name="hibernate.batch_versioned_data" value="true"/>
+
+          <!-- Raise error instead of warning if “HHH000104: firstResult/maxResults specified with collection fetch; applying in memory!” is detected (since Hibernate 5.2.13) -->
+          <property name="hibernate.query.fail_on_pagination_over_collection_fetch" value="true"/>
+
+          <!--
+              Infinispan second level cache (default Wildfly settings are
+              https://infinispan.org/docs/stable/titles/integrating/integrating.html#default_local_configuration_second_level):
+                - Eviction wake up interval is 5 seconds.
+                - Max number of entries are 10,000.
```
jbartece: It would be good to monitor the size of the cache. My feeling is that higher number might be better for us as we have milions of artifacts and my feeling is that we might fill the cache quite quickly. I'd consider adding a metrics for it and increasing the cache for artifacts.

pom.xml
```diff
@@ -131,6 +131,7 @@
     <version.jackson.databind>2.10.1</version.jackson.databind>
     <version.org.jboss.resteasy>3.6.1.SP5</version.org.jboss.resteasy>
     <version.hibernate>5.3.10.Final</version.hibernate>
+    <version.org.infinispan>9.3.6.Final</version.org.infinispan>
```
jbartece: Shouldn't we use Infinispan version bundled with EAP?

datastore/src/main/resources/META-INF/persistence.xml
vibe13: Yes, that is the purpose of the REST endpoints implemented in https://projects.engineering.redhat.com/browse/NCL-5408. They will expose (once 2LC is enabled) all those information, along with the total size of the cache in memory. I have also progressed in displaying many of the in Grafana (https://projects.engineering.redhat.com/browse/NCL-5417), by enhancing the pnc-status service (https://projects.engineering.redhat.com/browse/NCL-5411), so this is all working. The last dashboard will display the stats for the 2nd level cache as per above (rest endpoints now returns no value as 2nd level cache is not enable yet).

pom.xml
vibe13: Yes, that is only required for the tests in the model module which use a RESOURCE_LOCAL persistence unit, and that seems to need an additional artifact. Should work out of the box for the standard persistence unit.

datastore/src/main/resources/META-INF/persistence.xml
jbartece: @vibe13 Great! :-)

pom.xml
jbartece: Isn't it enough to use the version of infinispan defined in EAP BOM?

vibe13: that dependency is not in the EAP BOM

jbartece: I found it there. I think it should be removed. See: https://mvnrepository.com/artifact/org.jboss.bom/eap-runtime-artifacts/7.2.6.GA
and
https://maven.repository.redhat.com/ga/org/jboss/bom/eap-runtime-artifacts/7.2.6.GA/eap-runtime-artifacts-7.2.6.GA.pom

jbartece: I'd add the runtime BOM.

jbartece: I didn't realize we don't use the runtime-artifacts BOM. Not sure it is worth adding it just for this one artifact for tests. 
Let's leave it as is

PR #2834, opened 2020-02-05
[NCL-5386] - Implementation of native query to retrieve minimized list of artifact…

vibe13:
… dependencies given a build record

### Checklist:

### DIFF
--- a/datastore/src/main/java/org/jboss/pnc/datastore/repositories/internal/ArtifactSpringRepository.java
+++ b/datastore/src/main/java/org/jboss/pnc/datastore/repositories/internal/ArtifactSpringRepository.java
@@ -76,4 +76,50 @@ public interface ArtifactSpringRepository extends JpaRepository<Artifact, Intege
             nativeQuery = true)
     Object[] countMinimizedDependencyArtifactsForBuildRecord(Integer buildRecordId);
 
+    @Query(
+            value = "SELECT DISTINCT " +
+                    " artifact.id, " +
+                    " artifact.artifactQuality, " +
+                    " artifact.deployPath, " +
+                    " artifact.filename, " +
+                    " artifact.identifier, " +
+                    " artifact.importDate, " +
+                    " artifact.md5, " +
+                    " artifact.originUrl, " +
+                    " artifact.sha1, " +
+                    " artifact.sha256, " +
+                    " artifact.size, " +
+                    " artifact.targetRepository_id as targetRepositoryId, " +
+                    " targetRepository.temporaryRepo, " +
+                    " targetRepository.identifier as targetRepositoryIdentifier, " +
+                    " targetRepository.repositoryPath, " +
+                    " targetRepository.repositoryType " +
+                    "FROM Artifact artifact " +
+                    "INNER JOIN build_record_built_artifact_map built_artifact " +
+                    "  ON built_artifact.built_artifact_id = artifact.id " +
+                    "INNER JOIN BuildRecord buildRecord " +
+                    "  ON built_artifact.build_record_id = buildRecord.id " +
+                    "INNER JOIN TargetRepository targetRepository " +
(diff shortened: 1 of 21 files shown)


vibe13:
retest this please

vibe13:
retest this please

vibe13:
retest this please

vibe13:
retest this please

vibe13:
@thescouser89 @jbartece Hi could you please review again this one? I could find another way to override at runtime some EAGER fetching (see https://projects.engineering.redhat.com/browse/NCL-5376), so I created an ad-hoc endpoint that calls a native query. Thanks!

vibe13:
retest this please

vibe13:
retest this please

rest-model/src/main/java/org/jboss/pnc/rest/restmodel/TargetRepositoryRest.java
```diff
@@ -78,6 +78,15 @@
     @NotNull(groups = {WhenUpdating.class, WhenCreatingNew.class})
     private Set<Integer> artifactIds = new HashSet<>();
 
+    public TargetRepositoryRest(Integer id, Boolean temporaryRepo,
```
jbartece: I think, it is redundant as we have a Builder for this class.

datastore/src/main/java/org/jboss/pnc/datastore/repositories/internal/ArtifactSpringRepository.java
```diff
@@ -18,12 +18,108 @@
...
+                    "FROM Artifact artifact " +
+                    "INNER JOIN build_record_built_artifact_map built_artifact " +
+                    "  ON built_artifact.built_artifact_id = artifact.id " +
+                    "INNER JOIN BuildRecord buildRecord " +
+                    "  ON built_artifact.build_record_id = buildRecord.id " +
+                    "INNER JOIN TargetRepository targetRepository " +
+                    "  ON targetRepository.id = artifact.targetRepository_id " +
+                    "WHERE " +
+                    " buildRecord.id = ?1",
+            nativeQuery = true)
```
jbartece: What is the reason to set it as nativeQuery?

vibe13:
> What are the reasons behind this PR? To improve performance by not loading dependent entities?
@jbartece yes as mentioned in the call earlier it's because the number of joins that gets triggered from the rest endpoints. Unfortunately with the current model, I could not override at runtime the EAGER fetching, so they query is extremely slow and returns data which are not needed by the Maven Generation zip.

> What is the reason behind the RawArtifact? Is it not to contain the deps?
Yes it does not contain any deps from Artifact to BuildRecords. Using native queries, it's more convenient to have an interface which can automatically map the Object[] returned to something meaningful, rather than doing that mapping yourself. I could not use the ArtifactRest class as rest-model is not meant to be visible from that package. And did not want to use Artifact entity class because the results of a native query get managed by Hibernate if mapped to entities, and did not want to incur in the problem of wiping out the collections of the returned Artifacts :-)
 
> Do you plan it to port it to 2.0 as well?
First I will check what kind of queries get generated from that REST endpoints, and would rather fix the mapping if it's the case. I would not like to port native queries to 2.0

rest-model/src/main/java/org/jboss/pnc/rest/restmodel/TargetRepositoryRest.java
vibe13: Fair point, will fix that!

datastore/src/main/java/org/jboss/pnc/datastore/repositories/internal/ArtifactSpringRepository.java
vibe13: With standard JPA I could not mange to override at runtime some EAGER fetches, so I had to go to native queries.

vibe13:
retest this please

PR #2863, opened 2020-02-13
Build output checksum

matejonnet:
### Checklist:

### DIFF
--- a/rest-api/src/main/java/org/jboss/pnc/client/RemoteCollection.java
+++ b/rest-api/src/main/java/org/jboss/pnc/client/RemoteCollection.java
@@ -17,9 +17,36 @@
  */
 package org.jboss.pnc.client;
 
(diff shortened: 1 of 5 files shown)


rest-api/src/main/java/org/jboss/pnc/client/RemoteCollection.java
```diff
@@ -17,9 +17,36 @@
...
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
 /**
  * @author <a href="mailto:matejonnet@gmail.com">Matej Lazar</a>
  */
 public interface RemoteCollection<T> extends Iterable<T> {
     int size();
+
+    static<T> RemoteCollection<T> empty() {
```
matejonnet: @jbartece FYI

PR #2877, opened 2020-02-25
[v1.8][NCL-5520] Fix thread MDC context

thescouser89:
MDC works by setting a context map for each thread [1]. While this works quite
well, we have to be careful when using a threadpool, since threads get re-used.
This means that the context of a thread, set by a previous job, could be
erroneously used with a new job.

We are concerned with 3 main classes in this bug,
`DefaultBuildCoordinator`,`BpmBuildScheduler` and `DefaultBuildExecutor`.

This class handles the 'pod creation', Maven/Gradle building, and the
collection of results. It interacts with the other 2 classes indirectly via
Maitai. The MDC from Maitai -> BuildExecutor is set. `DefaultBuildExecutor`
uses a modified thread pool (`MDCThreadPoolExecutor`) that explicitly sets the
MDC for each thread processing a `Runnable`, `Callable` etc. As such, this
class properly deals with MDC in the threadpool. The MDC is originally set via
the REST call from Maitai.

It coordinates the whole build process. It also uses the
`MDCThreadPoolExecutor`. However, the issue is that the methods from this class
may be called from threads not part of the MDC aware executor (like the
`cancel` call chain). This causes bugs that this commit attempts to fix.

This class is called from `DefaultBuildCoordinator`, where it invokes the
Maitai process. When a build is done/cancelled, the method
`buildCoordinator@completeBuild` is called via a runnable from the
`BpmBuildScheduler`. As such, it is important to also set the MDC context in
that method also.

The fix involves to set the MDC context based on the build task, for all the
methods in the `DefaultBuildCoordinator`. While I recognize that this is
probably unnecessary for most cases, I wanted to cover potential corner cases
in the future.

The primary fix is for the issue described in NCL-5520, where the 'enqueued'
and 'waiting for dependencies' states are not reported to Kafka due to the
thread where the job is running having the wrong MDC context.

[1]: http://logback.qos.ch/manual/mdc.html

### Checklist:

### DIFF
--- a/build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/DefaultBuildCoordinator.java
+++ b/build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/DefaultBuildCoordinator.java
@@ -320,9 +320,16 @@ public class DefaultBuildCoordinator implements BuildCoordinator {
         } else {
             updateBuildTaskStatus(buildTask, BuildCoordinationStatus.WAITING_FOR_DEPENDENCIES);
             Runnable onTaskReady = () -> {
+
+                // we need to add this because we need to re-put those metadata in the thread context since the
+                // current thread context probably belongs to another build
+                MDCUtils.addContext(getMDCMeta(buildTask));
+
                 ProcessStageUtils.logProcessStageEnd(BuildCoordinationStatus.WAITING_FOR_DEPENDENCIES.toString());
                 updateBuildTaskStatus(buildTask, BuildCoordinationStatus.ENQUEUED);
                 ProcessStageUtils.logProcessStageBegin(BuildCoordinationStatus.ENQUEUED.toString());
+
+                MDCUtils.clear();
             };
             buildQueue.addWaitingTask(buildTask, onTaskReady);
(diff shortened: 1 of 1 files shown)


jbartece:
I'm not sure if it covers all the cases, but using the current approach it looks good. @janinko 's question about the EAP managed executors is a good one though. I'm not sure why we do not use the EAP ones.

thescouser89:
I believe we use Java CE executors (and modifying it to be MDC aware) to be able to control the number of builds by tweaking the number of threads in the executor.

janinko:
Can't we use the EAP managed executors inside our executors that will only tweak the number ... somehow?

jbartece:
The bumber of builds si limited by a variable in DefaultBuildCoordinator. It is not a thread per build AFAIK. The threadpool just serves the requests

thescouser89:
I'm afraid that changing to use EAP managed executors will require more changes than what we want to do at the closing stages of 1.8 release. 

https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/7.0/html/development_guide/concurrency_utilities

I guess you are right about how the threads work. There are so many thread pool size configurations that I'm confused on which one does what.

However I'd be wary of doing such a change to be honest. We'll have to test for the new behaviour.

If you guys want to go in that direction, could any of you take care of it? I don't know anything about the EAP managed executors and the docs aren't that helpful?

PR #2890, opened 2020-03-02
Fix package of moved GenericSettingProvider

pkocandr:
GenericSettingProvider in facade/src/main/java/org/jboss/pnc/facade/providers/ declared package org.jboss.pnc.rest.provider which is obviously wrong. Eclipse refused to compile the class.


thescouser89:
How did that even happen!?

PR #2894, opened 2020-03-04
Remove CUSTOM_PROJECT_MANIPULATOR_PARAMETERS, endpoint returns JSON e…

dominikbrazdil:
…xcept of String

### Checklist:

### DIFF
--- a/dto/src/main/java/org/jboss/pnc/dto/response/AlignmentParameters.java
+++ b/dto/src/main/java/org/jboss/pnc/dto/response/AlignmentParameters.java
@@ -1,0 +1,43 @@
+/**
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014-2019 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.pnc.dto.response;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+
+/**
+ * Default parameters for build configuration for a specific build type
+ *
+ * @author dbrazdil
+ */
+@Getter
+@AllArgsConstructor
+@Builder(builderClassName = "Builder")
+@JsonDeserialize(builder = AlignmentParameters.Builder.class)
+public class AlignmentParameters {
+    public final String buildType;
+
+    public final String parameters;
+
+    @JsonPOJOBuilder(withPrefix = "")
+    public static final class Builder {
+    }
+}
--- a/integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildConfigurationEndpointTest.java
+++ b/integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildConfigurationEndpointTest.java
@@ -42,6 +42,7 @@ import org.jboss.pnc.dto.Environment;
 import org.jboss.pnc.dto.GroupConfiguration;
 import org.jboss.pnc.dto.ProjectRef;
 import org.jboss.pnc.dto.SCMRepository;
+import org.jboss.pnc.dto.response.AlignmentParameters;
 import org.jboss.pnc.dto.response.Parameter;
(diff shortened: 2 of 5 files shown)


jbartece:
@janinko @alexcreasy  What is your opinion on that?

dto/src/main/java/org/jboss/pnc/dto/response/AlignmentParameters.java
```diff
@@ -0,0 +1,44 @@
...
+package org.jboss.pnc.dto.response;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Getter;
+
+/**
+ * // Default parameters for build configuration for a specific build type
```
michalovjan: "//" could be deleted

dto/src/main/java/org/jboss/pnc/dto/response/AlignmentParameters.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @author dbrazdil
+ * @since 4.0
```
michalovjan: From where did Intellij get this number?

jbartece:
@michalovjan I don't think it makes nicer code - it just complicates it. Another thing is that more important is, if it makes sense from the API perspective. I don't think so.

dto/src/main/java/org/jboss/pnc/dto/response/AlignmentParameters.java
jbartece: I am curious about that as well :-D

integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildConfigurationEndpointTest.java
```diff
@@ -482,9 +483,9 @@ public void testGetBuildTypeDefaultAlignmentParameters() throws RemoteResourceEx
         BuildConfigurationClient client = new BuildConfigurationClient(RestClientConfiguration.asAnonymous());
 
         for (BuildType buildType : BuildType.values()) {
-            Optional<String> params = client.getBuildTypeDefaultAlignmentParameters(buildType.name());
-            assertThat(params).isPresent();
-            assertThat(!params.get().isEmpty());
+            AlignmentParameters params = client.getBuildTypeDefaultAlignmentParameters(buildType.name());
+            assertThat(!params.getParameters().isEmpty());
+            assertThat(!params.getBuildType().isEmpty());
```
janinko: I think it should be like this?
```suggestion
            assertThat(params.getParameters()).isNotEmpty();
            assertThat(params.getBuildType()).isNotEmpty();
```

janinko:
@jbartece we should be returning JSON objects in responses.

Returning the `buildType` is what I advised Dominik as I thought it would make the returned object self-describing without the need to remember the request parameter. But I don't have strong opinion on including that parameter.

jbartece:
@janinko Why do you think we should return JSON here? E.g, for build and alignment logs we return StreamingOutput.
If you all think it is better, then OK. I'm just curious why do you think it is better.

alexcreasy:
@jbartece I admit it looks a little weird, but I agree with Honza, IMO the primary wire format of our API is JSON so we should return responses in JSON unless there is a good reason not to (e.g. a whole log file). 

The problem is, if you switch the method return type to a String, the content-type has to change from application/json to text/plain, I assume RESTEasy will automagically fix the headers for you.

In general it just places a burden on the client to remember there are different content types for different endpoints, the UI has quite a lot of interceptors and automatic code that assume JSON so I have to put some manual exceptions in for text/plain.

alexcreasy:
Just to add that for a log file, it's a large stream of plain text, it would be highly expensive for the client to parse that as JSON just to reconvert it back to plain/text... The client really wants to put that in a buffer and handle it as it comes. 

For structured data JSON is better since it just gets parsed, I can grab the String from the message easily as a string and work with it like any other variable.

jbartece:
OK, let's make it application/json then.

michalovjan:
> @michalovjan I don't think it makes nicer code - it just complicates it. Another thing is that more important is, if it makes sense from the API perspective. I don't think so.
> I propose to close this PR and just remove the CUSTOM_PROJECT_MANIPULATOR_PARAMETERS property.

@jbartece I was referencing to the changes in my comment that just make the code nicer, sorry for the confusion.

integration-test/src/test/java/org/jboss/pnc/integration_new/endpoint/BuildConfigurationEndpointTest.java
dominikbrazdil: I can't find anything similar to String.isNotEmpty(), hamcrest supports something like this assertThat(string, is(not(emptyString()))) but that does not seem any better to me

jbartece: You must call it on the object returned by assertThat not String.
E.g.: assertThat("test").isNotEmpty();

dto/src/main/java/org/jboss/pnc/dto/response/AlignmentParameters.java
dominikbrazdil: no idea, when does intellij add these annotations?

PR #2895, opened 2020-03-04
[NCL-5483] Add websocket support to rest-client

michalovjan:
### Checklist:

### DIFF
--- a/rest-client/src/main/java/org/jboss/pnc/restclient/websocket/VertxWebSocketClient.java
+++ b/rest-client/src/main/java/org/jboss/pnc/restclient/websocket/VertxWebSocketClient.java
@@ -46,9 +46,11 @@ import io.vertx.core.http.WebSocket;
 /**
  * @author <a href="mailto:jmichalo@redhat.com">Jan Michalov</a>
  */
-public class VertxWebSocketClient implements WebSocketClient {
+public class VertxWebSocketClient implements WebSocketClient, AutoCloseable {
 
     private static final Logger log = LoggerFactory.getLogger(VertxWebSocketClient.class);
+
+    private static final ObjectMapper objectMapper = JsonOutputConverterMapper.getMapper();
 
     private final Vertx vertx;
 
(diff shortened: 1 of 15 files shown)


rest-client/src/main/java/org/jboss/pnc/restclient/websocket/VertxWebSocketClient.java
```diff
@@ -0,0 +1,317 @@
...
+     * number of attempted retries before client throws an exception
+     */
+    private int maximumRetries = 5;
+
+    private int numberOfRetries = 0;
+
+    /**
+     * a multiplier that increases delay between reconnect attempts
+     */
+    private float delayMultiplier = 1.5F;
```
thescouser89: I'd say just use as delayMultiplier the number '2'.

And with that you can optimize by doing a bitwise shift left to "multiply" by 2.

thescouser89:
LGTM with my 1 tiny comment. With the CompletableFuture, I can chain callbacks after the results are obtained.

rest-client/src/main/java/org/jboss/pnc/restclient/websocket/VertxWebSocketClient.java
michalovjan: For sure can change it to 2. With bitwise shift it forces the number to be 2 and doesn't allow for the number to change (currently can be set through constructor). Do we want to hard code it?

michalovjan:
> LGTM with my 1 tiny comment. With the CompletableFuture, I can chain callbacks after the results are obtained.
> 
> Could you add a method for getting the live logs of a build? .

Oh yeah! Live logs! Completely forgot about them :D. 

> Same for creation of build configuration, and scm repository

Do I understand this correctly? To send every processed message/notification also to stdout? :)

matejonnet:
Nice!

rest-client/src/main/java/org/jboss/pnc/restclient/websocket/VertxWebSocketClient.java
janinko: I once heard some thing about premature optimization. Doesn't it apply here?

alexcreasy: errm, changing from 1.5 to 2 is a pretty big increase in backoff factor... isn't 1.3 usually the standard?

alexcreasy: actually I don't know where I got 1.3 from, my brain offered it up, but I can't find anything to back it up...

thescouser89:
+1 to merge it! :)

rest-client/src/main/java/org/jboss/pnc/restclient/websocket/VertxWebSocketClient.java
michalovjan: I've made that number up, so I am free to change it to whatever you want. We can also make it customizable through CLI config or smthing like that.

PR #2896, opened 2020-03-04
[v1.8][NCL-5528] websocket for generic settings

thescouser89:
### Checklist:

* [x] Have you added a note in the [CHANGELOG wiki](https://github.com/project-ncl/pnc/wiki/Changelog) for your change if user-facing?
* [ ] Have you added unit tests for your change?

PR #2897, opened 2020-03-04
[NCL-5528] websocket for generic settings

thescouser89:
### Checklist:


thescouser89:
Similar to #2896

thescouser89:
@alexcreasy This one (for 2.0) is re-using the only websocket notifications endpoint that exists. The 1.8 one is not re-using any notifications endpoint because they are specific to either the build executor, or the build records.

alexcreasy:
@thescouser89 ahhh lol, I just followed the link from the first email that was the v1.8 PR then saw this one and didn't think to check for differences.

Yep this one looks fine and for 1.8 it doesn't matter, I'll change the review, all good from my end :D

thescouser89:
Thanks for reviewing guys! :heart:

janinko:
too late, but :100:

PR #2898, opened 2020-03-05
NCL-5587 Fix bug: Link on Build Tree does not work properly

DnsZhou:
### Checklist:

### DIFF
--- a/ui/app/builds/directives/pnc-build-tree/pnc-build-tree-link.html
+++ b/ui/app/builds/directives/pnc-build-tree/pnc-build-tree-link.html
@@ -20,11 +20,15 @@
 <div>
   <pnc-build-status-icon build="$ctrl.buildItem"></pnc-build-status-icon>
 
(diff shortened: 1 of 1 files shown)


ui/app/builds/directives/pnc-build-tree/pnc-build-tree-link.html
```diff
@@ -20,11 +20,15 @@
 <div>
   <pnc-build-status-icon build="$ctrl.buildItem"></pnc-build-status-icon>
 
-  <a ng-if="$ctrl.build" ui-sref="projects.detail.build-configs.detail({ projectId: $ctrl.build.buildConfigRevision.project.id, configurationId: $ctrl.build.buildConfigRevision.id })" ng-click="$event.stopPropagation()">{{$ctrl.build.buildConfigRevision.name}}</a> 
+  <a ng-if="$ctrl.build"
+     ui-sref="projects.detail.build-configs.detail({ configurationId: $ctrl.build.buildConfigRevision.id })"
```
alexcreasy: A projectId is required to link to a configurationId. This only works here because of state inheritance, the projectId is already set in this state. It would be better to explicity add it in case this component is ever reused in another state, where the link would silently break... the projectId is available on the build object so adding projectId: $ctrl.build.project.id should work here.

matedo1:
Once Alex's changes are done I think we are ready to merge it

DnsZhou:
Ok, will update it accordingly, Thank you both! @matedo1 @alexcreasy

PR #2910, opened 2020-03-11
NCL-5482 Create UI features to support backend maintenance mode features - 2.0

DnsZhou:
### Checklist:


ui/app/common/components/pnc-admin-panel/pncAdminPanel.js
```diff
@@ -0,0 +1,134 @@
...
+  'use strict';
+
+  /**
+   * The component for Admin panel for generic settings
+   */
+  angular.module('pnc.common.components').component('pncAdminPanel', {
+    bindings: {
+    },
+    templateUrl: 'common/components/pnc-admin-panel/pnc-admin-panel.html',
+    transclude: true,
```
alexcreasy: Does the admin panel need to transclude any content? If not then it's better to remove this!

ui/app/common/pnc-client/message-bus/messageBus.js
```diff
@@ -50,6 +50,20 @@
...
 
+      function newGenericSettingProgressEventHandler(eventType) {
+        $rootScope.$broadcast(eventType);
+      }
+
+      messageBus.onGenericSettingMaintenanceChanged(notification => {
+        if (notification.message) {
+          let message = JSON.parse(notification.message);
+          if (message.maintenanceModeEnabled) {
+            newGenericSettingProgressEventHandler(events.MAINTENANCE_MODE_ON);
```
alexcreasy: I think here you can just invoke $rootScope.$brodcast(events.MAINTENANCE_MODE_  directly in the if / else clauses. The newGenericSettingProgressEventHandler doesn't actually seem to do anything.

That pattern is used for builds / group builds because they share some common behaviour. It isn't necessary here since it's just an abstraction over a single method call.

PR #2914, opened 2020-03-13
[NCL-5490] Create scm repository from bpm notification

thescouser89:
Instead of BPM creating the scm repository as part of its process, BPM
should now only report the internal url in its success notification.
From there on, PNC-Orch will use the internal url to create the scm
repository in its database.

This should simplify the BPM process for repository configuration.

### Checklist:


thescouser89:
Just not ready to be merged yet! Needs to do some maitai changes

PR #2925, opened 2020-03-20
[v1.8] NCL-5527 - Adding / Updating a description on build configurat…

vibe13:
…ion causes a rebuild

### Checklist:

### DIFF
--- a/build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildTasksInitializer.java
+++ b/build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/BuildTasksInitializer.java
@@ -19,7 +19,6 @@
 package org.jboss.pnc.coordinator.builder;
 
 import org.jboss.pnc.common.mdc.MDCUtils;
-import org.jboss.pnc.common.util.ProcessStageUtils;
 import org.jboss.pnc.coordinator.builder.datastore.DatastoreAdapter;
 import org.jboss.pnc.model.BuildConfigSetRecord;
 import org.jboss.pnc.model.BuildConfiguration;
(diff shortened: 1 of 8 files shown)


vibe13:
@janinko @jbartece @thescouser89 Hi! I have removed the description from the audited fields inside the BuildConfiguration. The name has to remain there (I have tried to remove that as well, but it brings in a lot of issues, so it's better to keep it where it is).
I have not changed the definition of the SPI https://github.com/project-ncl/pnc/blob/dd79fba82b265b03c47b624327b1786a5b3779c5/spi/src/main/java/org/jboss/pnc/spi/dto/BuildConfigurationRevisionRef.java, because I don't know where it's used at the moment. Maybe you guys know? I guess we can leave it as it is for now, and let the description stays there to avoid breaking any relation with other services? And we can remove it in PNC 2.0 branch, where the DTO will be shared with other services? Thanks!

thescouser89:
Yeah I think that's a sane decision! Maitai doesn't use the audited data. Same for repour. I think there's a 90% chance we can remove it in the RevisionRef without any issues, but then we'll break the API response I think for v1.8.

So it makes sense for me.

vibe13:
@jbartece @janinko @matejonnet If a field is marked as NotAudited, it won't be stored in the Audited Entity at all, so in this case the description would not be present inside the BuildConfigurationAudited anymore. This is the reason why I removed it entirely from the POJO and DTOs. 
In case the description was an entity, using the annotation @Audited(targetAuditMode = RelationTargetAuditMode.NOT_AUDITED) would have the effect you ask for, that is still storing the description inside the Audited table, without caring about tracking changes of it (== do not generate a new revision if description is changed). However, @Audited(targetAuditMode = RelationTargetAuditMode.NOT_AUDITED) can be used only on entities (we use it already on some relationships inside the BuildConfiguration) but not on fields.

janinko:
Not sure if we shouldn't left the field in DTO defaulting to null. It's an API change otherwise and I know we don't care about API compatibility in 1.x but I think we shouldn't break API in micro.

There will be PR for 2.0 later?

vibe13:
> Not sure if we shouldn't left the field in DTO defaulting to null. It's an API change otherwise and I know we don't care about API compatibility in 1.x but I think we shouldn't break API in micro.
> 
Good point. I will leave the BuildConfigurationAuditedRest as it was before, with null values for description then.

> There will be PR for 2.0 later?

Yep as soon as we agree on the above changes! :-)

vibe13:
retest this please

vibe13:
retest this please

PR #2943, opened 2020-03-27
[v1.8] NCL-5527 - Update lastModificationTime only if audited fields are updated

vibe13:
### Checklist:

### DIFF
--- a/datastore/src/main/java/org/jboss/pnc/datastore/repositories/BuildConfigurationRepositoryImpl.java
+++ b/datastore/src/main/java/org/jboss/pnc/datastore/repositories/BuildConfigurationRepositoryImpl.java
@@ -20,6 +20,7 @@ package org.jboss.pnc.datastore.repositories;
 import org.jboss.pnc.datastore.repositories.internal.AbstractRepository;
 import org.jboss.pnc.datastore.repositories.internal.BuildConfigurationSpringRepository;
 import org.jboss.pnc.model.BuildConfiguration;
+import org.jboss.pnc.model.GenericEntity;
 import org.jboss.pnc.spi.datastore.repositories.BuildConfigurationRepository;
 
 import javax.ejb.Stateless;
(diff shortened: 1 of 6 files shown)


datastore/src/main/java/org/jboss/pnc/datastore/repositories/BuildConfigurationRepositoryImpl.java
```diff
@@ -53,15 +54,51 @@ public BuildConfiguration save(BuildConfiguration buildConfiguration) {
...
+                equalsWithNull(persisted.getBuildScript(), toUpdate.getBuildScript()) &&
+                equalsId(persisted.getRepositoryConfiguration(), toUpdate.getRepositoryConfiguration()) &&
+                equalsWithNull(persisted.getScmRevision(), toUpdate.getScmRevision()) &&
+                equalsId(persisted.getProject(), toUpdate.getProject()) &&
+                equalsId(persisted.getBuildEnvironment(), toUpdate.getBuildEnvironment()) &&
+                (persisted.isArchived() == toUpdate.isArchived()) &&
+                (persisted.getBuildType() == toUpdate.getBuildType());
+    }
+
+    private boolean equalsWithNull(Object o1, Object o2) {
```
janinko: reinwenting wheel: https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object-

datastore/src/main/java/org/jboss/pnc/datastore/repositories/BuildConfigurationRepositoryImpl.java
```diff
@@ -53,15 +54,51 @@ public BuildConfiguration save(BuildConfiguration buildConfiguration) {
...
+            return true;
+        }
+
+        if (persisted == null && toUpdate != null) {
+            return false;
+        }
+
+        if (persisted != null && toUpdate == null) {
+            return false;
+        }
```
janinko: You can shorten it as:
```suggestion
        if (persisted == null || toUpdate == null) {
            return false;
        }
```
Not sure which one is better. For mine you need to consider previous guard, yours is more explicit.

integration-test/src/test/java/org/jboss/pnc/integration/BuildTest.java
```diff
@@ -264,25 +270,103 @@ public void shouldBuildTemporaryBuildAndNotAssignItToMilestone() {
...
+        RestResponse<BuildRecordRest> forcedPersistentBuild = triggerBCBuild(buildConfigurationRest, Optional.empty(), persistent);
+        assertThat(forcedPersistentBuild.getRestCallResponse().getStatusCode()).isEqualTo(200);
+        ResponseUtils.waitSynchronouslyFor(() -> {
+            RestResponse<BuildRecordRest> record = buildRecordRestClient.get(forcedPersistentBuild.getValue().getId(),
+                    false);
+            return record.hasValue() && record.getValue().getStatus().equals(BuildCoordinationStatus.DONE);
+        } , 15, TimeUnit.SECONDS);
+
+        // Update only description, should not create a new revision and keep same lastModificationTime
+        String oldDescription = buildConfigurationRest.getDescription() != null ? new String(buildConfigurationRest.getDescription()) : null;
```
janinko: why do you call `new String` here?

janinko: Why isn't this enough? Strings are immutable
```suggestion
        String oldDescription = buildConfigurationRest.getDescription();
```

integration-test/src/test/java/org/jboss/pnc/integration/BuildTest.java
```diff
@@ -264,25 +270,103 @@ public void shouldBuildTemporaryBuildAndNotAssignItToMilestone() {
...
+        Date oldLastModDate = buildConfigurationRest.getLastModificationTime() != null ? new Date(buildConfigurationRest.getLastModificationTime().getTime()) : null;
```
janinko: The same here. I know Date is mutable, but is the lastModificationTime being altered?

datastore/src/main/java/org/jboss/pnc/datastore/repositories/BuildConfigurationRepositoryImpl.java
vibe13: Ah thanks, ok using existing wheel now :-), here and in another place where I first found it

datastore/src/main/java/org/jboss/pnc/datastore/repositories/BuildConfigurationRepositoryImpl.java
vibe13: I applied yours as it's better, here an in another method in same class

integration-test/src/test/java/org/jboss/pnc/integration/BuildTest.java
vibe13: They are enough, this is leftover of junk code I put when I was frustrated by tests not passing (due to other issues), which I did not remove. Thanks!

integration-test/src/test/java/org/jboss/pnc/integration/BuildTest.java
vibe13: Same as above

vibe13:
@janinko is it all good now? Thanks! :-)

PR #2952, opened 2020-04-01
[NCL-5597] fix running conditions in BuildCoordinator's tests

michalovjan:
InGroupDependentBuildsTest.shouldCreateTaskForDependentBuilt

There were 2 running conditions:

1. concurrent modification of mocked BuildRecordRepository (circa 1/100 probability)

- repository was returning direct reference to the data collection
- a method looping through the collection (in a stream) had the collection modified in the middle of loop resulting in a ConcurrentModificationException thrown from the inside of the Collection
- tried to wrap the collection in a new array so it wouldn't have direct reference but it caused some failures so I've decided to change the collection implementation to CopyOnWriteArray which is thread-safe (I haven't seen noticeable performance overhead)

2. successful BuildTask failing to notify dependant BuildTasks (circa 1/400 probability)

- the BuildTasks in a BuildTaskSet were added to the BuildQueue in a random order
- running condition happened when there was a BuildTask with dependants added to readyQueue and finished before its dependants to waitingQueue
- sometimes it happened that BuildCoordinator thread picked up the BuildTask with dependants and finished before any dependant was added to the waiting queue
- during finishing process there was no Task in waiting queue so the thread didn't notify any task
- fixed by ensuring that the order for adding task to queue is ordered from dependants to dependencies so that dependants are added to queue before dependencies

SkippingBuiltConfigsTest.shouldNotTriggerTheSameBuildConfigurationViaDependency

There were 2 running condition:

1. updateBuildTaskStatus method took too long to finish
- BuildTask had an dependency that was in BUILDING status before condition that places new BuildTasks in appropriate queue (waiting or ready), therefore it was decided to be placed in waiting queue
- race condition happened when a dependent Task finished before the aforementioned task was added to the waiting queue and after it was decided it should be placed in the waiting queue
- race condition resulted in stuck BuildTask in WAITING_FOR_DEPENDENCIES as the dependency could not not notify the BuildTask because waiting queue was at that time empty
- solved by short sleep in the test

2. BuildTask being placed in ready queue twice
- happened when attempting to add a Task to a queue when it was already there and an condition didn't catch it
- solved by ignoring BuildTasks that are not NEW or ENQUEUED when adding to BuildQueue

@matejonnet WDYT

### DIFF
--- a/build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/DefaultBuildCoordinator.java
+++ b/build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/DefaultBuildCoordinator.java
@@ -49,13 +49,12 @@ import org.jboss.pnc.spi.coordinator.ProcessException;
 import org.jboss.pnc.spi.coordinator.events.DefaultBuildSetStatusChangedEvent;
 import org.jboss.pnc.spi.coordinator.events.DefaultBuildStatusChangedEvent;
 import org.jboss.pnc.spi.datastore.DatastoreException;
-import org.jboss.pnc.spi.events.BuildStatusChangedEvent;
 import org.jboss.pnc.spi.events.BuildSetStatusChangedEvent;
+import org.jboss.pnc.spi.events.BuildStatusChangedEvent;
 import org.jboss.pnc.spi.exception.BuildConflictException;
 import org.jboss.pnc.spi.exception.CoreException;
 import org.jboss.pnc.spi.executor.exceptions.ExecutorException;
 import org.jboss.pnc.spi.repour.RepourResult;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -63,7 +62,6 @@ import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.event.Event;
 import javax.inject.Inject;
-
 import java.time.Instant;
 import java.util.Date;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


build-coordinator/src/main/java/org/jboss/pnc/coordinator/builder/DefaultBuildCoordinator.java
```diff
@@ -49,35 +68,15 @@
...
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.TimeUnit;
-import java.util.function.Consumer;
-
-import static org.jboss.pnc.common.util.CollectionUtils.hasCycle;
-
```
jbartece, line 80: Have you imported the ide-config to your IDE?

michalovjan: Ya!. I even 'mvn clean compile ...'  before committing so that auto-formatter formats the changes. But I don't think the formatter touches stuff as ordering imports. Something else is broken on my side.

michalovjan: Hmm... The Eclipse Formatter IntelliJ plugin was forcing Eclipse style imports ordering. I turned it off and it is better. Watch out if you are using it with 'optimize imports'.

PR #2985, opened 2020-04-20
Add an option to define nodeId for sequence generator.

matejonnet:
### Checklist:


common/src/main/java/org/jboss/pnc/common/concurrent/Sequence.java
```diff
@@ -20,13 +20,47 @@
...
+ *
+ * Epoch timestamp in millisecond precision - 41 bits (gives us 69 years with a custom epoch) Configured machine id - 10
+ * bits (gives us up to 1024 machines) Sequence number - 12 bits (A local counter per machine that rolls over every
+ * 4096)
+ */
 public class Sequence {
 
-    private static final Map<Integer, SequenceGenerator> instance = new ConcurrentHashMap<>();
-    private static final int INSTANCE_KEY = 0;
+    private static final Map<Integer, Sequence> instance = new ConcurrentHashMap<>();
```
janinko, line 34:
```suggestion
    private static final Map<Integer, Sequence> instance = new ConcurrentHashMap<>(1,1.0f);
```

common/src/main/java/org/jboss/pnc/common/concurrent/Sequence.java
```diff
@@ -20,13 +20,47 @@
...
+    }
 
     public static Long nextId() {
-        SequenceGenerator sequenceGenerator = instance.computeIfAbsent(INSTANCE_KEY, k -> new SequenceGenerator());
-        return sequenceGenerator.nextId();
+        return getInstance(null).sequenceGenerator.nextId();
+    }
+
+    private static Sequence getInstance(Integer nodeId) {
+        return instance.computeIfAbsent(0, k -> new Sequence(nodeId));
```
janinko, line 51: Make the `0` constant.

matejonnet: What the benefit of using the constant, it's used only here and it can be anything ?

janinko: 1) readability - giving it a name may be less confusing ("Why is there zero? Why not one?")
2) negligible speed - making it Integer constant removes the autoboxing of int.

gcode with autoboxing:
```
       3: iconst_0
       4: invokestatic  #3                  // Method java/lang/Integer.valueOf:(I)Ljava/lang/Integer;
```
gcode with constant:
```
       3: getstatic     #3                  // Field KEY:Ljava/lang/Integer;
```

PR #2989, opened 2020-04-21
[v1.8][NCLSUP-40] Fix logic for keycloak `refreshRequired`

thescouser89:
The logic for the `refreshRequired` method in
`DefaultKeycloakServiceClient` is the reverse of what it should be
returning.

It should be:
- If the token is expiring *before* now + the max duration of the build,
  then we need to renew

It seems to have accidentally gone unnoticed for many years because
builds happening frequently (in less than 1 hour of each other)
accidentally renews the token. This is generally the case in Production
orch but not the case in Stage.

Furthermore if we redeploy our Devel environment after every commit
(which happens frequently), this tends to hide the issue.

The default for 'serviceTokenRefreshIfExpiresInSeconds' is also changed
from 1 hour to 24 hours to reflect the max allowed time for a build to
happen.

### Checklist:

### DIFF
--- a/auth/src/main/java/org/jboss/pnc/auth/DefaultKeycloakServiceClient.java
+++ b/auth/src/main/java/org/jboss/pnc/auth/DefaultKeycloakServiceClient.java
@@ -65,9 +65,14 @@ public class DefaultKeycloakServiceClient implements KeycloakServiceClient {
     }
 
     private boolean refreshRequired() {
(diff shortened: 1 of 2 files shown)


vibe13:
@thescouser89 so I guess also that keycloakToken.getExpiresIn() will be higher than serviceTokenRefreshIfExpiresInSeconds seconds right? Otherwise the token will also be refreshed?

thescouser89:
@vibe13 Yup 'getExpiresIn' will return 2 days of validity, and serviceTokenRefreshIfExpiresInSeconds is set to 1 day.

thescouser89:
Thanks for the reviews! :)

PR #3007, opened 2020-04-28
[NCL-5507] fix ENQUEUED status being IN_PROGRESS instead of PENDING

michalovjan:
### Checklist:


dto/src/main/java/org/jboss/pnc/dto/notification/BuildChangedNotification.java
```diff
@@ -67,22 +63,28 @@
...
-            return IN_PROGRESS;
+        switch (status) {
+            case PENDING:
+                return PENDING;
+            case FINISHED:
+                return FINISHED;
+            case IN_PROGRESS:
+                return IN_PROGRESS;
+            default:
+                return null;
```
janinko, line 87: 😬

PR #3024, opened 2020-05-07
[NCL-5720] - Remove build execution non-final status updates

vibe13:
### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/executor/BuildExecutorTriggerer.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/executor/BuildExecutorTriggerer.java
@@ -17,15 +17,9 @@
  */
 package org.jboss.pnc.facade.executor;
 
-import org.jboss.pnc.bpm.BpmManager;
-import org.jboss.pnc.bpm.BpmTask;
-import org.jboss.pnc.bpm.model.ProcessProgressUpdate;
-import org.jboss.pnc.bpm.task.BpmBuildTask;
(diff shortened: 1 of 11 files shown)


facade/src/main/java/org/jboss/pnc/facade/executor/BuildExecutorTriggerer.java
```diff
@@ -94,7 +94,10 @@ public BuildExecutionSession executeBuild(
                         .getBpmTaskByBuildTaskId(bpmManager, statusChangedEvent.getBuildTaskId());
 
                 if (bpmTask.isPresent()) {
```
matejonnet, line 96: I think this whole block is redundant as the final result is sent by bpmNotifier.sendBuildExecutionCompleted

vibe13: Ok I was worried that since only statusChangedEvent with buildResults are sent by bpmNotifier.sendBuildExecutionCompleted, we could have missed some notifications. But, looking at DefaultBuildExecutionSession.setStatus, every DefaultBuildExecutorStatusChangedEvent which is final, has buildResults set. So I think that yes we are fine to remove it.

vibe13: @matejonnet OK I have pushed the changes and made some cleanup

vibe13:
retest this please

vibe13:
retest this please

vibe13:
retest this please

PR #3036, opened 2020-05-12
[NCL-5754] add duplication validation to productMilestone validation endpoint

michalovjan:
### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/providers/ProductMilestoneProviderImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/providers/ProductMilestoneProviderImpl.java
@@ -20,11 +20,13 @@ package org.jboss.pnc.facade.providers;
 import org.jboss.pnc.bpm.causeway.ProductMilestoneReleaseManager;
 import org.jboss.pnc.common.concurrent.Sequence;
 import org.jboss.pnc.common.logging.MDCUtils;
+import org.jboss.pnc.constants.Patterns;
 import org.jboss.pnc.dto.ProductMilestone;
 import org.jboss.pnc.dto.ProductMilestoneCloseResult;
 import org.jboss.pnc.dto.ProductMilestoneRef;
 import org.jboss.pnc.dto.response.MilestoneInfo;
 import org.jboss.pnc.dto.response.Page;
+import org.jboss.pnc.dto.response.ValidationResponse;
 import org.jboss.pnc.dto.validation.groups.WhenUpdating;
 import org.jboss.pnc.facade.providers.api.ProductMilestoneProvider;
 import org.jboss.pnc.facade.util.UserService;
@@ -63,13 +65,17 @@ import javax.persistence.criteria.Path;
 import javax.persistence.criteria.Root;
 import javax.persistence.criteria.SetJoin;
 import java.time.Instant;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import static org.jboss.pnc.enums.ValidationErrorType.DUPLICATION;
+import static org.jboss.pnc.enums.ValidationErrorType.FORMAT;
 import static org.jboss.pnc.spi.datastore.predicates.ProductMilestonePredicates.withProductVersionId;
 import static org.jboss.pnc.spi.datastore.predicates.ProductMilestonePredicates.withProductVersionIdAndVersion;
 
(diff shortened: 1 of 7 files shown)


michalovjan:
retest this please

facade/src/main/java/org/jboss/pnc/facade/providers/ProductMilestoneProviderImpl.java
```diff
@@ -275,6 +281,34 @@ public void cancelMilestoneCloseProcess(String id) throws RepositoryViolationExc
...
     }
 
+    @Override
+    public ValidationResponse validateVersion(String productVersionId, String version) {
+        boolean matches = Pattern.matches(Patterns.PRODUCT_MILESTONE_VERSION, version);
+        ValidationResponse.Builder builder = ValidationResponse.builder().isValid(matches);
+        if (!matches) {
+            return builder.errorType(FORMAT)
+                    .hints(
+                            Lists.newArrayList(
```
janinko, line 291: I would prefer Collections.singletonList() instead of using guava library

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/ProductMilestoneEndpointImpl.java
```diff
@@ -132,12 +129,7 @@ public void cancelMilestoneClose(String id) {
 
     @Override
     public ValidationResponse validateVersion(VersionValidationRequest versionRequest) {
```
janinko, line 131: add `@Valid` to VersionValidationRequest and I belive then you can rely on the fields not be null.

facade/src/main/java/org/jboss/pnc/facade/providers/ProductMilestoneProviderImpl.java
```diff
@@ -275,6 +281,34 @@ public void cancelMilestoneCloseProcess(String id) throws RepositoryViolationExc
...
+                                    "Allowed format consists of 2 or 3 numeric components (separated by a dot) followed by a string "
+                                            + "qualifier starting with a character, eg. 3.0.0.GA, 1.0.11.CR2.ER1, 3.0.CR2"))
+                    .build();
+        }
+        org.jboss.pnc.model.ProductMilestone duplicate = null;
+        if (productVersionId != null) {
```
janinko, line 297: if you use `@Valid` then you don't need this check, I think.

constants/src/main/java/org/jboss/pnc/enums/ErrorType.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.pnc.enums;
+
+public enum ErrorType {
```
janinko, line 20: ValidationErrorType? + javadoc on class level

facade/src/main/java/org/jboss/pnc/facade/providers/ProductMilestoneProviderImpl.java
michalovjan: Sure

facade/src/main/java/org/jboss/pnc/facade/providers/ProductMilestoneProviderImpl.java
michalovjan: I mean, I don't mind having null there. If it is null, then only REGEX check will be done.

constants/src/main/java/org/jboss/pnc/enums/ErrorType.java
michalovjan: I don't like Java style of long names. I'd like to keep it simple and generic. This enum could be used somewhere else.

facade/src/main/java/org/jboss/pnc/facade/providers/ProductMilestoneProviderImpl.java
janinko: Then why did you added `@NotBlank` to it in the DTO?

constants/src/main/java/org/jboss/pnc/enums/ErrorType.java
janinko: Now both values are for validation. Do you have any example of other values that could be put here now?

facade/src/main/java/org/jboss/pnc/facade/providers/ProductMilestoneProviderImpl.java
michalovjan: That's a valid point :D. I forgot that NotBlank doesn't allow nulls. I wanted it to cover a case "if there is a string don't let it be empty". I'll add the @Valid there.

constants/src/main/java/org/jboss/pnc/enums/ErrorType.java
michalovjan: The enum is a simple collection of error types. If someone wants to specify a type of error in their DTO, he can add them to the enum and use it. Right now, only milestone validation uses it, so now no other values are needed.

janinko: Then why would ValidationResponse class have generic error type that can have value that has nothing to do with validation? If it was a generic response class then I would understand that, but it's specific validation response.

michalovjan: Ok, I am convinced.

PR #3037, opened 2020-05-12
[NCL-5701] - Port tests to RestResponseFormattingTest

vibe13:
### Checklist:


integration-test/src/test/java/org/jboss/pnc/integration_new/RestResponseFormattingTest.java
```diff
@@ -78,24 +58,23 @@ public void shouldReturnErrorInJsonFormat() {
                 .statusCode(Response.Status.INTERNAL_SERVER_ERROR.getStatusCode())
                 .body("errorMessage", IsEqual.equalTo("Test exception."))
                 .when()
-                .get("/pnc-rest/rest/test/throw")
+                .get("/pnc-rest-new/rest-new/debug/throw")
```
janinko, line 61: Please extract the common API path to constant. It will change in the future.

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/internal/DebugEndpointImpl.java
```diff
@@ -75,4 +78,23 @@ public static Build newBuild() {
...
+
+    @Override
+    public void throwEx() throws Exception {
+        RuntimeException nested = new RuntimeException("Root exception.");
+        throw new Exception("Test exception.", nested);
+    }
+
+    @Override
+    public void nocontent() throws Exception {
+        Response.status(Response.Status.NO_CONTENT).build();
```
janinko, line 90: you build a response here but then do nothing with it. 
From JAX-RS specification
> 3.3.3 Return Type
> `void` Results in an empty entity body with a 204 status code.

So Having just empty method will do the works.

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/internal/DebugEndpointImpl.java
```diff
@@ -75,4 +78,23 @@ public static Build newBuild() {
...
+    }
+
+    @Override
+    public void redirect() throws Exception {
+        Response.status(Response.Status.UNAUTHORIZED)
+                .header(HttpHeaders.WWW_AUTHENTICATE, "Basic realm=\"ApplicationRealm\"")
+                .build();
```
janinko, line 97: How can this work? You build a repsoinse but then don't use it anywhere.

integration-test/src/test/resources/test-ds.xml
```diff
@@ -30,7 +30,7 @@
       <!-- Properties for Hibernate -->
       <property name="hibernate.dialect" value="org.hibernate.dialect.HSQLDialect"/>
       <property name="hibernate.hbm2ddl.auto" value="create-drop"/>
-      <property name="hibernate.show_sql" value="true"/>
+      <property name="hibernate.show_sql" value="false"/>
```
janinko, line 33: Why changing it?

vibe13: This was set to true by me some days ago, I am just reverting it back

rest-new/src/main/java/org/jboss/pnc/rest/endpoints/internal/DebugEndpointImpl.java
vibe13: thanks!

PR #1403, opened 2016-12-15
Ncl 2710

alexcreasy:

PR #3044, opened 2020-05-18
Java 1.8 compliance

pkocandr:
### Checklist:


jbrazdil:
Wierd, we have source set to 1.8 and it compiled. Does it causes problems on deployment?

janinko:
Wierd, we have source set to 1.8 and it compiled. Does it causes problems on deployment?

pkocandr:
> Wierd, we have source set to 1.8 and it compiled. Does it causes problems on deployment?

It caused java.lang.NoSuchMethodError: java.util.List.toArray(Ljava/util/function/IntFunction;)[Ljava/lang/Object; in build #1403 on devel master.

PR #3053, opened 2020-05-20
[NCL-5663] Update Keycloak denepdencies

janinko:


causeway-client/pom.xml
```diff
@@ -65,9 +65,12 @@
...
     <dependency>
-      <groupId>org.keycloak</groupId>
-      <artifactId>keycloak-core</artifactId>
-      <scope>test</scope>
+        <groupId>com.fasterxml.jackson.core</groupId>
+        <artifactId>jackson-annotations</artifactId>
+    </dependency>
+    <dependency>
+        <groupId>com.fasterxml.jackson.datatype</groupId>
+        <artifactId>jackson-datatype-jsr310</artifactId>
```
pkocandr, line 73: How are these 2 deps related to the keycloak version change?

pom.xml
```diff
@@ -224,6 +224,13 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>com.redhat.bom.rh-sso</groupId>
+        <artifactId>rh-sso-adapter-bom</artifactId>
+        <version>7.4.0.GA</version>
```
pkocandr, line 230: Does the version match the one in EAP 7.2.2.GA? We are not on 7.3 yet...

pkocandr: Also I would add a comment like "if changed version.keycloak value must be changed accordingly" to avoid the mess in the future.

causeway-client/pom.xml
janinko: I removed keycloak from causeway-client, which was sneaking these dependencies here.

pom.xml
janinko: We are going to move to EAP 7.3 "soon". Show I merge it anyway or wait for the move to 7.3 EAP?

PR #3077, opened 2020-06-02
[NCL-5765] improved BuildStatus documentation

michalovjan:
### Checklist:

### DIFF
--- a/constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
+++ b/constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
@@ -40,29 +40,55 @@ public enum BuildStatus {
     /**
      * Build/GroupBuild has failed. The status is propagated through dependants as REJECTED_FAILED_DEPENDENCIES.
      *
-     * A GroupBuild is reported FAILED with is at least one underlying Build with the final status of (FAILED, REJECTED,
+     * A GroupBuild is reported FAILED when at least one underlying Build has the final status of (FAILED, REJECTED,
      * SYSTEM_ERROR)
      */
     FAILED(FINISHED, false),
 
     /**
-     * A Build has been requested (possibly via dependencies) but no actual build happened as it was not required (no
-     * updates to underlying BuildConfig and Configs of dependencies (depends in IMPLICIT/EXPLICIT dependency check)).
+     * A Build has been requested (possibly via dependants) but no actual build happened as it was not required.
      * 
-     * A GroupBuild is deemed unnecessary to be built if all the underlying Builds do not require rebuild.
+     * For a previously successful Build, a rebuild is required when (conditions depend on a type of check):
+     *
+     * **Explicit**
+     *
+     * 1. The BuildConfig has been modified since the last successful Build
+     * 
+     * 2. A BuildConfig, explicitly defined as a dependency of this one, has been modified since the last successful
+     * Build
+     * 
+     * 3. There exists a newer, successful Build of a BuildConfig explicitly defined as a dependency of this one.
+     *
+     *
+     * **Implicit**
+     *
+     * 1. Explicit criteria plus:
+     * 
+     * 2. There exists a newer version of an implicit dependency which can be: A Build which produced artifacts that
+     * were used for alignment during build process or automatically captured from sources such as Indy, MRRC or Maven
+     * Central
+     *
+     *
+     * **Force**
+     * 
+     * 1. Always
+     * 
+     * A GroupBuild is deemed unnecessary to be built if all the included Builds do not require rebuild.
+     *
+     *
      */
     NO_REBUILD_REQUIRED(FINISHED, true),
 
     /**
-     * Build satisfies all necessary requirements to start building process. Build is placed into a queue and waits for a free
-     * BuildCoordinator thread to be picked up.
+     * A Build satisfies all necessary requirements to start building process. Build is placed into a queue and waits
+     * for a free BuildCoordinator thread to be picked up.
      *
      * With lower number of concurrent Builds, the Build is more likely to wait for available thread.
      */
     ENQUEUED(PENDING),
 
     /**
-     * A Build has unfinished dependencies, therefore cannot start building.
+     * A Build has unfinished dependencies, therefore waits so it can start building.
      */
     WAITING_FOR_DEPENDENCIES(PENDING),
 
--- a/constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
+++ b/constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
@@ -24,45 +24,58 @@ import static org.jboss.pnc.enums.BuildProgress.IN_PROGRESS;
 import static org.jboss.pnc.enums.BuildProgress.PENDING;
 
 /**
- * Status of a running or finished build.
+ * Status of a running/completed Build/GroupBuild.
+ * 
+ * The status class is shared by both Build and GroupBuild dtos, which are published through UI/WebSocket.
  *
  * @author Jakub Bartecek &lt;jbartece@redhat.com&gt;
  *
  */
 public enum BuildStatus {
     /**
-     * Build completed successfully.
+     * Build/GroupBuild has completed successfully. The dependant Builds can be scheduled.
      */
     SUCCESS(FINISHED, true),
 
     /**
-     * Build failed.
+     * Build/GroupBuild has failed. The status is propagated through dependants as REJECTED_FAILED_DEPENDENCIES.
+     *
+     * A GroupBuild is reported FAILED with is at least one underlying Build with the final status of (FAILED, REJECTED,
+     * SYSTEM_ERROR)
      */
     FAILED(FINISHED, false),
 
     /**
-     * A build has been requested (possibly via dependencies) but no actual build happened as it was not required (no
-     * updates).
+     * A Build has been requested (possibly via dependencies) but no actual build happened as it was not required (no
+     * updates to underlying BuildConfig and Configs of dependencies (depends in IMPLICIT/EXPLICIT dependency check)).
+     * 
+     * A GroupBuild is deemed unnecessary to be built if all the underlying Builds do not require rebuild.
      */
     NO_REBUILD_REQUIRED(FINISHED, true),
 
     /**
-     * A build has been placed in a queue. It will be processed shortly.
+     * Build satisfies all necessary requirements to start building process. Build is placed into a queue and waits for a free
+     * BuildCoordinator thread to be picked up.
+     *
+     * With lower number of concurrent Builds, the Build is more likely to wait for available thread.
      */
     ENQUEUED(PENDING),
 
     /**
-     * Build is waiting for dependencies to finish.
+     * A Build has unfinished dependencies, therefore cannot start building.
      */
     WAITING_FOR_DEPENDENCIES(PENDING),
 
     /**
-     * Build currently running.
+     * The Build/GroupBuild building process has started.
      */
     BUILDING(IN_PROGRESS),
 
     /**
-     * Build rejected due to conflict with another build.
+     * Build rejected due to conflict with another build. This happens when there is already a Build with the same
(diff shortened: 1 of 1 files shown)


constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
```diff
@@ -24,45 +24,58 @@
...
-     * Build completed successfully.
+     * Build/GroupBuild has completed successfully. The dependant Builds can be scheduled.
      */
     SUCCESS(FINISHED, true),
 
     /**
-     * Build failed.
+     * Build/GroupBuild has failed. The status is propagated through dependants as REJECTED_FAILED_DEPENDENCIES.
+     *
+     * A GroupBuild is reported FAILED with is at least one underlying Build with the final status of (FAILED, REJECTED,
```
janinko, line 43: `s/with is/when/` `s/with the/has the/`? Or smething like that, it doesn't read well now.

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
```diff
@@ -24,45 +24,58 @@
...
+     * SYSTEM_ERROR)
      */
     FAILED(FINISHED, false),
 
     /**
-     * A build has been requested (possibly via dependencies) but no actual build happened as it was not required (no
-     * updates).
+     * A Build has been requested (possibly via dependencies) but no actual build happened as it was not required (no
```
janinko, line 49: dependencies? I think it's dependants.
But do we require new build of a dependency when it doesn't require rebuild? Why and in which cases?

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
```diff
@@ -24,45 +24,58 @@
...
+     * updates to underlying BuildConfig and Configs of dependencies (depends in IMPLICIT/EXPLICIT dependency check)).
+     * 
+     * A GroupBuild is deemed unnecessary to be built if all the underlying Builds do not require rebuild.
```
janinko, line 52: `s/underlying/included/` or something like that?

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
```diff
@@ -24,45 +24,58 @@
...
+     * A GroupBuild is reported FAILED with is at least one underlying Build with the final status of (FAILED, REJECTED,
+     * SYSTEM_ERROR)
      */
     FAILED(FINISHED, false),
 
     /**
-     * A build has been requested (possibly via dependencies) but no actual build happened as it was not required (no
-     * updates).
+     * A Build has been requested (possibly via dependencies) but no actual build happened as it was not required (no
+     * updates to underlying BuildConfig and Configs of dependencies (depends in IMPLICIT/EXPLICIT dependency check)).
```
janinko, line 50: I would make it more clear by "(rebuild is required when there is ..."

I think rebuild is required when:
 * the BC changed
 * an explicit/implicit (transitive) dependency BC was rebuilt
 * an explicit/implicit (transitive) dependency BC changed
(if this is not true, do we have good implementation?)

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
```diff
@@ -24,45 +24,58 @@
...
+     * 
+     * A GroupBuild is deemed unnecessary to be built if all the underlying Builds do not require rebuild.
      */
     NO_REBUILD_REQUIRED(FINISHED, true),
 
     /**
-     * A build has been placed in a queue. It will be processed shortly.
+     * Build satisfies all necessary requirements to start building process. Build is placed into a queue and waits for a free
```
janinko, line 57: Why did you remove the articles? I belive there should be `a`? @alexcreasy

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
```diff
@@ -24,45 +24,58 @@
...
+     * BuildCoordinator thread to be picked up.
+     *
+     * With lower number of concurrent Builds, the Build is more likely to wait for available thread.
      */
     ENQUEUED(PENDING),
 
     /**
-     * Build is waiting for dependencies to finish.
+     * A Build has unfinished dependencies, therefore cannot start building.
```
janinko, line 65: Here you added article `a`

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
```diff
@@ -24,45 +24,58 @@
...
      */
     WAITING_FOR_DEPENDENCIES(PENDING),
 
     /**
-     * Build currently running.
+     * The Build/GroupBuild building process has started.
```
janinko, line 70: and here you added article `the`. Please make your mind :D 
cc @alexcreasy what should be correct?

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
janinko: I would use word waiting somewhere. Now the description looks like the build failed.

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
michalovjan: +:100:

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
michalovjan: +:100:

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
michalovjan: In my defense 'the' applies to process

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
michalovjan: +:100:

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
michalovjan: Yeah, there should be 'a'.

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
michalovjan: Good catch, it should be dependants. 

We require a rebuild of dependency if its BC has changed or if dependency's dependency needs rebuild(and so on). How you define which dependencies you are checking relies on IMPLICIT/EXPLICIT parameter.

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
michalovjan: Sure I can make it much more clear.

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
janinko: Yeah, but you say "A Build has been requested (possibly via dependants)". Do we request dependecies to build even when they don't require rebuild?

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
matedo1: Regarding rebuild possibilities, there are three modes:

- **Explicit**
    - The build config has been modified since the last successful build
    - A build config, explicitly defined as a dependency of this one, has been modified
    - since the last successful build
    - There exists a newer, successful build of a build config explicitly defined as a dependency of this one
- **Implicit**
    - Explicit criteria plus:
    - There exists a newer version of an implicit dependency (automatically
    - captured from sources such as Indy, MRRC or Maven Central)
- **Force**
    - Always

alexcreasy: @matedo1 that's the definition we use in the UI right? Hopefully someone with a better idea of the backend implementation, e.g. Honza or Matej can verify that our description is still accurate.

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
alexcreasy: Yes, the indefinite article is necessary here

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
michalovjan: Yes. If you request a build, it goes through explicit dependencies. The dependency is requested, but it finishes immediately with NO_REBUILD_REQUIRED(REJECTED_ALREADY_BUILT on prod) if its BC was not modified.

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
matedo1: @alexcreasy yes, the definition is coming from the UI

michalovjan: @alexcreasy @matedo1 WDYT about changing the Implicit description a little. For me, it's a little confusing with what we have now. 
I'd add a clause similar to this: 
     - implicit dependencies are also Builds that produced artifacts used during build process

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
janinko: I don't like that behaviour, but fine :)

constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java
alexcreasy: @michalovjan Martin's paste is a bit innacurate, it is missing the first line:

"Rebuild a previously successful build config when:"

So the bullet points are the minimum criteria which must be met for a rebuild to take place, for each rebuild mode. Your description of "implicit builds" makes no sense in that context, as it's not a criteria that must be met for a rebuild to take place.

If you want to rework the text some more so that your clarification fits in overall, by all means feel free. I don't have a complete understanding of rebuild modes, as they're exclusively backend features, so I'm not the best person to write / critique. Small modifications need to fit the overal context though.

Here's the full text from the UI popup menu, for reference:

```
Rebuild a previously successful build config when:

**Explicit**
- The build config has been modified since the last successful build
- A build config, explicitly defined as a dependency of this one, has been modified since the last successful build
- There exists a newer, successful build of a build config explicitly defined as a dependency of this one.

**Implicit**
- Explicit criteria plus:
- There exists a newer version of an implicit dependency (automatically captured from sources such as Indy, MRRC or Maven Central)

**Force**
- Always
```

alexcreasy: Oh on second reading, I'm wondering if you were just proposing to change the implicit clause rather than add a new one?

michalovjan: Yes, only implicit clause. I thought maybe instead of using parentheses, use further indentation.

Rebuild a previously successful build config when:

**Explicit**
- The build config has been modified since the last successful build
- A build config, explicitly defined as a dependency of this one, has been modified since the last successful build
- There exists a newer, successful build of a build config explicitly defined as a dependency of this one.

**Implicit**
- Explicit criteria plus:
- There exists a newer version of an implicit dependency that can be:
        - automatically captured from sources such as Indy, MRRC or Maven Central
        - a Build which produced artifacts that were used for alignment during build process

**Force**
- Always

michalovjan: ^ @alexcreasy @matedo1

michalovjan:
The formatting for Dependency checks is weird due to our code-formatter. It merges everything into one line unless there is empty line in between. It makes making bullet points hell.

michalovjan:
retest this please

PR #3080, opened 2020-06-04
[NCL-5680] - Update the model related to Artifacts and revision of quality levels change

vibe13:
Created a new ArtifactAudited table, to watch and record changes related to the ArtifactQuality related fields. Added new fields to record the creationTime, creationUser, modificationTime and modificationUser, and the reason why an ArtifactQuality has changed.

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/providers/ArtifactProviderImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/providers/ArtifactProviderImpl.java
@@ -26,7 +26,6 @@ import static org.jboss.pnc.spi.datastore.predicates.ArtifactPredicates.withSha1
 import static org.jboss.pnc.spi.datastore.predicates.ArtifactPredicates.withSha256;
 
 import java.time.Instant;
-import java.util.Date;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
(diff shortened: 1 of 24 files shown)


vibe13:
@janinko Hi! Sorry to pick you one more time, but as I discussed the above changes with Kuba, it would be great to also have you eagle-eyes on this :-). Thanks!

vibe13:
retest this please

facade/src/main/java/org/jboss/pnc/facade/providers/ArtifactProviderImpl.java
```diff
@@ -91,14 +119,26 @@ public ArtifactProviderImpl(
...
 
     @Override
     @RolesAllowed(SYSTEM_USER)
     public org.jboss.pnc.dto.Artifact update(String id, org.jboss.pnc.dto.Artifact restEntity)
             throws DTOValidationException {
-        return super.update(id, restEntity);
+        org.jboss.pnc.model.User currentUser = userService.currentUser();
+        User user = userMapper.toDTO(currentUser);
+        Instant now = Instant.now();
+        return super.update(id, restEntity.toBuilder().modificationUser(user).modificationTime(now).build());
```
janinko, line 141: Do we want to create new revision when the Quality level doesn't change? That means there will be 2 revisions only different with `modificationUser` and `modificationTime` but nothing else changed.

vibe13: Right, but in the datastore/src/main/java/org/jboss/pnc/datastore/repositories/ArtifactRepositoryImpl.java in the save method, if there are no changes to the audited values the modificationUser and time are reset, and no revision is created. I guess I could also revert the logic (only set modification time and users only if audited fields changed, in the datastore) if you prefer

PR #3090, opened 2020-06-10
Javadoc

janinko:


dto/src/main/java/org/jboss/pnc/dto/ArtifactRevision.java
```diff
@@ -30,6 +30,7 @@
 import lombok.ToString;
 
 /**
+ * Revision of a representation of an artifact created or used by build.
```
pkocandr, line 33: Drop the "representation of " also here please.

PR #3093, opened 2020-06-11
Add AdvancedSCMRepositoryClient for websocket notifications

thescouser89:
This will handle websocket notifications for both success and failures
of scm creation

### Checklist:

### DIFF
--- a/rest-client/src/main/java/org/jboss/pnc/restclient/AdvancedSCMRepositoryClient.java
+++ b/rest-client/src/main/java/org/jboss/pnc/restclient/AdvancedSCMRepositoryClient.java
@@ -1,0 +1,133 @@
+/**
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014-2020 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.pnc.restclient;
+
+import org.jboss.pnc.client.ClientException;
+import org.jboss.pnc.client.Configuration;
+import org.jboss.pnc.client.RemoteResourceException;
+import org.jboss.pnc.client.SCMRepositoryClient;
+import org.jboss.pnc.dto.notification.RepositoryCreationFailure;
+import org.jboss.pnc.dto.notification.SCMRepositoryCreationSuccess;
+import org.jboss.pnc.dto.requests.CreateAndSyncSCMRequest;
+import org.jboss.pnc.dto.response.RepositoryCreationResponse;
+import org.jboss.pnc.restclient.websocket.VertxWebSocketClient;
+import org.jboss.pnc.restclient.websocket.WebSocketClient;
+
+import java.util.concurrent.CompletableFuture;
+
+import static org.jboss.pnc.restclient.websocket.predicates.SCMRepositoryNotificationPredicates.withFailedTaskId;
+import static org.jboss.pnc.restclient.websocket.predicates.SCMRepositoryNotificationPredicates.withSuccessTaskId;
+
+public class AdvancedSCMRepositoryClient extends SCMRepositoryClient {
+
+    public AdvancedSCMRepositoryClient(Configuration configuration) {
+        super(configuration);
+    }
+
+    private CompletableFuture<RepositoryCreationFailure> waitForScmCreationFailure(
+            WebSocketClient webSocketClient,
+            String taskId) {
+        return webSocketClient.catchRepositoryCreationFailure(withFailedTaskId(taskId));
+    }
+
+    private CompletableFuture<SCMRepositoryCreationSuccess> waitForScmCreationSuccess(
+            WebSocketClient webSocketClient,
+            String taskId) {
+
+        return webSocketClient.catchSCMRepositoryCreationSuccess(withSuccessTaskId(taskId));
+    }
+
+    /**
+     * If the scm repository is already created and present in the Orch database, a RemoteResourceException is thrown.
+     * It's recommended that you check the presence of the repository before creating it.
+     *
+     * @param request scm to create
+     * @return CompletableFuture of the result, whether it was successful or not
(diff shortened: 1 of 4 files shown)


thescouser89:
Example of its use in CLI:

```
➜  bacon git:(master) ✗ java -jar cli/target/bacon.jar pnc scm-repository create-and-sync https://github.com/project-ncl/aprox-data.git   
---
success: true
scmRepositoryCreationSuccess:
  scmRepository:
    id: "368"
    internalUrl: "git+ssh://<internal url>/project-ncl/aprox-data.git"
    externalUrl: "https://github.com/project-ncl/aprox-data.git"
    preBuildSyncEnabled: true
  taskId: "14"
  job: "SCM_REPOSITORY_CREATION"
  notificationType: "SCMR_CREATION_SUCCESS"
  progress: "FINISHED"
  oldProgress: "IN_PROGRESS"
  message: null
repositoryCreationFailure: null
```

The switch to concurrent set is to avoid exceptions I got with `ConcurrentModificationException`

rest-client/src/main/java/org/jboss/pnc/restclient/AdvancedSCMRepositoryClient.java
```diff
@@ -0,0 +1,131 @@
...
+     *
+     * @param request scm to create
+     * @return CompletableFuture of the result, whether it was successful or not
+     * @throws RemoteResourceException Most likely thrown if repository already exists
+     * @throws ClientException if the response from the server is strange: both task id and scm repository is null
+     */
+    public CompletableFuture<SCMCreationResult> createNewAndWait(CreateAndSyncSCMRequest request)
+            throws RemoteResourceException, ClientException {
+
+        RepositoryCreationResponse response = super.createNew(request);
```
michalovjan, line 69: I'd recommend attaching WS listener before making REST requests. You have a guarantee you'll avoid race condition of the WS message coming earlier than WS listeners being registered.

rest-client/src/main/java/org/jboss/pnc/restclient/AdvancedSCMRepositoryClient.java
```diff
@@ -0,0 +1,131 @@
...
+import static org.jboss.pnc.restclient.websocket.predicates.SCMRepositoryNotificationPredicates.withFailedTaskId;
+import static org.jboss.pnc.restclient.websocket.predicates.SCMRepositoryNotificationPredicates.withSuccessTaskId;
+
+public class AdvancedSCMRepositoryClient extends SCMRepositoryClient {
+
+    private WebSocketClient webSocketClient = new VertxWebSocketClient();
+
+    public AdvancedSCMRepositoryClient(Configuration configuration) {
+        super(configuration);
+        webSocketClient.connect("ws://" + configuration.getHost() + "/pnc-rest-new/notifications").join();
```
michalovjan, line 42: With using the join() in the constructor, you theoretically limit reusability. If one of the wait* methods disconnects from the client, then other wait* methods won't work.(not a problem in this case) On the other hand, calling join() more then once is harmless, that's why I've been putting them inside wait* methods. 

If there is an issue with running join() concurrently, then adding `synchronized` keyword to join() method should fix that.

rest-client/src/main/java/org/jboss/pnc/restclient/AdvancedSCMRepositoryClient.java
thescouser89: I cannot do that since I need to listen to the task id :(

rest-client/src/main/java/org/jboss/pnc/restclient/AdvancedSCMRepositoryClient.java
thescouser89: Do you recommend I create 2 websocket clients then? one for each wait method?

thescouser89:
@michalovjan I moved the creation of the `VertxWebSocketClient` so that it's local to the method. I found out otherwise, for most operations in SCMRepository that *does not* involve websockets, the client would stay running and block the termination of the app because vertx is still running.

It's only when the `createAndSync` is called that the websocket client is closed, and therefore the app is terminated. So now with that tweak it's fixed. I wonder if that affects the other 'Advanced*' objects.

rest-client/src/main/java/org/jboss/pnc/restclient/AdvancedSCMRepositoryClient.java
michalovjan: shite, you are right...

rest-client/src/main/java/org/jboss/pnc/restclient/AdvancedSCMRepositoryClient.java
michalovjan: > @michalovjan I moved the creation of the `VertxWebSocketClient` so that it's local to the method. I found out otherwise, for most operations in SCMRepository that _does not_ involve websockets, the client would stay running and block the termination of the app because vertx is still running.
> 
> It's only when the `createAndSync` is called that the websocket client is closed, and therefore the app is terminated. So now with that tweak it's fixed. I wonder if that affects the other 'Advanced*' objects.

@thescouser89 
Yes, the circumstances are the same also for other Advanced clients... :( I guess we could create Vertx on join() if it's null (instead of the constructor of VertxWebSocketClient) and on disconnect() we close Vertx and make it null. This way Vertx is not created on WS client creation and its threads won't block CLI shutdown.

michalovjan:
> @michalovjan I moved the creation of the `VertxWebSocketClient` so that it's local to the method. I found out otherwise, for most operations in SCMRepository that _does not_ involve websockets, the client would stay running and block the termination of the app because vertx is still running.
> 
> It's only when the `createAndSync` is called that the websocket client is closed, and therefore the app is terminated. So now with that tweak it's fixed. I wonder if that affects the other 'Advanced*' objects.

@thescouser89
Yes, the circumstances are the same also for other Advanced clients... :( I guess we could create Vertx on join() if it's null (instead of the constructor of VertxWebSocketClient) and on disconnect() we close Vertx and make it null. This way Vertx is not created on WS client creation and its threads won't block CLI shutdown.

rest-client/src/main/java/org/jboss/pnc/restclient/AdvancedSCMRepositoryClient.java
michalovjan: I've made relevant comment below that could prevent this.

thescouser89:
I've pushed some changes to improve the filtering. Previously the lack of strict filtering caused a notification to be considered both a failure and a success. Woops!

thescouser89:
@michalovjan Not sure if better, but right now I just create the websocket client in the method so that it's local to the method (not global to the object). And then disconnect it at the end. The client is passed as parameter to the listeners.

thescouser89:
@michalovjan I'll merge it, but could we please continue the discussion on what further changes I could do when you come back? I'd like to get further feedback from you on those changes when you are back :)

PR #3099, opened 2020-06-16
[NCL-5665] Make sure build is successful even after OpenShift connect…

dominikbrazdil:
…ion error when destroying environment

### Checklist:

### DIFF
--- a/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
+++ b/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
@@ -27,6 +27,7 @@ import com.openshift.internal.restclient.model.properties.ResourcePropertiesRegi
 import com.openshift.restclient.ClientBuilder;
 import com.openshift.restclient.IClient;
 import com.openshift.restclient.NotFoundException;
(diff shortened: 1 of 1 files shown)


build-executor/pom.xml
```diff
@@ -129,6 +129,10 @@
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
     </dependency>
+      <dependency>
+          <groupId>com.openshift</groupId>
+          <artifactId>openshift-restclient-java</artifactId>
+      </dependency>
```
matejonnet, line 135: Build executor should not depend on Openshift client. The Openshift client is an implementation detail of the driver.

build-executor/src/main/java/org/jboss/pnc/executor/DefaultBuildExecutor.java
```diff
@@ -446,6 +447,8 @@ private void destroyEnvironment(BuildExecutionSession buildExecutionSession) {
             } else {
                 userLog.warn("Unable to destroy environment. Most likely due to cancelled operation.");
             }
+        } catch (OpenShiftException e) {
+            log.error("OpenShift connection ERROR.", e);
```
matejonnet, line 451: If the case is that the environment is not destroyed that have to be written in the log.

dominikbrazdil:
retest this please

PR #3120, opened 2020-06-23
[NCL-5705] Brew push: selecting build result in case of NO_REBUILD_RE…

jomrazek:
…QUIRED

### Checklist:


facade/src/main/java/org/jboss/pnc/facade/impl/BrewPusherImpl.java
```diff
@@ -151,6 +151,11 @@
 
     @Override
     public BuildPushResult pushBuild(String buildId, BuildPushParameters buildPushParameters) throws ProcessException {
+        BuildRecord build = buildRecordRepository.queryById(Integer.valueOf(buildId));
+        if (build.getStatus().equals(BuildStatus.NO_REBUILD_REQUIRED)) {
+            throw new OperationNotAllowedException(
+                    "Build has NO_REBUILD_REQUIRED status, force-rebuild build or push different one.");
```
janinko, line 157: I would maybe mention to push the one already built then to build a new one to push.

PR #3123, opened 2020-06-23
[NCL-5740] The "My Builds" section in the dashboard date order is reversed

vibe13:
### Checklist:


vibe13:
@alexcreasy I saw that no sorting is now put in the various ui/app/common/pnc-client/resources/ files. I guess that without using the SortHelper, it would be fine to pass parameters like this? Of course, I am ready to be scolded :-)

alexcreasy:
@vibe13 I think this is fine, the SortHelper is for use with the toolbars on the table component. As we don't use those here I don't see an issue!

Thanks for fixing this, it's been annoying me for a while but not hit the top of my todo list yet!

vibe13:
great, merging, thanks Alex! :-) will grab some easy UI issues maybe to alleviate your workload ;-)

PR #3131, opened 2020-06-25
Cut ties with the rest & rest-model modules

janinko:
Note that the change in `BuildTaskEndpointImpl` is changing the response frome these endpoints. It's no longer wrapped in Singleton wrapper, that means there is no longer this:
```
{
 "content": {...}
}
```
I think that this will break something. Should we:
- Change "something" to accept it without the wrapper
- Add the wrapper to this response? (new REST standard is not to use the Signleton wrapper. However, this is internal endpoint, there can be justifiable exceptions.)


janinko:
retest this please

thescouser89:
@janinko From my scanning of Maitai code, it looks like we ignore the results of the endpoint response. As long as the endpoint returns 200, we're happy.

So I'd suggest that you merge it for now and if anything is broken, we can discuss it further on changes required (and where the changes should be done)

janinko:
ONWARDS to breaking the builds!

PR #3141, opened 2020-06-26
[v1.8.1][NCLSUP-82] Search for temp, successful builds

thescouser89:
When deciding whether to only search for temp builds, or to include
permanent builds during implicit dependency check, we need to search for
*successful* temporary builds.

If only the search for *successful* temporary builds is empty, should we
consider permanent builds only.

If the extra successful check is not added, we'll also consider the not successful builds, which is not ok.

### Checklist:


thescouser89:
If @vibe13 is ok, @thescouser89 don't forget to add this to the changelog.

PR #3147, opened 2020-06-29
[NCL-5646]Redirection error after automatic login when click "update …

DnsZhou:
…milestone" button as a guest user

### Checklist:


DnsZhou:
When we use pnc-requires-auth directive, it will append a "disabled" property to the button. However, the "ng-disable" will override the "disabled" property with its own value. In order to keep the consistency of pnc-requires-auth directive, thus I found a walkaround to use "ng-show" instead of "ng-disable" to control the button for those milestones that have end date and can not be modified.

matedo1:
The same as Alex, code is looking good, let's merge it and test it on master, thanks for this fix!

PR #3177, opened 2020-07-09
[NCL-5787] Editing build configuration dependencies works

dominikbrazdil:
### Checklist:

### DIFF
--- a/ui/app/build-configs/detail/dependencies-tab/pncBuildConfigDependenciesTab.js
+++ b/ui/app/build-configs/detail/dependencies-tab/pncBuildConfigDependenciesTab.js
@@ -45,10 +45,14 @@
 
     function onRemove(dependency) {
       console.log('Remove dependency: %O', dependency);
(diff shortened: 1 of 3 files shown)


dominikbrazdil:
Thanks Alex for the comment, 
you are right that I was not fully aware of the reusability of the mentioned component, that is why I renamed the passed parameter to 'dependencies' instead of 'page' which seemed too broad to me. But I was also trying not to edit the component too much and I believe that all the other functional changes remained general enough for all the other usages. I only recreated the page with passed table data so the changes in it would be visible immediately (which did not happen when they were connected to parent component) and that should not affect the usages in other contexts. But tell me if I am overlooking something.  I cannot test it in other contexts much because other tabs are not working and in project version detail page there are some other problems with the functionality.

dominikbrazdil:
@alexcreasy I edited the variable's name so it can work everywhere. I think that the reasons for not working dependencies modification anywhere else than the dependencies tab are always caused by something else.

PR #3186, opened 2020-07-17
[NCL-5803] - Re-enable Artifact cache

vibe13:
### Checklist:


vibe13:
@matejonnet I haven't found which exact change caused the difference of behavior yet (has to be infinispan to me, which changed a minor version from 9.3.6 to 9.4.18 in EAP 7.3.1, hibernate seems unrelated). However, this simple change in code makes sure the field stays immutable also in the cache representation, and fixes the problem.

PR #3208, opened 2020-07-28
[NCL-5751] PR 3 - Moving 'Products' tab information to Details tab in…

DnsZhou:
… Build Configuration page

### Checklist:

### DIFF
--- a/ui/app/build-configs/detail/details-tab/pnc-build-config-details-editor.html
+++ b/ui/app/build-configs/detail/details-tab/pnc-build-config-details-editor.html
@@ -38,6 +38,32 @@
             <span class="help-block" ng-show="editBuildConfigForm.scmRevision.$invalid && editBuildConfigForm.scmRevision.$touched">Required field.</span>
           </div>
 
+          <!-- Validation: This field is optional-->
+          <div class="form-group"
+               ng-class="{ 'has-error': editBuildConfigForm.product.$invalid && editBuildConfigForm.product.$touched, 'has-success': editBuildConfigForm.product.$touched && editBuildConfigForm.version.$touched }">
+            <label class="control-label">Product</label>
+            <pnc-product-combobox
+              name="product"
+              ng-change="$ctrl.checkVersionData()"
+              ng-model="$ctrl.formData.product">
+            </pnc-product-combobox>
+          </div>
+
+          <!-- Validation: This field is mandatory if Product is selected-->
+          <div class="form-group"
+               ng-if="$ctrl.formData.product"
+               ng-class="{ 'has-error' : editBuildConfigForm.version.$invalid && editBuildConfigForm.version.$touched, 'has-success': editBuildConfigForm.product.$touched && editBuildConfigForm.version.$touched }">
+            <label class="control-label required">Version</label>
+            <pnc-product-version-combobox
+              name="version"
+              product="$ctrl.formData.product"
+              ng-model="$ctrl.formData.productVersion">
+            </pnc-product-version-combobox>
+            <p ng-show="editBuildConfigForm.version.$invalid && editBuildConfigForm.version.$touched"
+               class="help-block">
+              Version is a required field if you selected any product.</p>
+          </div>
+
         </div>
       </div>
 
--- a/ui/app/build-configs/detail/details-tab/pnc-build-config-details-editor.html
+++ b/ui/app/build-configs/detail/details-tab/pnc-build-config-details-editor.html
@@ -24,7 +24,7 @@
       <div class="row">
         <div class="col-md-12">
 
-          <pnc-create-build-config-general-form is-edit-form="$ctrl.isEditForm" original-build-type="$ctrl.originalBuildType" ng-model="$ctrl.formData.general"></pnc-create-build-config-general-form>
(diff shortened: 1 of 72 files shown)


ui/app/build-configs/detail/details-tab/pnc-build-config-details-editor.html
```diff
@@ -38,6 +38,32 @@
             <span class="help-block" ng-show="editBuildConfigForm.scmRevision.$invalid && editBuildConfigForm.scmRevision.$touched">Required field.</span>
           </div>
 
+          <!-- Validation: This field is optional-->
+          <div class="form-group"
```
alexcreasy, line 42: It would be much better if the 2 step product version selection code was encapsulated in its own reusable component:
- It is used in multiple places and it seems like a lot of code is copy and pasted to acheive this. The parent component then gets poluted with logic like the "checkVersionData" function which should be encapsulated in the new component.
- It also makes the codebase harder to maintain as fixes may have to take place in multiple locations, this often means they are only fixed at the site reported in a bug report by whoever picks up the ticket!

ui/app/build-configs/detail/details-tab/pncBuildConfigDetailsEditor.js
```diff
@@ -60,14 +64,31 @@
...
+
     function submit() {
       $ctrl.working = true;
       var buildConfig = toBuildConfig($ctrl.formData, $ctrl.buildConfig);
 
       BuildConfigResource.safePatchRemovingParameters($ctrl.buildConfig, buildConfig).$promise
-          .then(resp => $ctrl.onSuccess({ buildConfig: resp}))
+          .then(resp => {
+            $ctrl.onSuccess({ buildConfig: resp});
+            $state.go('projects.detail.build-configs.detail', {
```
alexcreasy, line 85: Why are you navigating to the same state? Especially with reload true this causes everything to be reloaded from the backend.

DnsZhou: Hi Alex,
The purpose of this action is to solve the issue that when the user edited the product version information, it can't be updated until the page was refreshed, see https://github.com/project-ncl/pnc/pull/3188#discussion_r459250436
I will try to find some solution to update the data without refreshing the page or using the "=" symbol to pass the product version information either.

Thanks

ui/app/build-configs/detail/details-tab/pnc-build-config-details-editor.html
DnsZhou: Hi Alex(@alexcreasy ),

Thanks for the review!

I tried to encapsulate it in an individual module, there are 3 usages for the product versions but they are somehow different: 

1. **New Build Config wizard**(https://github.com/DnsZhou/pnc/blob/NCL-5751-P2/ui/app/build-configs/directives/pnc-create-build-config-product-version-form/pnc-create-build-config-product-version-form.html#L21-L33), 

2. **Create Group Config page**(https://github.com/DnsZhou/pnc/blob/NCL-5751-P2/ui/app/group-configs/create/pnc-group-config-create-page.html#L38-L54) 

3. **Build Config detail tab**(which is the code you mentioned here)

The biggest difference is that the UI layout it use was different, e.g., 2. **Create Group Config page** uses "col-sm-3 col-lg-2" to control the label and for the input field it is "col-sm-9 col-lg-7". But 1. **New Build Config wizard** do not have such layouts.

So I can write a new component for 1. **New Build Config wizard** and 3. **Build Config detail tab**, and will launch another review request to you after I finished and commit that.

Cheers!

-Dennis

matedo1: Hi @alexcreasy , there is already existing jira (NCL-5875) for this and I think it's already in progress and @DnsZhou is working on it

matedo1: Hi @DnsZhou if layout customisation is the only problem, I think we should be able to solve it using some custom parameter at least, let's move some additional investigation into jira

DnsZhou: Thanks @matedo1 . 

So @alexcreasy if you don't mind, I think we can proceed the encapsulation of the product version component in jira NCL-5875, for this PR I will just fix the second issue you mentioned and merge the code after you and Martin reviewed the fix, so that we can unblock Dominik's task about product version link.

Thanks!
-Dennis

ui/app/build-configs/detail/details-tab/pncBuildConfigDetailsEditor.js
DnsZhou: Fixed and committed to this PR

DnsZhou:
Hi @alexcreasy @matedo1 ,

As I mentioned in the previous comments, I will do the encapsulation work in NCL-5875, and for the second issue of unnecessary page refreshment, I have fixed it in the latest commit.
Because of the code conflict again, I raised another PR to commit the clean patch with latest codebase in https://github.com/project-ncl/pnc/pull/3223 , so I will close this PR and move everything to that PR.

Thanks!
-Dennis

PR #3226, opened 2020-08-05
[NCL-5895] Auto-close vertx websocket for the advanced clients

thescouser89:
In some cases, we can use the Advanced* clients without having to use
any websocket listener. In those cases, the websocket client is never
closed / disconnected and the vertx event loop runs forever in the
background, causing the application to never exit properly.

This commit adapts the try-with-resources (AutoCloseable) feature in
those clients so that the websocket clients are properly closed once we
are done with them.

### Checklist:

### DIFF
--- a/rest-client/src/main/java/org/jboss/pnc/restclient/AdvancedBuildClient.java
+++ b/rest-client/src/main/java/org/jboss/pnc/restclient/AdvancedBuildClient.java
@@ -34,7 +34,13 @@ import java.util.concurrent.TimeoutException;
 import static org.jboss.pnc.restclient.websocket.predicates.BuildPushResultNotificationPredicates.withBuildId;
 import static org.jboss.pnc.restclient.websocket.predicates.BuildPushResultNotificationPredicates.withPushCompleted;
 
-public class AdvancedBuildClient extends BuildClient {
+/**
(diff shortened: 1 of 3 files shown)


thescouser89:
@michalovjan Trying auto-close to see if that's better to close those pesky vertx websocket clients

michalovjan:
@thescouser89 I was thinking of a solution that would use JVMs shutdown hooks to do that automatically but I haven't tried it.

dwalluck:
For the normal clients, it seems that RESTEasy uses the `Object.finalize()` method to close the sockets during garbage collection. But, proper practice would be to implement `Closable` or even `AutoClosable`. The clients aren't even offering the ability to try to close them. This finalizer is supposed to be a last-ditch effort and results in the following warning:

```
RESTEASY004687: Closing a class org.jboss.resteasy.client.jaxrs.engines.ApacheHttpClient43Engine instance for you. Please close clients yourself.
```

michalovjan:
@dwalluck @thescouser89 my design philosophy was that noone using the AdvancedClients should have to close WS client by hand with try-with-resources or invoking some Close() (should be done automatically). I have improvements in mind so we are able to do that but I just didn't get to it quite frankly.

thescouser89:
Ah no worries! Let's discuss it

PR #3233, opened 2020-08-12
[NCL-5891] add generalized way to update OneToMany relationships and add functionality to endpoints

michalovjan:
### Checklist:

### DIFF
--- a/datastore/src/main/java/org/jboss/pnc/datastore/repositories/internal/AbstractRepository.java
+++ b/datastore/src/main/java/org/jboss/pnc/datastore/repositories/internal/AbstractRepository.java
@@ -28,8 +28,14 @@ import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
 import javax.inject.Inject;
 import javax.persistence.EntityManager;
 import java.io.Serializable;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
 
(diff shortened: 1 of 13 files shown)


datastore/src/main/java/org/jboss/pnc/datastore/repositories/internal/AbstractRepository.java
```diff
@@ -110,4 +116,54 @@ public int count(Collection<Predicate<T>> andPredicates, Collection<Predicate<T>
...
+     * @param <N>
+     */
+    public <N extends GenericEntity<ID>> void cascadeUpdates(
+            N managedNonOwning,
+            N updatedNonOwning,
+            Function<N, Collection<T>> collectionGetter,
+            BiConsumer<T, N> owningSetter,
+            java.util.function.Predicate<T>... filter) {
+        // we want only to compare IDs because the data from request can be partial (due to dto refs in Maps)
+        Comparator<T> idComparator = (t, t1) -> t.getId().equals(t1.getId()) ? 0 : 1;
```
janinko, line 136: This will break the TreeSet as `if A != B  then A > B && A < B`. (or in best case will be as efficient as array list)

You would want to use `Map<ID, Entity>` I think.

thescouser89:
@michalovjan I'll defer my review to somebody more competent :( I'm not familiar enough with repository / JPA stuff to be able to review it properly.

datastore/src/main/java/org/jboss/pnc/datastore/repositories/internal/AbstractRepository.java
michalovjan: Yes, TreeSet will be inefficient, you are absolutely correct. Map with ID as a Key basically does exactly what I need. Thx for the suggestion.

PR #3234, opened 2020-08-12
create Vertx on first WS connection to avoid blocking on creation

michalovjan:
### Checklist:


thescouser89:
Thanks for the PR! I believe this doesn't cover all the corner cases in bacon, but let's see! We'll have to double-check.

michalovjan:
@thescouser89 You are absolutely correct. This should alleviate some at least **some** blocking pressure from Vertx. To handle all edge cases, much more aggressive approach is needed. (custom Java 9 Cleaner or @Override Object.finalize()). Sorry for the troubles.

thescouser89:
Ah please don't apologize for it!!! It's a ridiculously hard problem, and your code is not the reason for it. Silly stupid Vertx for creating threads that never die

PR #3239, opened 2020-08-17
[NCL-5894] Return 404 when there is no build group.

matejonnet:
### Checklist:


facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
```diff
@@ -423,7 +423,9 @@ public SSHCredentials getSshCredentials(String buildId) {
     @Override
     public Graph<Build> getBuildGraphForGroupBuild(String groupBuildId) {
         List<String> runningAndStoredIds = getBuildIdsInTheGroup(groupBuildId);
-
+        if (runningAndStoredIds.isEmpty()) {
+            throw new EmptyEntityException("Build group " + groupBuildId + " does not exists or has no builds.");
```
janinko, line 427: We should return 404 when group build does not exists and empty graph when there are no builds in existing group build.

matejonnet: https://github.com/project-ncl/pnc/pull/3246

PR #3246, opened 2020-08-19
[NCL-5894] Return 404 when there is no build group

matejonnet:
…n there are no builds in it.

### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
@@ -422,10 +422,12 @@ public class BuildProviderImpl extends AbstractProvider<Integer, BuildRecord, Bu
 
     @Override
     public Graph<Build> getBuildGraphForGroupBuild(String groupBuildId) {
-        List<String> runningAndStoredIds = getBuildIdsInTheGroup(groupBuildId);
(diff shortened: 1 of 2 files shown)


facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
```diff
@@ -422,10 +422,12 @@ public SSHCredentials getSshCredentials(String buildId) {
 
     @Override
     public Graph<Build> getBuildGraphForGroupBuild(String groupBuildId) {
-        List<String> runningAndStoredIds = getBuildIdsInTheGroup(groupBuildId);
-        if (runningAndStoredIds.isEmpty()) {
-            throw new EmptyEntityException("Build group " + groupBuildId + " does not exists or has no builds.");
+        BuildConfigSetRecord buildConfigSetRecord = buildConfigSetRecordRepository
+                .queryById(Integer.valueOf(groupBuildId));
+        if (buildConfigSetRecord != null) {
```
janinko, line 427: should there be `==`?

janinko (changes requested):

facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
matejonnet: Right :)

PR #3270, opened 2020-08-31
[NCL-5989] don't send WS message with no async job running

michalovjan:
### Checklist:


janinko:
When not sending error message, I think we should throw an exception instead.

janinko (approved):

PR #3279, opened 2020-09-03
[NCL-6008] - Milestone Brewpush not showing brew-url

vibe13:
### Checklist:

### DIFF
--- a/ui/app/product-milestones/close-result/pnc-product-milestone-close-result-page.html
+++ b/ui/app/product-milestones/close-result/pnc-product-milestone-close-result-page.html
@@ -30,22 +30,38 @@
   <dd>{{ $ctrl.data.startingDate | date:'medium' }}</dd>
   <dt>Close Finished:</dt>
   <dd>{{ ($ctrl.data.endDate | date:'medium') || '&mdash;' }}</dd>
-
-  <div ng-repeat="buildPushResult in $ctrl.data.buildPushResults">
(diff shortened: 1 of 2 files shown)


alexcreasy (changes requested):
So I'd change this slightly, I think just having each build in the main list will be a bit confusing. Also I'm pretty sure having a <div> tag as a direct child of a <dl> is illegal... (then again most browsers don't care)

Can I suggest that rather than embed them in the main definition list, it might be better to add a new sub heading with a title such as "Pushed Builds" (similar to the "log" sub heading) and then have a table, doing an ng-repeat on each row, with a column for status, id, brewBuildId and brew Build URL ?

alexcreasy:
Ok so I felt a bit nerdy and looked up the div within a dl issue... the rules are:

Either: Zero or more groups each consisting of one or more <dt> elements followed by one or more <dd> elements, optionally intermixed with <script> and <template> elements.
Or: One or more <div> elements, optionally intermixed with <script> and <template> elements.

So, my interpretation is you can either have dt tags or div tags? Either way it's complicated... :)

vibe13:
@alexcreasy Thanks for the review and the suggestion, much appreciated!! Just pushed another attempt :-D thanks!

alexcreasy (approved):
Looks good to me!

vibe13:
thanks Alex!!

PR #3282, opened 2020-09-04
Run Intellij inspections on the common module

thescouser89:
Code cleanups include:

- Use diamond operator
- Remove redundant fields in annotations
- Use static inner class when possible for memory
- Use method reference for lambdas
- Replace string "UTF-8" with StandardCharsets.UTF_8
- Remove unecessary variables that store result that we immediately
  return
- Use Java style array declaration rather than C style

### Checklist:

### DIFF
--- a/common/src/main/java/org/jboss/pnc/common/json/AbstractModuleGroup.java
+++ b/common/src/main/java/org/jboss/pnc/common/json/AbstractModuleGroup.java
@@ -18,12 +18,7 @@
 package org.jboss.pnc.common.json;
 
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
-import com.fasterxml.jackson.annotation.JsonTypeInfo.As;
 
-@JsonTypeInfo(
-        use = JsonTypeInfo.Id.NAME,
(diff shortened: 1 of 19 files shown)


common/src/main/java/org/jboss/pnc/common/json/AbstractModuleGroup.java
```diff
@@ -18,12 +18,7 @@
...
-import com.fasterxml.jackson.annotation.JsonTypeInfo.As;
 
-@JsonTypeInfo(
-        use = JsonTypeInfo.Id.NAME,
-        include = As.PROPERTY,
-        property = "@module-group",
-        defaultImpl = DefaultModuleGroup.class)
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "@module-group", defaultImpl = DefaultModuleGroup.class)
 public abstract class AbstractModuleGroup {
 }
```
dwalluck, line 24: Should configure IDEA to add newline at end of file (Editor > General > Ensure an empty line at the end of the file on Save) or add `.editorconfig` to do it automatically.

PR #3296, opened 2020-09-09
[NCL-6032] - PNC should wait for a pod to be available instead of giv…

vibe13:
…ing a SYSTEM_ERROR after 300s

### Checklist:

### DIFF
--- a/common/src/main/java/org/jboss/pnc/common/monitor/PollingMonitor.java
+++ b/common/src/main/java/org/jboss/pnc/common/monitor/PollingMonitor.java
@@ -41,33 +41,24 @@ public class PollingMonitor {
 
     /** Time how long to wait until all services are fully up and running (in seconds) */
     public static final int DEFAULT_TIMEOUT = 300;
-    private static final String PULLING_MONITOR_TIMEOUT_KEY = "pulling_monitor_timeout";
 
     /** Interval between two checks if the services are fully up and running (in second) */
-    private static final int DEFAULT_CHECK_INTERVAL = 1;
-    private static final String PULLING_MONITOR_CHECK_INTERVAL_KEY = "pulling_monitor_check_interval";
+    public static final int DEFAULT_CHECK_INTERVAL = 1;
 
     /** */
     private static final TimeUnit DEFAULT_TIME_UNIT = TimeUnit.SECONDS;
 
-    private static final String PULLING_MONITOR_THREADPOOL_KEY = "pulling_monitor_threadpool";
+    private static final String POLLING_MONITOR_THREADPOOL_KEY = "polling_monitor_threadpool";
     private static final int DEFAULT_EXECUTOR_THREADPOOL_SIZE = 4;
 
     private ScheduledExecutorService executorService;
     private ScheduledExecutorServiceWithTimeout scheduledExecutor;
 
-    private int timeout;
-    private int checkInterval;
-
     public PollingMonitor() {
         ReadEnvProperty reader = new ReadEnvProperty();
 
-        timeout = reader.getIntValueFromPropertyOrDefault(PULLING_MONITOR_TIMEOUT_KEY, DEFAULT_TIMEOUT);
-        checkInterval = reader
-                .getIntValueFromPropertyOrDefault(PULLING_MONITOR_CHECK_INTERVAL_KEY, DEFAULT_CHECK_INTERVAL);
-
         int threadSize = reader
-                .getIntValueFromPropertyOrDefault(PULLING_MONITOR_THREADPOOL_KEY, DEFAULT_EXECUTOR_THREADPOOL_SIZE);
+                .getIntValueFromPropertyOrDefault(POLLING_MONITOR_THREADPOOL_KEY, DEFAULT_EXECUTOR_THREADPOOL_SIZE);
 
         executorService = MDCExecutors.newScheduledThreadPool(threadSize);
         scheduledExecutor = new ScheduledExecutorServiceWithTimeout(executorService);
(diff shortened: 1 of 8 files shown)


common/src/main/java/org/jboss/pnc/common/monitor/PollingMonitor.java
```diff
@@ -84,6 +84,19 @@ public PollingMonitor() {
         return monitor(condition, checkInterval, timeout, DEFAULT_TIME_UNIT);
```
thescouser89, line 84: I think we can get rid of like 65:
```
timeout = reader.getIntValueFromPropertyOrDefault(PULLING_MONITOR_TIMEOUT_KEY, DEFAULT_TIMEOUT);
```

Since that'll make it confusing on how to configure the default polling monitor timeout (env var v/s config file). Then on line 84, we can just use the 'DEFAULT_TIMEOUT' as our timeout. WDYT?

thescouser89 (approved):
LGTM! Just a suggested cleanup to avoid confusion.

common/src/main/java/org/jboss/pnc/common/monitor/PollingMonitor.java
thescouser89: I think I also added that env var in the README.md. So if we decide to remove it, we should also remove it in the README.md

vibe13: @thescouser89  It makes sense to me Dustin! While looking at the code I was already wondering about that, but should we then add pulling_monitor_check_interval to the env vars as well (and remove it as you suggested for pulling_monitor_timeout)? As it would also be confusing that 2 similar properties are set in different ways? 
The pulling_monitor_threadpool is trickier to be moved to I would leave it where it is.
Thanks!

vibe13:
@thescouser89 @matejonnet  I also pushed some more changes to handle additional statuses

common/src/main/java/org/jboss/pnc/common/monitor/PollingMonitor.java
thescouser89: I'm a tiny bit confused. pulling_monitor_check_interval should already be an env var.

Out of topic, but I think I remember now why I did set it as env vars. It was because I couldn't access the module config there and had to find some loophole to configure it.

moduleconfig/src/main/java/org/jboss/pnc/common/json/moduleconfig/OpenshiftEnvironmentDriverModuleConfig.java
```diff
@@ -92,9 +98,30 @@ public OpenshiftEnvironmentDriverModuleConfig(
...
+                this.creationPodRetry = Integer.parseInt(creationPodRetry);
+            } catch (NumberFormatException e) {
+                log.error(
+                        "Couldn't parse the value of creation pod retry from the configuration. Using default ({} retries)",
+                        DEFAULT_CREATION_POD_RETRY);
+            }
+        }
+
+        this.pollingMonitorTimeout = PollingMonitor.DEFAULT_TIMEOUT;
+        if (pollingMonitorTimeout != null) {
```
matejonnet, line 115: Since the value is assigned in the previous line from the constant, what is the case for null ?

openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
```diff
@@ -476,8 +478,21 @@ private void monitorInitialization(
                         if (retries == 0) {
                             onError.accept(new Exception(throwable));
                         } else {
-                            logger.error("Creating build environment failed! Retrying...", throwable);
-                            retryEnvironment(onComplete, onError, retries);
+                            PodFailedStartException pfsExc = null;
```
matejonnet, line 481: Please use "readable" variable name. eg: failedStartException

matejonnet (changes requested):

common/src/main/java/org/jboss/pnc/common/monitor/PollingMonitor.java
vibe13: Ok sorry, I created the confusion. So looking more carefully, both pulling_monitor_check_interval and pulling_monitor_timeout are configurable via env vars.  I would leave them as indeed there is no access to the module config file there, and they can be used to override the default values.

However, the PollingMonitor is used not only in the OpenshiftStartedEnvironment, but also in the DefaultBuildCoordinator, so I would not change the default values. What if we leave the chance to override the default values also via config file, so that:
* if PollingMonitor is used as a utility class from the common package, users can use default or override them via env vars
* if OpenShiftStartedEnvironment needs to apply different default values for the timeout or check interval, it can do so without impacting others (e.g. DefaultBuildCoordinator)
Depending on the use case, different checks may be tuned with different timeouts and intervals?

Last thing: it is pOllingmonitor right? Not pUllingmonitor. What if I unify all the naming to and env vars to the correct name? :-)
Thanks!

moduleconfig/src/main/java/org/jboss/pnc/common/json/moduleconfig/OpenshiftEnvironmentDriverModuleConfig.java
vibe13: pollingMonitorTimeout comes from the config file mapping and is passed as parameter, it's not the same variable

openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
vibe13: sure, done.

common/src/main/java/org/jboss/pnc/common/monitor/PollingMonitor.java
thescouser89: To be honest, I think the env vars would be confusing. I would personally remove it and just make the caller specify a custom check internal and timeout if required with method parameters. We also don't use the env vars anymore in our deployments so it's safe to remove.

vibe13:
@thescouser89 @matejonnet Ok I made all the changes, let me know if this is good to be merged now. Thanks! :-)

thescouser89:
Looks good to be merged now!

PR #3300, opened 2020-09-11
[NCL-6016] Validate ExternalUrl does not exist when updating SCMR

dominikbrazdil:
### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/providers/SCMRepositoryProviderImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/providers/SCMRepositoryProviderImpl.java
@@ -184,15 +184,15 @@ public class SCMRepositoryProviderImpl
     }
 
     @Override
-    public SCMRepository update(String id, SCMRepository restEntity) {
-        validateBeforeUpdating(id, restEntity);
+    public void validateBeforeUpdating(String id, SCMRepository restEntity) {
+        super.validateBeforeUpdating(id, restEntity);
         SCMRepository before = getSpecific(id);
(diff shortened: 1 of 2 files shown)


facade/src/main/java/org/jboss/pnc/facade/providers/SCMRepositoryProviderImpl.java
```diff
@@ -190,6 +190,7 @@ public SCMRepository update(String id, SCMRepository restEntity) {
         if (!before.getInternalUrl().equals(restEntity.getInternalUrl())) {
             throw new InvalidEntityException("Updating internal URL is prohibited. SCMRepo: " + id);
         }
+        validateRepositoryWithExternalURLDoesNotExist(restEntity.getExternalUrl());
```
michalovjan, line 193: I'd Override the validateBeforeUpdating method and move this line there.

michalovjan (changes requested):

michalovjan (approved):

facade/src/main/java/org/jboss/pnc/facade/providers/SCMRepositoryProviderImpl.java
```diff
@@ -186,13 +186,21 @@ public RepositoryCreationResponse createSCMRepository(
     @Override
     public SCMRepository update(String id, SCMRepository restEntity) {
```
michalovjan, line 187: You can just delete this method now :D No need to override. The update from AbstractProvider is the same after all now.

michalovjan (changes requested):

michalovjan (approved):

PR #3324, opened 2020-09-24
[NCL-6085] Allow only SUCCESS builds to be pushed to brew

dominikbrazdil:
### Checklist:

### DIFF
--- a/ui/app/common/pnc-client/resources/helpers/buildStatusHelper.js
+++ b/ui/app/common/pnc-client/resources/helpers/buildStatusHelper.js
@@ -23,11 +23,11 @@
       .factory('buildStatusHelper', [
         function () {
 
-          var PENDING = 'PENDING';
-          var IN_PROGRESS = 'IN_PROGRESS';
-          var FINISHED = 'FINISHED';
+          const PENDING = 'PENDING';
+          const IN_PROGRESS = 'IN_PROGRESS';
+          const FINISHED = 'FINISHED';
 
-          var buildStatus = Object.freeze({
+          const buildStatus = Object.freeze({
             'NEW': {
               progress: PENDING,
               failed: false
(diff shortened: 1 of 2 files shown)


matedo1 (approved):

ui/app/common/pnc-client/resources/helpers/buildStatusHelper.js
```diff
@@ -125,10 +125,11 @@
           }
 
           function isSuccess(build) {
-            return !isFailed(build);
+            var status = getStatus(build);
```
matedo1, line 128: @dominikbrazdil  please preferf `const` next time, or `let` if it needs to be reassigned

I think status `DONE` is not valid option anymore, so I would prefer using `$isSuccess()` in `pncBrewPushButton.js` or add comment why it can not be there

matedo1 (changes requested):

ui/app/common/pnc-client/resources/helpers/buildStatusHelper.js
dominikbrazdil: Ok, if it is not valid anymore, then I'll leave there just SUCCESS state. Till now it returned true on every state that is not failed and that includes many of them. I just did not want to break something that may be using DONE status. Should I create an issue for removing the old statuses?

matedo1: +1 to create new issue to refactor `DONE` status out in favour of `SUCCESS`, there can be also some other deprecated statuses so creating a new jira fixing it all would be the best

Feel free to mention also this file https://github.com/project-ncl/pnc/blob/master/constants/src/main/java/org/jboss/pnc/enums/BuildStatus.java#L165 there.

matedo1 (approved):

PR #3332, opened 2020-09-30
[NCL-5812] Make sure LazyLoading Proxy is used when accesing fields

janinko:
When a proxy is loaded, the data is not stored in thr proxy object, but in the
“target” object. The fields in the proxy object will remain null forever (or any
value that they have been initialized with), since all methods invoked on the
proxy will be delegated to the target object. [1]

[1]: https://xebia.com/blog/advanced-hibernate-proxy-pitfalls/

### Checklist:


vibe13 (approved):
+1 LGTM! Thanks!

PR #3335, opened 2020-10-01
[NCL-5812] Update equals & hashCode on model entities to work with JPA

janinko:
An entity must be equal to itself across all JPA object states: transient,
attached, detached, removed [1]

The entities should use business key (unique, unmodifiable fields) in equals &
hashCode when possible, in the rest cases they can use primary key for equals
and constant for hash code [2]

Entities with business key:
BuildEnvironment
BuildRecordAttribute
GenericSetting
RepositoryConfiguration
TargetRepository
User

Some entities don't have business key but their ID is generated prior
persistence to DB, so we can use the ID without problems:
BuildConfiguration
BuildRecord
BuildRecordPushResult
ProductMilestoneRelease

Audited classes are only read from DB so their ID is unmodifiable:
ArtifactAudited
BuildConfigurationAudited

Enitities that don't have business key and their ID is generated when stored to
DB must have constat for their hash code:
Artifact
BuildConfigSetRecord
BuildConfigurationSet
Product
ProductMilestone
ProductRelease
ProductVersion
Project

[1]: https://vladmihalcea.com/hibernate-facts-equals-and-hashcode/
[2]: https://vladmihalcea.com/how-to-implement-equals-and-hashcode-using-the-jpa-entity-identifier/

### Checklist:


janinko:
retest this please

janinko:
retest this please

model/src/main/java/org/jboss/pnc/model/ArtifactAudited.java
```diff
@@ -130,17 +131,14 @@ public String toString() {
...
         if (this == o)
             return true;
-        if (o == null || getClass() != o.getClass())
+        if (!(o instanceof ArtifactAudited))
             return false;
-
-        ArtifactAudited that = (ArtifactAudited) o;
-
-        return (idRev != null ? idRev.equals(that.getIdRev()) : false);
+        return idRev != null && idRev.equals(((ArtifactAudited) o).getId());
```
vibe13, line 136: Isn't it `idRev != null && idRev.equals(((ArtifactAudited) o).getIdRev());` ?

model/src/main/java/org/jboss/pnc/model/BuildConfigurationAudited.java
```diff
@@ -224,17 +225,14 @@ public String toString() {
...
         if (this == o)
             return true;
-        if (o == null || getClass() != o.getClass())
+        if (!(o instanceof ArtifactAudited))
             return false;
-
-        BuildConfigurationAudited that = (BuildConfigurationAudited) o;
-
-        return (idRev != null ? idRev.equals(that.getIdRev()) : false);
+        return idRev != null && idRev.equals(((ArtifactAudited) o).getId());
```
vibe13, line 230: `idRev != null && idRev.equals(((BuildConfigurationAudited) o).getIdRev());`

model/src/main/java/org/jboss/pnc/model/BuildConfigurationAudited.java
```diff
@@ -224,17 +225,14 @@ public String toString() {
     public boolean equals(Object o) {
         if (this == o)
             return true;
-        if (o == null || getClass() != o.getClass())
+        if (!(o instanceof ArtifactAudited))
```
vibe13, line 228: `(!(o instanceof BuildConfigurationAudited))`

model/src/main/java/org/jboss/pnc/model/BuildConfigurationSet.java
```diff
@@ -212,18 +212,16 @@ public void setArchived(boolean archived) {
     public boolean equals(Object o) {
         if (this == o)
             return true;
-        if (o == null || getClass() != o.getClass())
+        if (!(o instanceof BuildConfigSetRecord))
```
vibe13, line 215: `(!(o instanceof BuildConfigurationSet))`

model/src/main/java/org/jboss/pnc/model/BuildEnvironment.java
```diff
@@ -19,6 +19,7 @@
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
```
vibe13, line 22: I would also modify the unique constraint to be the same? systemImageId and systemImageRepositoryUrl

model/src/main/java/org/jboss/pnc/model/Product.java
```diff
@@ -208,6 +208,22 @@ public boolean addVersion(ProductVersion version) {
         return productVersions.add(version);
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o)
+            return true;
+        if (!(o instanceof Product))
+            return false;
+        return id != null && id.equals(((BuildConfigSetRecord) o).getId());
```
vibe13, line 217: `return id != null && id.equals(((Product) o).getId());`

vibe13 (changes requested):
Hi @janinko big +1 to the above changes. I saw the same guides some weeks ago when enabling the caches and I decided to park them and wait sometime after the 2.x release to get back to them. Apart from some fixes to do, I am +1. Thanks!!

model/src/main/java/org/jboss/pnc/model/BuildEnvironment.java
janinko: I don't know.
The `systemImageId` is already unique so paired with anything (e.g. `systemImageRepositoryUrl`) it's still unique. At first I wanted to use only the `systemImageId` in the equals thinking that `systemImageRepositoryUrl` will be added later when the BuildEnvironment is refactored. But then I thought that it is not a problem to add it now and we won't forget to add it later. But I didn't plan to do any DB changes in this PR.
If you think new unique constraint should be added (and the one from `systemImageId` removed?) I'll do it.

vibe13: Yes it was more to catch 2 birds with 1 stone while you are making such changes. We might use the same systemimageID with different systemimagerepositoryurl soonish (for Quay). I am ok in making DB changes when they will be needed, it's just a bit strange that the code and DB have different uniqueness though.

vibe13 (approved):

janinko:
retest this please

PR #3344, opened 2020-10-12
[NCL-6158]: Add 'Brew pull active' flag on build config

jbartece:
### Checklist:

### DIFF
--- a/dto/src/main/java/org/jboss/pnc/dto/BuildConfiguration.java
+++ b/dto/src/main/java/org/jboss/pnc/dto/BuildConfiguration.java
@@ -126,7 +126,7 @@ public class BuildConfiguration extends BuildConfigurationRef {
             User creationUser,
             User modificationUser,
             String defaultAlignmentParams,
-            boolean brewPullActive) {
+            Boolean brewPullActive) {
         super(
                 id,
                 name,
--- a/dto/src/main/java/org/jboss/pnc/dto/BuildConfiguration.java
+++ b/dto/src/main/java/org/jboss/pnc/dto/BuildConfiguration.java
@@ -125,7 +125,8 @@ public class BuildConfiguration extends BuildConfigurationRef {
             BuildType buildType,
             User creationUser,
             User modificationUser,
-            String defaultAlignmentParams) {
+            String defaultAlignmentParams,
+            boolean brewPullActive) {
         super(
                 id,
                 name,
@@ -135,7 +136,8 @@ public class BuildConfiguration extends BuildConfigurationRef {
                 creationTime,
                 modificationTime,
                 buildType,
-                defaultAlignmentParams);
+                defaultAlignmentParams,
+                brewPullActive);
         this.scmRepository = scmRepository;
         this.project = project;
         this.environment = environment;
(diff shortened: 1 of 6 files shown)


dto/src/main/java/org/jboss/pnc/dto/BuildConfigurationRef.java
```diff
@@ -104,6 +104,11 @@
      */
     protected final String defaultAlignmentParams;
 
+    /**
+     * Indicates whether the Brew Bridge Pull feature is active or not. It defaults to false.
+     */
+    protected boolean brewPullActive = false;
```
janinko, line 110: Not final. You will have to solve the default differently.

Note: @JsonProperty(defaultValue="false") is only for **ducumenting** default value, not setting it.

janinko (changes requested):

dto/src/main/java/org/jboss/pnc/dto/BuildConfigurationRef.java
pkocandr: Since false is a default for a boolean field, is it possible to drop the "=false" and rely on the java default?

dto/src/main/java/org/jboss/pnc/dto/BuildConfiguration.java
```diff
@@ -125,7 +125,8 @@ private BuildConfiguration(
             BuildType buildType,
             User creationUser,
             User modificationUser,
-            String defaultAlignmentParams) {
+            String defaultAlignmentParams,
+            boolean brewPullActive) {
```
janinko, line 129:
```suggestion
            Boolean brewPullActive) {
```

dto/src/main/java/org/jboss/pnc/dto/BuildConfigurationRef.java
```diff
@@ -104,6 +104,11 @@
      */
     protected final String defaultAlignmentParams;
 
+    /**
+     * Indicates whether the Brew Bridge Pull feature is active or not. It defaults to false.
+     */
+    protected Boolean brewPullActive;
```
janinko, line 110: neni final

mapper/src/main/java/org/jboss/pnc/mapper/api/BuildConfigurationMapper.java
```diff
@@ -87,6 +88,7 @@ default BuildConfiguration toIDEntity(BuildConfigurationRef dtoEntity) {
     @Mapping(target = "parameters", source = "genericParameters")
     @Mapping(target = "creationUser", qualifiedBy = Reference.class)
     @Mapping(target = "modificationUser", source = "lastModificationUser", qualifiedBy = Reference.class)
+    @Mapping(target = "brewPullActive", source = "brewPullActive")
```
janinko, line 91: this shouldn't be required if they have the same name

janinko (changes requested):

mapper/src/main/java/org/jboss/pnc/mapper/api/BuildConfigurationMapper.java
jbartece: OK

janinko (approved):

PR #3347, opened 2020-10-13
[NCL-6163] New endpoint on Artifact with direct filtering for DA

dominikbrazdil:
### Checklist:

### DIFF
--- a/dto/src/main/java/org/jboss/pnc/dto/response/ArtifactInfo.java
+++ b/dto/src/main/java/org/jboss/pnc/dto/response/ArtifactInfo.java
@@ -1,0 +1,66 @@
+/**
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014-2020 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.pnc.dto.response;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import lombok.Builder;
+import lombok.Data;
+import org.jboss.pnc.enums.ArtifactQuality;
+import org.jboss.pnc.enums.RepositoryType;
+
+/**
+ * Really basic Artifact info for optimized queries.
+ *
+ * @author Dominik Brázdil &lt;dbrazdil@redhat.com&gt;
+ */
+@Data
+@Builder(builderClassName = "Builder")
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonDeserialize(builder = ArtifactInfo.Builder.class)
+public class ArtifactInfo {
+
+    /**
+     * ID of the artifact.
+     */
+    protected final String id;
+
+    /**
+     * A unique identifier of the artifact in a repository. For example, for a maven artifact this is the GATVC
+     * (groupId:artifactId:type:version[:classifier] The format of the identifier string is determined by the repository
+     * type.
+     */
+    protected final String identifier;
+
+    /**
+     * Quality level of the artifact.
+     */
+    protected final ArtifactQuality artifactQuality;
+
+    /**
+     * The type of repository which hosts this artifact (Maven, NPM, etc). This field determines the format of the
+     * identifier string.
+     */
+    protected final RepositoryType repositoryType;
+
+    @JsonPOJOBuilder(withPrefix = "")
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    public static final class Builder {
+    }
+}
--- a/facade/src/main/java/org/jboss/pnc/facade/providers/ArtifactProviderImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/providers/ArtifactProviderImpl.java
@@ -17,37 +17,12 @@
  */
 package org.jboss.pnc.facade.providers;
 
-import static org.jboss.pnc.common.util.StreamHelper.nullableStreamOf;
-import static org.jboss.pnc.facade.providers.api.UserRoles.SYSTEM_USER;
-import static org.jboss.pnc.spi.datastore.predicates.ArtifactPredicates.withBuildRecordId;
-import static org.jboss.pnc.spi.datastore.predicates.ArtifactPredicates.withDependantBuildRecordId;
-import static org.jboss.pnc.spi.datastore.predicates.ArtifactPredicates.withIdentifierLike;
-import static org.jboss.pnc.spi.datastore.predicates.ArtifactPredicates.withMd5;
-import static org.jboss.pnc.spi.datastore.predicates.ArtifactPredicates.withQualityIn;
-import static org.jboss.pnc.spi.datastore.predicates.ArtifactPredicates.withRepositoryType;
-import static org.jboss.pnc.spi.datastore.predicates.ArtifactPredicates.withSha1;
-import static org.jboss.pnc.spi.datastore.predicates.ArtifactPredicates.withSha256;
-
-import java.time.Instant;
-import java.util.EnumSet;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
-import java.util.stream.Collectors;
-
-import javax.annotation.security.DenyAll;
-import javax.annotation.security.PermitAll;
-import javax.annotation.security.RolesAllowed;
-import javax.ejb.Stateless;
-import javax.inject.Inject;
-
-import org.commonjava.atlas.npm.ident.ref.NpmPackageRef;
-import org.jboss.pnc.causewayclient.remotespi.NpmBuiltArtifact;
 import org.jboss.pnc.common.maven.Gav;
 import org.jboss.pnc.coordinator.maintenance.BlacklistAsyncInvoker;
 import org.jboss.pnc.dto.ArtifactRef;
 import org.jboss.pnc.dto.ArtifactRevision;
 import org.jboss.pnc.dto.User;
+import org.jboss.pnc.dto.response.ArtifactInfo;
 import org.jboss.pnc.dto.response.Page;
 import org.jboss.pnc.enums.ArtifactQuality;
 import org.jboss.pnc.enums.RepositoryType;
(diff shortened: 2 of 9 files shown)


rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ArtifactEndpoint.java
```diff
@@ -113,6 +115,43 @@
...
+                            responseCode = INVALID_CODE,
+                            description = INVALID_DESCRIPTION,
+                            content = @Content(schema = @Schema(implementation = ErrorResponse.class))),
+                    @ApiResponse(
+                            responseCode = SERVER_ERROR_CODE,
+                            description = SERVER_ERROR_DESCRIPTION,
+                            content = @Content(schema = @Schema(implementation = ErrorResponse.class))) })
+    @GET
+    @Path("/filter")
+    Page<Artifact> getAllFiltered(
```
michalovjan, line 149: What I don't get is why won't we just use RSQL on getArtifacts endpoint instead of creating entire new endpoint. DA needs to prepare request anyway.

RSQL that should be equivalent example is: q=identifier=like=group:*:version;artifactQuality=in=(DELETED,BLACKLISTED,NEW...);targetRepository.repositoryType==MAVEN
 
If you don't want to use =in= you can group multiple ORs like this: q=identifier=like=group:*:version;(artifactQuality==NEW,artifactQuality==TEMPORARY...);targetRepository.repositoryType==NPM 

RSQL will create just one query to DB.

janinko: I have other question. Is it necessary for the endpoint to return whole Artifact? Isn't just subset of the data needed? For example, does DA need Build or TargetRepository?
I belive this endpoint should be really optimized.

dominikbrazdil: @michalovjan  I believe that the new endpoint was suggested because of optimization as Petr describes in his doc https://docs.engineering.redhat.com/display/JP/Design%3A+Limit+usage+of+Brew+pull -> (to avoid frequent rsql parsing). @pkocandr is that right? 

@janinko I am afraid I do not know what exactly DA needs from the Artifact. I can discuss this with Petr in more detail to crop the unnecessary data.

pkocandr: @dominikbrazdil Yes, the concern was raised by @jbartece and it will definitely help.
@dominikbrazdil AFAIK DA currently needs only list of versions and in the future it will need also the artifact quality, so I would add it right away. The result would be just a map version -> quality.  @janinko Correct?

jbartece: This endpoint should be optimized as much as possible. Using the generic endpoint with RSQL is not a good option. It will be heavily used. 
I second Honza's point, if it needs to return the whole artifact. ArtifactRef should be sufficient or possibly we could even create a new DTO entity.

michalovjan: I still think that RSQL parsing is not a performance issue but with Honza's suggestion, there is need for a new endpoint anyway.

facade/src/main/java/org/jboss/pnc/facade/providers/ArtifactProviderImpl.java
```diff
@@ -288,6 +321,41 @@ public ArtifactRevision getRevision(String id, Integer rev) {
...
+
+        CriteriaQuery<Tuple> query = cb.createTupleQuery();
+
+        Root<Artifact> artifact = query.from(org.jboss.pnc.model.Artifact.class);
+        query.multiselect(
+                artifact.get(Artifact_.id),
+                artifact.get(Artifact_.identifier),
+                artifact.get(Artifact_.artifactQuality));
+        Predicate withIdentifierLike = identifierPattern.isPresent()
+                ? cb.like(artifact.get(Artifact_.identifier), "%" + identifierPattern.get().replace("*", "%") + "%")
```
michalovjan, line 338: Is there a particular reason to surround the identifier with '%'? 

If you'd asked for "org.jboss.pnc:pnc:jar:2.0.0.GA-redhat-00001" specifically. You could potentially get 3 artifacts (or more): 
- org.jboss.pnc:pnc:jar:2.0.0.GA-redhat-00001
- org.jboss.pnc:pnc:jar:2.0.0.GA-redhat-00001:sources
- org.jboss.pnc:pnc:jar:2.0.0.GA-redhat-00001:javadocs

rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ArtifactEndpoint.java
```diff
@@ -113,6 +116,43 @@
...
+     * @param repoType {@value FILTER_REPOSITORY_TYPE_DESC}
+     * @return
+     */
+    @Operation(
+            summary = GET_ALL_FILTERED_DESC,
+            responses = {
+                    @ApiResponse(
+                            responseCode = SUCCESS_CODE,
+                            description = SUCCESS_DESCRIPTION,
+                            content = @Content(schema = @Schema(implementation = ArtifactPage.class))),
```
michalovjan, line 139: ArtifactInfoPage.class?

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
```diff
@@ -18,12 +18,16 @@
 package org.jboss.pnc.spi.datastore.predicates;
 
 import org.apache.commons.collections.CollectionUtils;
+import org.jboss.pnc.enums.ArtifactQuality;
+import org.jboss.pnc.enums.RepositoryType;
```
michalovjan, line 22: left imports

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
```diff
@@ -18,12 +18,16 @@
...
 import org.jboss.pnc.model.Artifact;
 import org.jboss.pnc.model.Artifact_;
 import org.jboss.pnc.model.BuildRecord;
 import org.jboss.pnc.model.BuildRecord_;
 import org.jboss.pnc.model.ProductMilestone;
 import org.jboss.pnc.model.ProductMilestone_;
+import org.jboss.pnc.model.TargetRepository;
+import org.jboss.pnc.model.TargetRepository_;
```
michalovjan, line 30: left imports

michalovjan (changes requested):

facade/src/main/java/org/jboss/pnc/facade/providers/ArtifactProviderImpl.java
dominikbrazdil: Well I thought that if the endpoint input is the identifier "pattern" then it should behave as LIKE functionality all the time. But if the case is more that it will be used for specific version calls then it should be probably based only on "*" present in the pattern.

facade/src/main/java/org/jboss/pnc/facade/providers/ArtifactProviderImpl.java
```diff
@@ -288,6 +321,41 @@ public ArtifactRevision getRevision(String id, Integer rev) {
...
+                : cb.and();
+        Predicate withQualityIn = qualities.isPresent() && !qualities.get().isEmpty()
+                ? artifact.get(Artifact_.ARTIFACT_QUALITY).in(qualities.get())
+                : cb.and();
+        Predicate withRepoType = repoType.isPresent() ? cb
+                .equal(artifact.join(Artifact_.targetRepository).get(TargetRepository_.REPOSITORY_TYPE), repoType.get())
+                : cb.and();
+        query.where(cb.and(withIdentifierLike, withQualityIn, withRepoType));
+        query.orderBy(cb.desc(artifact.get(Artifact_.id)));
```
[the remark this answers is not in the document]
dominikbrazdil: Alright, didn't know that.

dto/src/main/java/org/jboss/pnc/dto/response/ArtifactInfo.java
```diff
@@ -0,0 +1,59 @@
...
+@Data
+@Builder(builderClassName = "Builder")
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonDeserialize(builder = ArtifactInfo.Builder.class)
+public class ArtifactInfo {
+
+    /**
+     * ID of the artifact.
+     */
+    protected final String id;
```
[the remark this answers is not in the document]
janinko: DA doesn't necessary need it, but I don't see any harm in including it.

dto/src/main/java/org/jboss/pnc/dto/response/ArtifactInfo.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * A unique identifier of the artifact in a repository. For example, for a maven artifact this is the GATVC
+     * (groupId:artifactId:type:version[:classifier] The format of the identifier string is determined by the repository
+     * type.
+     */
+    protected final String identifier;
```
janinko, line 48: Do we have artifact type (MVN/NPM) easily available? I guess we know it from the context, but I think that we could have it here so we know how to parse the identifier.

facade/src/main/java/org/jboss/pnc/facade/providers/ArtifactProviderImpl.java
janinko: The queries for the identifier MUST NEVER have "%" at the begining, so we can utilize index prefix search in DB.

dominikbrazdil: OK, so I will leave only the 'replace' part and DA will specify by "*" whenever it will need to search all identifiers that match the pattern.

dto/src/main/java/org/jboss/pnc/dto/response/ArtifactInfo.java
dominikbrazdil: I can add it easily

michalovjan (approved):

rest/src/main/java/org/jboss/pnc/rest/endpoints/ArtifactEndpointImpl.java
```diff
@@ -75,6 +79,20 @@ public void init() {
...
+    public Page<ArtifactInfo> getAllFiltered(
+            PaginationParameters paginationParameters,
+            String identifier,
+            Set<ArtifactQuality> qualities,
+            RepositoryType repoType) {
+        return artifactProvider.getAllFiltered(
+                paginationParameters.getPageIndex(),
+                paginationParameters.getPageSize(),
+                Optional.ofNullable(identifier),
+                Optional.ofNullable(qualities),
```
janinko, line 92: No need to wrap set to Optional:
> If `@DefaultValue` is not used and the corresponding meta-data is not present in the request, the value will be an empty collection for List, Set or SortedSet, null for other object types, and the Java-defined default for primitive types.
https://docs.oracle.com/javaee/7/api/javax/ws/rs/DefaultValue.html

rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ArtifactEndpoint.java
```diff
@@ -113,6 +117,43 @@
...
+                            responseCode = SERVER_ERROR_CODE,
+                            description = SERVER_ERROR_DESCRIPTION,
+                            content = @Content(schema = @Schema(implementation = ErrorResponse.class))) })
+    @GET
+    @Path("/filter")
+    Page<ArtifactInfo> getAllFiltered(
+            @Valid @BeanParam PaginationParameters paginationParameters,
+            @Parameter(description = FILTER_IDENTIFIER_DESC) @QueryParam("identifier") String identifier,
+            @Parameter(description = FILTER_QUALITY_DESC) @QueryParam("qualities") Set<ArtifactQuality> qualities,
+            @Parameter(description = FILTER_REPOSITORY_TYPE_DESC) @QueryParam("repoType") RepositoryType repoType);
```
janinko, line 155: Why are all the parameters optional? Wouldn't we always want to specify at least the identifier? And I would say repoType too.

dominikbrazdil: @pkocandr What parameters will be used at all times?

rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ArtifactEndpoint.java
jbartece: It is definitely worse than without RSQL, so there is no point in parsing the string 100 000+ times a day.

rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ArtifactEndpoint.java
jbartece: @janinko Why to make some parameters optional? It is filtering, so it is up to the user/use-case. I'd keep all optional. It doesn't hurt and makes the usage more flexible.

janinko: because of KISS

dominikbrazdil: I think that the Optionals do not make it somehow unreadable and the endpoint is more flexible this way. But if it sure that it won't be used any other way then it wouldn't make much sense to keep it optional.

jbartece: @janinko If it is optional, it is simple, isn't it? :-)

pkocandr: For the version list read by DA we will specify all the params everytime, but for a different use-case in the future I would not limit it by making the params mandatory.

dominikbrazdil: Ok, I will leave them optional

jbartece:
@dominikbrazdil Don't forget to update Changelog https://github.com/project-ncl/pnc/wiki/Changelog

janinko (approved):

PR #3348, opened 2020-10-14
[NCL-6128]: Add LIKE support to filter builds by BuildConfig.name

jbartece:
### Checklist:

### DIFF
--- a/datastore/src/main/java/org/jboss/pnc/datastore/repositories/BuildConfigurationAuditedRepositoryImpl.java
+++ b/datastore/src/main/java/org/jboss/pnc/datastore/repositories/BuildConfigurationAuditedRepositoryImpl.java
@@ -192,6 +192,8 @@ public class BuildConfigurationAuditedRepositoryImpl implements BuildConfigurati
 
     @Override
     public List<BuildConfigurationAudited> searchForBuildConfigurationName(String buildConfigurationName) {
+        buildConfigurationName = buildConfigurationName.replaceAll("\\*", "%");
+
         List<Object[]> result = AuditReaderFactory.get(entityManager)
                 .createQuery()
                 .forRevisionsOfEntity(BuildConfiguration.class, false, false)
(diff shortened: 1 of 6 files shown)


datastore/src/main/java/org/jboss/pnc/datastore/repositories/BuildConfigurationAuditedRepositoryImpl.java
```diff
@@ -203,6 +205,8 @@ public BuildConfigurationAudited queryById(IdRev idRev) {
 
     @Override
     public List<IdRev> searchIdRevForBuildConfigurationName(String buildConfigurationName) {
+        buildConfigurationName = buildConfigurationName.replaceAll("\\*", "%");
```
janinko, line 208: shouldn't this be in the provider? Here I think it's in too deep level.

facade/src/test/java/org/jboss/pnc/facade/providers/BuildProviderImplTest.java
```diff
@@ -299,6 +300,44 @@ public void testFilterRunningBuildsByBuildConfigName() {
         assertEquals(BuildMapper.idMapper.toDto(givenBT.getId()), it.next().getId());
     }
 
+    @Test
+    public void testFilterLikeRunningBuildsByBuildConfigName() {
+        // Given
+        mockBuildRecord();
+        mockBuildTask();
+
+        String givenBcName = "VeryLongAndComplicatedBcName";
```
janinko, line 309: You missed an opportunity to name it "VeryLongAndHardToRememberBcName" :smile:

datastore/src/main/java/org/jboss/pnc/datastore/repositories/BuildConfigurationAuditedRepositoryImpl.java
jbartece: I think it could be in both. I prefered it here as * is more common than % and wanted to hide the logic. It is described clearly in the javadoc

facade/src/test/java/org/jboss/pnc/facade/providers/BuildProviderImplTest.java
jbartece: :-)

janinko (approved):

PR #3369, opened 2020-10-26
[NCL-5709]Implement negation in search filters

DnsZhou:
### Checklist:

### DIFF
--- a/ui/app/common/pnc-client/rsql/comparator.js
+++ b/ui/app/common/pnc-client/rsql/comparator.js
@@ -71,6 +71,13 @@
           return ctx.next();
         };
 
+        that.notlike = function (value) {
+          value = value.replace(/\*/g, '%');
+          value = value.replace(/\?/g, '_');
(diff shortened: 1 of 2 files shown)


ui/app/common/pnc-client/rsql/comparator.js
```diff
@@ -71,6 +71,13 @@
           return ctx.next();
         };
 
+        that.notlike = function (value) {
+          value = value.replace(/\*/g, '%');
```
alexcreasy, line 75: Just a thought these replace lines should have been removed in 2.0 ... the rest API should now accept '*' and '?' natively. That said they could be removed in a different issue

alexcreasy (approved):

ui/app/common/pnc-client/rsql/comparator.js
DnsZhou: Got it, I will create another jira and fix it in next sprint.

PR #3373, opened 2020-10-29
[NCL-6011] Fire product milestone events

jomrazek:
### Checklist:

### DIFF
--- a/bpm/src/main/java/org/jboss/pnc/bpm/causeway/ProductMilestoneReleaseManager.java
+++ b/bpm/src/main/java/org/jboss/pnc/bpm/causeway/ProductMilestoneReleaseManager.java
@@ -25,8 +25,10 @@ import org.jboss.pnc.bpm.model.causeway.BuildImportResultRest;
 import org.jboss.pnc.bpm.model.causeway.BuildImportStatus;
 import org.jboss.pnc.bpm.model.causeway.MilestoneReleaseResultRest;
 import org.jboss.pnc.bpm.task.MilestoneReleaseTask;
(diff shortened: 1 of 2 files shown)


bpm/src/main/java/org/jboss/pnc/bpm/causeway/ProductMilestoneReleaseManager.java
```diff
@@ -187,6 +199,8 @@ private void storeResult(ProductMilestone milestone, MilestoneReleaseResultRest
             milestoneRepository.save(milestone);
 
             removeCurrentFlagFromMilestone(milestone);
+
+            productMilestoneCloseResultEvent.fire(mapper.toDTO(productMilestoneRelease));
```
michalovjan, line 203: What about when milestone closing fails? We definitely need to send a notification in that case.

michalovjan (changes requested):

bpm/src/main/java/org/jboss/pnc/bpm/causeway/ProductMilestoneReleaseManager.java
michalovjan: My point still stands. You are firing an event only when `result.getReleaseStatus().getMilestoneReleaseStatus() == MilestoneCloseStatus.SUCCEEDED`, which means, only when the closing succeeded. You don't take care of other cases, (f.e. exception thrown, FAILED result...) and users need to be notified.

michalovjan (approved):

PR #3385, opened 2020-11-02
[v2.0.x] [NCL-6199] fix NPE, eliminate cross joins, fix no results returned

michalovjan:
### Checklist:


michalovjan:
1. NPE is thrown because Java Streams will throw NPE if they have a null in the stream and you call findFirst() method.
2. The queries that were generated by this endpoint used CARTESIAN(CROSS) JOIN with WHERE instead of INNER JOIN so I fixed that.
3. The 'no content returned' bug happened when the ProductMilestone didn't have a ProductRelease. I've added a JoinType.Left so that the MilestoneInfo will be returned by the query.

janinko (approved):

PR #3388, opened 2020-11-04
Build Agent use HTTP callback instead of WS.

matejonnet:
### Checklist:


rest/src/main/java/org/jboss/pnc/rest/endpoints/internal/api/BuildExecutionEndpoint.java
```diff
@@ -0,0 +1,50 @@
...
+import static org.jboss.pnc.rest.configuration.SwaggerConstants.SUCCESS_DESCRIPTION;
+
+@Tag(name = "Internal")
+@Path("/build-execution")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public interface BuildExecutionEndpoint {
+
+    @Operation(
+            summary = "Used by Build Agent to nitify completion .",
```
janinko, line 43: notify

PR #3389, opened 2020-11-04
[NCL-6206]Fix Unhandled rejection when click cancel on edit GC and BC…

DnsZhou:
… popup

### Checklist:


DnsZhou:
Hi @alexcreasy , I updated this service created by you, because if we do not cope with the rejection, there will be a console error because when user clicks cancel button the rejection can not be found. Feel free to ping me if there is any problem with this. Thanks!

alexcreasy (approved):

PR #3428, opened 2020-12-04
[NCLSUP-217] Improvements to internal url check for builder pod creation

thescouser89:
## First commit
This PR fixes an issue where the internal url may not be properly
constructed at the beginning of the
`OpenshiftStartedEnvironment::monitorInitialization` method. When the
method is called, we don't know yet if the service object has been
created with a proper cluster IP yet. The url might look like this in
those cases: 'http:/pnc-ba-1234/', which strangely the 'URL' object
doesn't complain.

This PR tries to get the latest internal url everytime it checks if
the internal url is online.

This PR also sets a read timeout of 2 seconds so that we do not wait
forever for data to be sent back.

## Second commit
The `isBuildAgentFuture` completablefuture checks whether the internal
url of the service can be used to talk to the build agent server. This
establishes the readiness of the pod. The same url is used later on when
uploading the build script to PNC.

Unfortunately, due to Java's confusing completable future chain methods,
the wrong method 'thenApply' is used instead of `thenCompose`.

`thenApply` expects a lambda returning a 'regular' value and
ignores whether the value is a completable future before passing it to the next
chain. The result of `isBuildAgentFuture` is therefore lost in the next
chain.

`thenCompose` expects to receive a lambda returning a completable
future, waits for that completable future to finish, before moving on to
the next chain.

From: https://stackoverflow.com/a/43025751/2907906

### Checklist:

### DIFF
--- a/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
+++ b/openshift-environment-driver/src/main/java/org/jboss/pnc/environment/openshift/OpenshiftStartedEnvironment.java
@@ -413,14 +413,6 @@ public class OpenshiftStartedEnvironment implements StartedEnvironment {
             Consumer<RunningEnvironment> onComplete,
             Consumer<Exception> onError,
             int retries) {
-        URL buildAgentUrl;
-        try {
-            buildAgentUrl = new URL(getInternalEndpointUrl());
-        } catch (MalformedURLException e) {
-            onError.accept(e);
-            return;
-        }
-
         cancelHook = () -> onComplete.accept(null);
 
         CompletableFuture<Void> podFuture = creatingPod.thenComposeAsync(nul -> {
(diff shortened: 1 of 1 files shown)


vibe13 (approved):
@thescouser89 those are nice improvements indeed! Great to check the buildAgentURL with the latest calculated internal URL, I would have never thought of a null clusterIP, but surely makes sense to rely on the latest value we are given. The thenApply/theCompose fix I would have never caught that...but it's perfect. I am not sure I have worsened the things with my changes (using allOfOrException instead of CompletableFuture.allOf) when trying to fail as soon an exception was thrown instead of waiting for the timeouts anyway (which would have given the time to isBuildAgentUpFuture to finish) :-/. But surely a great spot. 
Anyway, I would try out and deploy these changes asap, they would surely improve our current situation :-/

thescouser89:
Thanks Andrea for the review! :) I will backport this PR also to PNC 2.0.x branch. Next step is to add retries to the upload script and PR!

PR #3429, opened 2020-12-04
[NCLSUP-217] Add retries on upload/download function for termd

thescouser89:
### Checklist:

### DIFF
--- a/termd-build-driver/src/main/java/org/jboss/pnc/termdbuilddriver/transfer/DefaultFileTranser.java
+++ b/termd-build-driver/src/main/java/org/jboss/pnc/termdbuilddriver/transfer/DefaultFileTranser.java
@@ -34,6 +34,14 @@ import java.util.function.Consumer;
 
 public class DefaultFileTranser implements FileTranser {
 
+    /**
+     * Stupid functional interface to encapsulate upload / download function. Used for the retry method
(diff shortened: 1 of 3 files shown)


termd-build-driver/src/main/java/org/jboss/pnc/termdbuilddriver/transfer/DefaultFileTranser.java
```diff
@@ -152,4 +181,64 @@ public void setConnectTimeout(int connectTimeout) {
...
+        for (int i = 1; i <= httpRetryMaxAttempts; i++) {
+
+            try {
+                return myRunnable.run();
+            } catch (TransferException e) {
+                // if we have reached the max number of attempts, give up and throw the transfer exception
+                if (i == httpRetryMaxAttempts) {
+                    throw e;
+                } else {
+                    logger.warn("{} failed after {} attempts. Sleeping...", taskName, 1, e);
```
vibe13, line 235: It probably should be "i" instead of "1"

thescouser89: Done! Thanks for spotting this! Haa silly bug :)

PR #3432, opened 2020-12-06
[NCL-6274] - Update UI for the Scala support

vibe13:
### Checklist:


vibe13:
@alexcreasy this is the only place in UI I spotted, based also on previous changes for other build types. wdyt? thanks!

alexcreasy:
@vibe13 yes, that's all that should be required to add scala support to the UI! :)

PR #3439, opened 2020-12-11
Always use new BPM connector - fix executor response.

matejonnet:
### Checklist:

### DIFF
--- a/bpm/src/main/java/org/jboss/pnc/bpm/ConnectorSelector.java
+++ b/bpm/src/main/java/org/jboss/pnc/bpm/ConnectorSelector.java
@@ -35,11 +35,13 @@ public class ConnectorSelector {
             Map<String, String> genericParameters = buildTask.getBuildTask()
                     .getBuildConfigurationAudited()
                     .getGenericParameters();
-            if (true || genericParameters.getOrDefault(GENERIC_PARAMETER_KEY, "").equals(RHPAM)) {
-                return true;
-            }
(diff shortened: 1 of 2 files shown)


bpm/src/main/java/org/jboss/pnc/bpm/ConnectorSelector.java
```diff
@@ -35,9 +35,14 @@ public static boolean useNewProcess(BpmTask task) {
...
-                return true;
-            }
+            return useNewProcessForBuild(genericParameters);
+        }
+        return false;
+    }
+
+    public static boolean useNewProcessForBuild(Map<String, String> genericParameters) {
+        if (true || genericParameters.getOrDefault(GENERIC_PARAMETER_KEY, "").equals(RHPAM)) {
+            return true;
```
janinko, line 45: you don't need if, you can just return the value of the condition;

janinko: BTW there is plan to change it in the future to not always return true?

janinko (changes requested):

bpm/src/main/java/org/jboss/pnc/bpm/ConnectorSelector.java
matejonnet: "true or" is there for easy roll-back if we need it, at some point we have to clean it up.

matejonnet:
Re-test this please!

PR #3463, opened 2021-01-26
Return relevant error in case of BuildConflictException

pkocandr:
Error handling enhancement. When a build trigger failed because of a
conflicting build already running, it got thrown as a RuntimeException.
Switched that to AlreadyRunningException with conflicting task ID as
the response object.

### Checklist:


thescouser89 (approved):
lgtm

PR #3470, opened 2021-02-09
Migrate to fabric8 client instead of openshift-restclient-java

thescouser89:
This is done to support the latest version of OCP 4.6, which
unfortunately the version we used for openshift-restclient-java does not
support anymore. Since the latter is also deprecated and fabric8 is the
recommended library, let's use this instead

### Checklist:


thescouser89:
Gotta fix that test!

thescouser89:
This is for NCL-6312

PR #3477, opened 2021-02-25
[NCL-6415] Support RSQL ManyToMany filtering

dominikbrazdil:
### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/providers/GenericSettingProvider.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/providers/GenericSettingProvider.java
@@ -117,7 +117,6 @@ public class GenericSettingProvider {
         GenericSetting pncVersion = createGenericParameterIfNotFound(PNC_VERSION);
         pncVersion.setValue(version);
         genericSettingRepository.save(pncVersion);
-        notifier.sendMessage(GenericSettingNotification.newAnnoucement(version));
     }
 
     public String getPNCVersion() {
--- a/facade/src/main/java/org/jboss/pnc/facade/rsql/mapper/AbstractRSQLMapper.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/rsql/mapper/AbstractRSQLMapper.java
@@ -26,6 +26,7 @@ import org.jboss.pnc.model.GenericEntity;
 import javax.inject.Inject;
 import javax.persistence.criteria.From;
 import javax.persistence.criteria.Path;
+import javax.persistence.metamodel.SetAttribute;
 import javax.persistence.metamodel.SingularAttribute;
 import java.io.Serializable;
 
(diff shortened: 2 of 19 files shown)


facade/src/main/java/org/jboss/pnc/facade/providers/GenericSettingProvider.java
```diff
@@ -117,7 +117,6 @@ public void setPNCVersion(String version) {
         GenericSetting pncVersion = createGenericParameterIfNotFound(PNC_VERSION);
         pncVersion.setValue(version);
         genericSettingRepository.save(pncVersion);
-        notifier.sendMessage(GenericSettingNotification.newAnnoucement(version));
```
janinko, line 120: why removing this?

facade/src/main/java/org/jboss/pnc/facade/rsql/mapper/AbstractRSQLMapper.java
```diff
@@ -65,6 +68,13 @@ public AbstractRSQLMapper(Class<DB> type) {
                 return mapEntity(from, toEntity(name), selector.next());
             }
         }
+        if (toEntitySet(name) != null) {
+            if (selector.isFinal()) {
+                return (PluralAttributePath<?>) from.get(toEntitySet(name));
```
janinko, line 73: Why are you casting it to `PluralAttributePath` when the return type is only `Path`?

facade/src/main/java/org/jboss/pnc/facade/providers/GenericSettingProvider.java
dominikbrazdil: This was an error from my previous PR which a noticed radnomly now and did not want to create a special PR for it.

facade/src/main/java/org/jboss/pnc/facade/rsql/mapper/AbstractRSQLMapper.java
dominikbrazdil: I wanted it to be more specific to make it clear that its for multiple attributes, and from.get returns Expression so some cast is need. You think that casting to Path<> would be better?

facade/src/main/java/org/jboss/pnc/facade/providers/GenericSettingProvider.java
janinko: can you put it in separate commit at least then?

facade/src/main/java/org/jboss/pnc/facade/rsql/mapper/AbstractRSQLMapper.java
janinko: yes, this is unnecessarily too specific.

facade/src/main/java/org/jboss/pnc/facade/providers/GenericSettingProvider.java
dominikbrazdil: Sure

janinko (approved):

PR #3495, opened 2021-04-08
[NCL-6162] Add metadata to Indy build group for brew pull

thescouser89:
When brew pull is active, set the metadata of the Indy build group
created accordingly so that it selects Brew. The contrary applies when
brew pull is inactive.

Since at the time of this commit, the metadata key is unknown, it is
configurable by the user in the 'indy repo driver' config with the
additional key: 'brew-pull-active-metadata-key'

Most of the changes relate to setting the metadata in
`RepositoryManager`, and making sure `DefaultBuildExecutor` properly sets
the brew pull info when creating a `RepositoryManager` object.

The rest of the changes is about propagating the new field in all the
test classes.

A new test is also created (`BuildGroupSetBrewPullActiveMetadataTest`) to
verify that:
1. the metadata key is set
2. the metadata is set to the proper value

### Checklist:


thescouser89:
@thescouser89 If merged, don't forget to add it to the sidecar

indy-repository-manager/src/test/java/org/jboss/pnc/indyrepositorymanager/AbstractRepositoryManagerDriverTest.java
```diff
@@ -53,6 +53,7 @@
     protected static final String CONFIG_SYSPROP = "pnc-config-file";
 
     protected static final String PNC_BUILDS = "pnc-builds";
+    protected static final String BREW_PULL_ACTIVE_METADATA_KEY = "koji-active";
```
pkocandr, line 56: According to the latest comments at MMENG-1262 the key will be "koji-pull". Not sure if it matters to the tests though.

pkocandr (approved):
Looks good!

indy-repository-manager/src/test/java/org/jboss/pnc/indyrepositorymanager/AbstractRepositoryManagerDriverTest.java
thescouser89: Thanks for letting me know! I updated the field to minimize confusion to future readers of this code.

PR #3496, opened 2021-04-08
[NCL-6456] Add helper method replaceHostInUrl

thescouser89:
This helper method replaces the host in a particular url with a new
host (including the new scheme and port). The new host is in the format:

- `<scheme>://<host>[:<port>]`

where 'port' is optional

If the default ports for http / https are used, then the port
information is omitted.

### Checklist:

### DIFF
--- a/common/src/main/java/org/jboss/pnc/common/util/UrlUtils.java
+++ b/common/src/main/java/org/jboss/pnc/common/util/UrlUtils.java
@@ -181,4 +181,39 @@ public final class UrlUtils {
         return (userInfo == null ? "" : userInfo + "@") + (host == null ? "" : host) + (port == -1 ? "" : ":" + port)
                 + (path == null ? "" : path) + queryAppend + fragmentAppend;
     }
+
+    /**
+     * Substitute the url with a new host (with new scheme and port), leaving the rest unchanged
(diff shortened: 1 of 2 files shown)


common/src/main/java/org/jboss/pnc/common/util/UrlUtils.java
```diff
@@ -181,4 +181,39 @@ public static String stripProtocol(String url) {
...
+        }
+
+        URL originalUrl = new URL(url);
+        URL newHostUrl = new URL(newHost);
+
+        boolean hostHasPort = newHost.contains(":");
+
+        int newPort = -1;
+        if (hostHasPort) {
+            newPort = newHostUrl.getPort();
```
vibe13, line 207: @thescouser89 I believe the newHostUrl.getPort() would already return -1 if the port is not set, so probably LN203-208 can be replaced just with 
int newPort = newHostUrl.getPort();
?

thescouser89: Thanks for catching this! I can confirm what you said haa. Good find! :)

PR: https://github.com/project-ncl/pnc/pull/3500

PR #3499, opened 2021-04-19
[NCL-6465] [NCL-6466]: Deprecate option to build temporary builds wit…

jbartece:
…h timestamp alignment

### Checklist:


spi/src/main/java/org/jboss/pnc/spi/BuildOptions.java
```diff
@@ -34,23 +34,25 @@
 @AllArgsConstructor
 @ToString
 @EqualsAndHashCode
-@Getter
```
janinko, line 37: Why have you removed it from here? Lombok doesn't generate getter when there is already one present

janinko (approved):

PR #3502, opened 2021-05-04
[NCLSUP-317] Prefer BREW_BUILD_NAME Build attribute to BCR parameter

janinko:
BREW_BUILD_NAME Build attribute (in fact implemented as
executionRootName field) is modifiable by users, BCR is immutable.

### Checklist:


matejonnet (approved):

janinko:
@janinko Change name to "Prefer BREW_BUILD_NAME from Build attribute rather then from BCR parameter"

PR #3511, opened 2021-05-19
Allow target repository without ID in BuildResult.

matejonnet:
### Checklist:


mapper/src/main/java/org/jboss/pnc/mapper/api/ArtifactMapper.java
```diff
@@ -68,8 +68,19 @@
     @BeanMapping(ignoreUnmappedSourceProperties = { "deployUrl", "publicUrl" })
     Artifact toEntity(org.jboss.pnc.dto.Artifact dtoEntity);
 
+    /**
+     * Created model.Artifact can contain target repository entity in transient (in case id is null)
+     * or unspecified (in case id is not null) JPA state.
+     *
+     * @param dtoEntity
+     * @return
```
janinko, line 76: Either fill or the `@param` and `@return` or delete them

PR #3513, opened 2021-05-25
Base32

janinko:
### Checklist:

The BuildRecord class now uses embedable id Base32LongID which is stored in DB as long, but is able to be constructed from both long and (base 32 encded) string and provides methods to read the ID as both long and string, with preference to the string form.

DB queries are using the new Base32LongID objects, which can be easily created from both String and Long objects.

Orchestrator now internally uses String form of the Build identifier - in Build Task, Build Execution, etc - and only when touching DB is the identifier used in the form of Base32LongID object.


janinko:
retest this please

matejonnet (approved):

vibe13:
@janinko massive work! +1 for me, thanks!

PR #3521, opened 2021-06-21
Add pattern index on artifact identifier

janinko:
We want to quickly search by prefix of an identifier for dependency
alignment and version increment. Adding this index should allow much
faster search.

### Checklist:


vibe13:
Our locale is:
$ show lc_collate;
 lc_collate 
------------
 en_US.utf8
(1 row)
so I believe we should stick to that unless there is a specific reason?

vibe13 (approved):

PR #3524, opened 2021-06-24
[NCL-6632] Use String IDs when communicating with Causeway

janinko:
### Checklist:


janinko:
retest this please

mapper/src/main/java/org/jboss/pnc/mapper/api/ArtifactMapper.java
```diff
@@ -18,6 +18,7 @@
 package org.jboss.pnc.mapper.api;
 
 import org.jboss.pnc.dto.ArtifactRef;
+import org.jboss.pnc.mapper.Base32LongIdMapper;
```
jomrazek, line 21: is this needed? I don't see it used

janinko: it's not, removed

jomrazek (approved):

PR #3525, opened 2021-06-28
[NCL-6653] Revert "Try to ignore nested reference chain bits which ca…

thescouser89:
…uses the NPE"

This reverts commit 3397a5b3adb26570f4726f14673b7b799b19e4ab.

This is because we switched to Fabric8 Openshift libraries instead of
the old one

### Checklist:


vibe13 (approved):
Nice catch, thanks @thescouser89 ! :-)

PR #3533, opened 2021-07-14
[NCL-6652] - Add purl field to artifacts metadata

vibe13:
### Checklist:

### DIFF
--- a/indy-repository-manager/src/main/java/org/jboss/pnc/indyrepositorymanager/IndyRepositorySession.java
+++ b/indy-repository-manager/src/main/java/org/jboss/pnc/indyrepositorymanager/IndyRepositorySession.java
@@ -54,6 +54,10 @@ import org.jboss.pnc.spi.repositorymanager.model.RepositorySession;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.github.packageurl.MalformedPackageURLException;
+import com.github.packageurl.PackageURL;
+import com.github.packageurl.PackageURLBuilder;
+
 import javax.validation.ConstraintViolation;
 import javax.validation.Validation;
 import javax.validation.Validator;
@@ -71,6 +75,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
+import static org.apache.commons.lang.StringUtils.isEmpty;
 import static org.commonjava.indy.model.core.GenericPackageTypeDescriptor.GENERIC_PKG_KEY;
 import static org.commonjava.indy.pkg.maven.model.MavenPackageTypeDescriptor.MAVEN_PKG_KEY;
 import static org.commonjava.indy.pkg.npm.model.NPMPackageTypeDescriptor.NPM_PKG_KEY;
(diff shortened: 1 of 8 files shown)


indy-repository-manager/src/main/java/org/jboss/pnc/indyrepositorymanager/IndyRepositorySession.java
```diff
@@ -710,6 +719,95 @@ private String computeIdentifier(final TrackedContentEntryDTO transfer) {
...
+                    if (pathInfo != null) {
+                        // See https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#maven
+                        PackageURLBuilder purlBuilder = PackageURLBuilder.aPackageURL()
+                                .withType(PackageURL.StandardTypes.MAVEN)
+                                .withNamespace(pathInfo.getProjectId().getGroupId())
+                                .withName(pathInfo.getProjectId().getArtifactId())
+                                .withVersion(pathInfo.getVersion())
+                                .withQualifier(
+                                        "type",
+                                        StringUtils.isEmpty(pathInfo.getType()) ? "jar" : pathInfo.getType());
```
pkocandr, line 745: This is probably an unnecessary default, because when you parse the ArtifactPathInfo from path, it always has a type. If it is not there, the parser is not able to complete and returns null.

indy-repository-manager/src/main/java/org/jboss/pnc/indyrepositorymanager/IndyRepositorySession.java
```diff
@@ -710,6 +719,95 @@ private String computeIdentifier(final TrackedContentEntryDTO transfer) {
...
+
+                    NpmPackagePathInfo npmPathInfo = NpmPackagePathInfo.parse(transfer.getPath());
+                    if (npmPathInfo != null) {
+                        // See https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#npm
+                        PackageURLBuilder purlBuilder = PackageURLBuilder.aPackageURL()
+                                .withType(PackageURL.StandardTypes.NPM)
+                                .withVersion(npmPathInfo.getVersion().toString());
+
+                        String[] scopeAndName = npmPathInfo.getName().split("/");
+                        if (scopeAndName != null && scopeAndName.length > 0) {
```
pkocandr, line 764: I think there is no way how scopeAndName could be null.

pkocandr (approved):
Looks good although there are some unnecessary checks.

indy-repository-manager/src/main/java/org/jboss/pnc/indyrepositorymanager/IndyRepositorySession.java
vibe13: Ah ok, I took the hint from SimpleTypeAndClassifier constructor (used by SimpleArtifactRef which is used in the creation of the identifier)

indy-repository-manager/src/main/java/org/jboss/pnc/indyrepositorymanager/IndyRepositorySession.java
vibe13: right, changing

vibe13:
retest this please

PR #3537, opened 2021-07-28
Switch to Indy client 2.6.0

pkocandr:
### Checklist:


pkocandr:
There is no need to switch to 2.6.0 in orch. It will be done in repository driver instead.

PR #3545, opened 2021-08-31
[NCLSUP-438] Try to get a new JMS connection on Exception

thescouser89:
This change is done especifically for exceptions when the JMS connection is
closed abruptly:
```
stackTrace: : org.jboss.pnc.messaging.spi.MessagingRuntimeException: Cannot send the message:
...
Caused by: javax.jms.IllegalStateException: The Connection is closed
        at org.apache.activemq.ra.ManagedConnectionProxy.getConnection(ManagedConnectionProxy.java:98)
        at org.apache.activemq.ra.ManagedConnectionProxy.createSessionProxy(ManagedConnectionProxy.java:124)
        at org.apache.activemq.ra.ManagedConnectionProxy.createSession(ManagedConnectionProxy.java:110)
        at org.jboss.pnc.messaging.DefaultMessageSender.doSendMessage(DefaultMessageSender.java:117)
```

On those failures, the `DefaultMessageSender` keeps failing to send data
afterwards since it doesn't try to establish a new connection. The only
way to recover from this is to restart the application server. This
commit attempts to re-establish a new connection on exceptions so that
it tries to recover without having to restart the application server.

Another alternative is to request a new connection on every sending of
data, but it seems that the connections are not pooled and it would
become very costly to request a new connection for each new data send.

### Checklist:

### DIFF
--- a/messaging/src/main/java/org/jboss/pnc/messaging/DefaultMessageSender.java
+++ b/messaging/src/main/java/org/jboss/pnc/messaging/DefaultMessageSender.java
@@ -17,6 +17,7 @@
  */
 package org.jboss.pnc.messaging;
 
+import lombok.extern.slf4j.Slf4j;
 import org.jboss.pnc.messaging.spi.Message;
 import org.jboss.pnc.messaging.spi.MessageSender;
 import org.jboss.pnc.messaging.spi.MessagingRuntimeException;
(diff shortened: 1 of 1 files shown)


messaging/src/main/java/org/jboss/pnc/messaging/DefaultMessageSender.java
```diff
@@ -118,6 +118,8 @@ protected void doSendMessage(String message, Map<String, String> headers) {
             messageProducer = session.createProducer(destination);
             sendUsingProducer(message, headers, session, messageProducer);
         } catch (Exception e) {
+            // try to set up a new connection on exception for the next message
+            init();
```
pkocandr, line 122: Shouldn't we retry after init? And throw the MessagingRuntimeException only if it fails again?

thescouser89: Yeah can do. I'll try to get that change in. Thanks for the suggestion!

thescouser89:
retest this please

PR #3576, opened 2021-12-08
Fix duplicate key value violates unique constraint `uk_targetrepo_identifier_repopath`

vibe13:


datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
```diff
@@ -246,7 +246,8 @@ private void fetchOrSaveRequiredTargetRepositories(
         for (Artifact artifact : artifacts) {
             TargetRepository targetRepository = artifact.getTargetRepository();
             logger.trace("Adding repository for artifact: {}.", artifact.toString());
-            if (!storedTargetRepositories.containsKey(targetRepository.getIdentifierPath())) {
+            if (!storedTargetRepositories.containsKey(targetRepository.getIdentifierPath())
+                    && !requiredTargetRepositories.containsKey(targetRepository.getIdentifierPath())) {
```
janinko, line 250: Why was this an issue? `requiredTargetRepositories` is a map so it doesn't contain the same key twice

janinko:
I think this PR can be closed

PR #3577, opened 2021-12-16
[NCL-6861] Don't overwrite existing Build when saving it

janinko:
When the /build-task/<ID>/complete endpoit is called multiple times at
the seme time, it lead to race condition in the
DatastoreAdapter.storeResult and DefaultDatastore.storeCompletedBuild
methods.

The first invocation of storeCompletedBuild could successfully store
the artifacts and build, which lead to conflicts in the simultaneous
calls to that method. These conflicts resulted in exception catched in
storeResult which was handled by marking the build as SYSTEM-ERROR and
trying to save the failed build again. This now overwrote* the first
correctly saved build with the failed build.

* We use Spring CrudRepository.save method which calls em.merge if the
  entity already is in DB.

### Checklist:

### DIFF
--- a/datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
+++ b/datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
@@ -141,6 +141,12 @@ public class DefaultDatastore implements Datastore {
             List<Artifact> dependencies) {
         BuildRecord buildRecord = buildRecordBuilder.build(true);
         logger.debug("Storing completed build {}.", buildRecord);
+        BuildRecord previouslySavedBuild = buildRecordRepository.queryById(buildRecord.getId());
+        if (previouslySavedBuild != null) {
+            throw new IllegalStateException(
+                    "When trying to save build " + buildRecord + " previously saved build with status "
+                            + previouslySavedBuild.getStatus() + " was found.");
+        }
         if (logger.isTraceEnabled()) {
             logger.trace("Build Log: {}.", buildRecord.getBuildLog());
         }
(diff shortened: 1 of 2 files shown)


datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
```diff
@@ -30,6 +30,7 @@
 import org.jboss.pnc.model.User;
 import org.jboss.pnc.spi.coordinator.BuildTask;
 import org.jboss.pnc.spi.datastore.Datastore;
+import org.jboss.pnc.spi.datastore.DatastoreException;
```
pkocandr, line 33: This added import doesn't seem to be used.

datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
```diff
@@ -141,6 +142,12 @@ public BuildRecord storeCompletedBuild(
             List<Artifact> dependencies) {
         BuildRecord buildRecord = buildRecordBuilder.build(true);
         logger.debug("Storing completed build {}.", buildRecord);
+        BuildRecord previouslySavedBuild = buildRecordRepository.queryById(buildRecord.getId());
+        if (previouslySavedBuild != null) {
+            throw new IllegalStateException(
+                    "When trying to save build " + buildRecord + " previously saved build " + previouslySavedBuild
```
pkocandr, line 148: Maybe I would consider adding the dto in the exception message only once if they are the same.

pkocandr (approved):
Mostly ok, just a few details.

datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
janinko: they may have different status

PR #3579, opened 2022-01-18
[NCLSUP-507] Send revision info to RHPAM when creating build config

thescouser89:
On creation of a build configuration, PNC:
1. creates a repository configuration (via RHPAM)
2. creates the build configuration with the repository configuration info

In step 1, the external scm repository is synced to the internal scm repository.
If a revision is specified, then we should only sync that revision. This
is already implemented in Repour, but the revision information is not
being sent to Repour.

To do so, we need to pass the information of the revision from PNC to
RHPAM. This commit is the implementation of this.

### Checklist:


thescouser89:
@thescouser89  Don't forget to add the changelog aftewards.

PR #3580, opened 2022-01-19
[NCL-6906] Add AlignmentPreference.

matejonnet:
### Checklist:

### DIFF
--- a/rest-api/pom.xml
+++ b/rest-api/pom.xml
@@ -40,6 +40,10 @@
         <!-- Project dependencies -->
         <dependency>
             <groupId>org.jboss.pnc</groupId>
+            <artifactId>pnc-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.pnc</groupId>
             <artifactId>dto</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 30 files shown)


janinko (changes requested):

rest-api/pom.xml
```diff
@@ -38,10 +38,19 @@
...
+            <artifactId>pnc-api</artifactId>
+        </dependency>
         <dependency>
             <groupId>org.jboss.pnc</groupId>
             <artifactId>dto</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.hibernate.validator</groupId>
+            <artifactId>hibernate-validator</artifactId>
+            <scope>provided</scope>
```
janinko, line 52: It's used in tests only? If so, please change it to test scope

janinko (changes requested):

rest-api/pom.xml
matejonnet: It's the same as we have in the DTO module.

janinko: but in DTO it¨s because of org.hibernate.validator.constraints.URL that is used in DeliverablesAnalysisRequest

matejonnet: ok, changing it to test.

janinko (approved):

PR #3584, opened 2022-02-07
[NCL-6695] Update Deliverable analyzer endpoins and add Operations

janinko:
### Checklist:


janinko:
Test will probably fail and I might add more.

model/src/main/resources/db-update/20011-from-2.1.1-to-2.2.0.sql
```diff
@@ -0,0 +1,52 @@
...
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+
+-- NCL-6695: Add Operation table
+BEGIN;
+
+    CREATE TABLE operation (
+        id bigint NOT NULL,
+        starttime timestamptz NOT NULL,
```
vibe13, line 24: @janinko this should be probably nullable, as we said the operation will be started after it is inserted

vibe13 (changes requested):
Looks good to me, wow lot of content! I believe we should change the NOT NULL constraint on the sql script, otherwise all looks ok from our discussions. Thanks!

facade/src/main/java/org/jboss/pnc/facade/impl/OperationsManagerImpl.java
```diff
@@ -0,0 +1,142 @@
...
+        if (milestone == null) {
+            throw new EmptyEntityException("Milestone with id " + milestoneId + " doesn't exist");
+        }
+
+        String operationId = Sequence.nextBase32Id();
+        MDCUtils.addProcessContext(operationId);
+        org.jboss.pnc.model.DeliverableAnalyzerOperation operation = org.jboss.pnc.model.DeliverableAnalyzerOperation.Builder
+                .newBuilder()
+                .progressStatus(ProgressStatus.NEW)
+                .startTime(Date.from(Instant.now()))
```
vibe13, line 99: should be submitTime now

vibe13 (changes requested):
One more thing to change I believe, then all is good!

matejonnet (approved):

vibe13 (approved):

PR #3589, opened 2022-03-07
[NCL-6957] - Fix requires rebuild checks

vibe13:


pkocandr (approved):
Looks good, thank you!

spi/src/main/java/org/jboss/pnc/spi/datastore/repositories/BuildRecordRepository.java
```diff
@@ -77,12 +79,32 @@ default BuildRecord getLatestSuccessfulBuildRecord(List<BuildRecord> buildRecord
                 .orElse(null);
     }
 
+    BuildRecord getAnyLatestSuccessfulBuildRecordWithBuildConfig(Integer configurationId, boolean temporaryBuild);
+
+    default BuildRecord getAnyLatestSuccessfulBuildRecord(List<BuildRecord> buildRecords, boolean temporaryBuild) {
+        return buildRecords.stream()
+                .filter(record -> record.getStatus() == BuildStatus.SUCCESS)
+                // First part includes temporary BRs and second part includes persistent BRs
+                .filter(record -> (temporaryBuild && record.isTemporaryBuild()) || !record.isTemporaryBuild())
```
janinko, line 88: note: maybe this whole method can be deleted, see other comments
```suggestion
                .filter(record -> temporaryBuild || !record.isTemporaryBuild())
```

model/src/main/resources/db-update/20030-from-2.2.0-to-2.3.0.sql
```diff
@@ -20,4 +20,7 @@
 BEGIN transaction;
     ALTER TABLE buildrecord ADD COLUMN alignmentpreference varchar(255);
     ALTER TABLE buildconfigsetrecord ADD COLUMN alignmentpreference varchar(255);
+
+    UPDATE buildrecord SET alignmentpreference = 'PREFER_TEMPORARY' WHERE temporarybuild IS TRUE AND alignmentpreference IS NULL;
+    UPDATE buildconfigsetrecord SET alignmentpreference = 'PREFER_TEMPORARY' WHERE temporarybuild IS TRUE AND alignmentpreference IS NULL;
```
janinko, line 25: BTW why do we have it as field in buildRecord and not build attribute?

facade/src/main/java/org/jboss/pnc/facade/impl/BrewPusherImpl.java
```diff
@@ -251,7 +251,7 @@ private BuildRecord getLatestSuccessfullyExecutedBuildRecord(Base32LongID buildR
             // if status is NO_REBUILD_REQUIRED, find the last BuildRecord with status SUCCESS for the same idRev.
             IdRev idRev = buildRecord.getBuildConfigurationAuditedIdRev();
             BuildRecord latestSuccessfulBuildRecord = buildRecordRepository
-                    .getLatestSuccessfulBuildRecord(idRev, buildRecord.isTemporaryBuild());
+                    .getAnyLatestSuccessfulBuildRecordWithRevision(idRev, buildRecord.isTemporaryBuild());
```
janinko, line 254: I think it's better to use ` buildRecord.getNoRebuildCause();`
```java
            BuildRecord noRebuildCause = buildRecord.getNoRebuildCause();
            if (noRebuildCause != null) {
                return noRebuildCause;
            } else {
                String message = "There is no SUCCESS build before NO_REBUILD_REQUIRED.";
                log.error(message);
                throw new InconsistentDataException(message);
            }
```

datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
```diff
@@ -382,25 +380,31 @@ public boolean requiresRebuild(
             BuildConfigurationAudited buildConfigurationAudited,
             boolean checkImplicitDependencies,
             boolean temporaryBuild,
+            AlignmentPreference alignmentPreference,
             Set<Integer> processedDependenciesCache,
             Consumer<BuildRecord> nonRebuildCauseSetter) {
+
         IdRev idRev = buildConfigurationAudited.getIdRev();
+        // Step 1 - Check the existence of a build with the same revision
```
janinko, line 388: :heart:

datastore/src/main/java/org/jboss/pnc/datastore/repositories/BuildRecordRepositoryImpl.java
```diff
@@ -133,6 +134,14 @@ public BuildRecord getLatestSuccessfulBuildRecord(Integer configurationId, boole
         return getLatestSuccessfulBuildRecord(buildRecords, temporaryBuild);
     }
 
+    @Override
+    public BuildRecord getAnyLatestSuccessfulBuildRecordWithBuildConfig(
+            Integer configurationId,
+            boolean temporaryBuild) {
+        List<BuildRecord> buildRecords = queryWithBuildConfigurationId(configurationId);
+        return getAnyLatestSuccessfulBuildRecord(buildRecords, temporaryBuild);
```
janinko, line 142: Maybe we can use simmilar method to `getAnyLatestSuccessfulBuildRecordWithRevision` but with `withBuildConfigurationId` instead of `withBuildConfigurationIdRev`. Then we don't need `getAnyLatestSuccessfulBuildRecord` and we are getting less data from DB.

I guess before the logic was too complex for easy query like this.
```suggestion
        PageInfo pageInfo = new DefaultPageInfo(0, 1);
        SortInfo sortInfo = new DefaultSortInfo(SortInfo.SortingDirection.DESC, BuildRecord_.submitTime.getName());

        List<BuildRecord> buildRecords = queryWithPredicates(
                pageInfo,
                sortInfo,
                withBuildConfigurationId(configurationId),
                withSuccess(),
                includeTemporary(temporaryBuild));

        if (buildRecords.size() == 0) {
            return null;
        } else {
            return buildRecords.get(0);
        }
```

pnc-mock/src/main/java/org/jboss/pnc/mock/repository/BuildRecordRepositoryMock.java
```diff
@@ -88,6 +97,18 @@ public static BuildRecord getLatestSuccessfulBuildRecord(
...
     }
 
+    public static BuildRecord getLatestSuccessfulBuildRecordFromList(
+            Integer configurationId,
+            List<BuildRecord> buildRecords,
+            boolean temporaryBuild) {
+        return buildRecords.stream()
+                .filter(br -> br.getBuildConfigurationId().equals(configurationId))
+                .filter(br -> br.getStatus().equals(BuildStatus.SUCCESS))
+                .filter(br -> (temporaryBuild && br.isTemporaryBuild()) || !br.isTemporaryBuild())
```
janinko, line 107:
```suggestion
                .filter(br -> temporaryBuild && || !br.isTemporaryBuild())
```

pnc-mock/src/main/java/org/jboss/pnc/mock/repository/BuildRecordRepositoryMock.java
```diff
@@ -117,6 +157,56 @@ public static BuildRecord getLatestSuccessfulBuildRecord(
...
+            List<BuildRecord> buildRecords,
+            boolean temporaryBuild) {
+        Optional<BuildRecord> first = buildRecords.stream()
+                .filter(buildRecord -> buildRecord.getStatus().equals(BuildStatus.SUCCESS))
+                .filter(
+                        buildRecord -> buildRecord.getBuildConfigurationAuditedIdRev()
+                                .equals(buildConfigurationAuditedIdRev))
+                .filter(
+                        buildRecord -> (temporaryBuild && buildRecord.isTemporaryBuild())
+                                || !buildRecord.isTemporaryBuild())
```
janinko, line 171:
```suggestion
                        buildRecord -> temporaryBuild || !buildRecord.isTemporaryBuild())
```

janinko (changes requested):

model/src/main/resources/db-update/20030-from-2.2.0-to-2.3.0.sql
vibe13: We stored it similarly to the temporary build option, which is also easier to consume.

spi/src/main/java/org/jboss/pnc/spi/datastore/repositories/BuildRecordRepository.java
vibe13: ack ... maybe less readable at first glance ? but...who will ever forget... and putting a comment will do :-)

facade/src/main/java/org/jboss/pnc/facade/impl/BrewPusherImpl.java
vibe13: ah nice one

datastore/src/main/java/org/jboss/pnc/datastore/repositories/BuildRecordRepositoryImpl.java
vibe13: I like it much better yes, thanks for that, I overlooked the possible optimization although I never liked the way it retrieved all the build records before. Thanks!

pnc-mock/src/main/java/org/jboss/pnc/mock/repository/BuildRecordRepositoryMock.java
vibe13: ack, thanks!

pnc-mock/src/main/java/org/jboss/pnc/mock/repository/BuildRecordRepositoryMock.java
vibe13: ack thanks!

vibe13:
@janinko @michalovjan All the steps should be changed now!

facade/src/main/java/org/jboss/pnc/facade/impl/BrewPusherImpl.java
janinko: Please remove the comment `                // In case of temporary builds it can happen. The SUCCESS build might be already garbage-collected.`, it's not true anymore

pnc-mock/src/main/java/org/jboss/pnc/mock/repository/BuildRecordRepositoryMock.java
```diff
@@ -166,9 +154,7 @@ public static BuildRecord getLatestSuccessfulBuildRecord(
                 .filter(
                         buildRecord -> buildRecord.getBuildConfigurationAuditedIdRev()
                                 .equals(buildConfigurationAuditedIdRev))
-                .filter(
-                        buildRecord -> (temporaryBuild && buildRecord.isTemporaryBuild())
-                                || !buildRecord.isTemporaryBuild())
+                .filter(buildRecord -> temporaryBuild || !buildRecord.isTemporaryBuild())
```
janinko, line 171: I just noticed the condition at line 95 is the same, just with applied De Morgan's rule. It would be nice to have the same one everywhere.

janinko (changes requested):

janinko (approved):

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/BuildRecordPredicates.java
```diff
@@ -124,6 +125,139 @@
...
+                if (AlignmentPreference.PREFER_TEMPORARY.equals(alignmentPreference)) {
+                    Subquery<Long> temporaryCount = query.subquery(Long.class);
+                    Root<BuildRecord> subRoot = temporaryCount.from(BuildRecord.class);
+                    temporaryCount.select(cb.count(subRoot.get(BuildRecord_.id)));
+                    temporaryCount.where(
+                            cb.and(
+                                    cb.isTrue(subRoot.get(BuildRecord_.temporaryBuild)),
+                                    cb.equal(subRoot.get(BuildRecord_.status), BuildStatus.SUCCESS),
+                                    cb.equal(subRoot.get(BuildRecord_.buildConfigurationId), idRev.getId()),
+                                    cb.equal(subRoot.get(BuildRecord_.buildConfigurationRev), idRev.getRev())));
```
michalovjan, line 175:
```suggestion                    
```
This subquery should actually look at all builds in the history of the BC and get a count of temporary builds. Looking at just the ID REV is not what DA does as DA doesn't know about IDREVs and just looks at what versions have been built. So maybe just the method `includeTemporary( Integer configurationId, boolean temporary, AlignmentPreference alignmentPreference)` is required and this one can actually be deleted. 

The logic for figuring out whether a dependency 'has newer version' should be the same for all dependencies, doesn't matter if they are explicit or implicit.

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/BuildRecordPredicates.java
```diff
@@ -124,6 +125,139 @@
...
+                } else {
+                    Subquery<Long> persistentCount = query.subquery(Long.class);
+                    Root<BuildRecord> subRoot = persistentCount.from(BuildRecord.class);
+                    persistentCount.select(cb.count(subRoot.get(BuildRecord_.id)));
+                    persistentCount.where(
+                            cb.and(
+                                    cb.isFalse(subRoot.get(BuildRecord_.temporaryBuild)),
+                                    cb.equal(subRoot.get(BuildRecord_.status), BuildStatus.SUCCESS),
+                                    cb.equal(subRoot.get(BuildRecord_.buildConfigurationId), idRev.getId()),
+                                    cb.equal(subRoot.get(BuildRecord_.buildConfigurationRev), idRev.getRev())));
```
michalovjan, line 190:
```suggestion
```
already explained in other comment

michalovjan (changes requested):

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/BuildRecordPredicates.java
vibe13: I have made the change to use the same method (checking only for the same build config) .... and I left all the previous methods on purpose in case we need to switch back :-)

spi/src/main/java/org/jboss/pnc/spi/datastore/repositories/BuildRecordRepository.java
```diff
@@ -83,6 +85,26 @@ default BuildRecord getLatestSuccessfulBuildRecord(List<BuildRecord> buildRecord
...
 
+    // Used in Step 1 in DefaultDataStore -> requiresRebuild
+    BuildRecord getAnyLatestSuccessfulBuildRecordWithRevision(
+            IdRev buildConfigurationAuditedIdRev,
+            boolean temporaryBuild);
+
+    // Used in Step 2 in DefaultDataStore -> requiresRebuild
+    BuildRecord getAnyLatestSuccessfulBuildRecordWithBuildConfig(Integer configurationId, boolean temporaryBuild);
+
+    // Used in Step 3 in DefaultDataStore -> requiresRebuild
```
michalovjan, line 96: Not used in step 3 anymore

spi/src/main/java/org/jboss/pnc/spi/datastore/repositories/BuildRecordRepository.java
```diff
@@ -83,6 +85,26 @@ default BuildRecord getLatestSuccessfulBuildRecord(List<BuildRecord> buildRecord
...
+    BuildRecord getPreferredLatestSuccessfulBuildRecordWithRevision(
+            IdRev buildConfigurationAuditedIdRev,
+            boolean temporaryBuild,
+            AlignmentPreference alignmentPreference);
+
+    // Used in Step 4 in DefaultDataStore -> requiresRebuild
```
michalovjan, line 102: and step 3

michalovjan (changes requested):

vibe13:
retest this please

PR #3596, opened 2022-03-25
[NCL-6965] - Fix running builds search by BuildConfigName

vibe13:
Handle the % wildcard as received from the UI


janinko (approved):

janinko:
I'm thinking if we could mabye use `Pattern.quote()` to quote everything and then replace `*` and `%` with `\E.*\Q`?

vibe13:
@janinko Yep that can be a good idea, I will consider it for the following work needed for the "_" handling. Merging this one now.

PR #3637, opened 2022-10-17
[NCLSUP-758] don't try to delete artifacts from indy on failed builds

michalovjan:
### Checklist:

### DIFF
--- a/build-coordinator/src/main/java/org/jboss/pnc/coordinator/maintenance/DefaultRemoteBuildsCleaner.java
+++ b/build-coordinator/src/main/java/org/jboss/pnc/coordinator/maintenance/DefaultRemoteBuildsCleaner.java
@@ -145,12 +145,16 @@ public class DefaultRemoteBuildsCleaner implements RemoteBuildsCleaner {
         try {
             IndyStoresClientModule indyStores = indy.stores();
             if (pkgKey != null) {
-                StoreKey tempHostedKey = new StoreKey(pkgKey, StoreType.hosted, tempBuildPromotionGroup);
                 // delete artifacts from consolidated repository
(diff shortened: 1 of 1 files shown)


build-coordinator/src/main/java/org/jboss/pnc/coordinator/maintenance/DefaultRemoteBuildsCleaner.java
```diff
@@ -144,7 +144,7 @@ private Result deleteBuildsFromIndy(BuildRecord buildRecord, String authToken) {
         Indy indy = indyFactory.get(authToken);
         try {
             IndyStoresClientModule indyStores = indy.stores();
-            if (pkgKey != null) {
+            if (buildRecord.getStatus().completedSuccessfully() && pkgKey != null) {
```
pkocandr, line 147: This would cause that the uploaded content and the hosted repo cotaining it would not be deleted for failed builds. Please wrap only the deletion from temporary builds (following 6 lines) into the condition. The lines below `delete the content` still have to be executed.

pkocandr (changes requested):
Not happy.

build-coordinator/src/main/java/org/jboss/pnc/coordinator/maintenance/DefaultRemoteBuildsCleaner.java
michalovjan: Thank you very much, you are absolutely correct. I've not noticed that there are 2 deletions happening in the `if` block.

pkocandr (approved):
Now I'm happy :-)

PR #3650, opened 2022-11-03
Ncl 7311 log events of interest

matejonnet:
### Checklist:

### DIFF
--- a/rest/src/main/java/org/jboss/pnc/rest/SecurityConstraintFilter.java
+++ b/rest/src/main/java/org/jboss/pnc/rest/SecurityConstraintFilter.java
@@ -1,0 +1,80 @@
+/**
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014-2022 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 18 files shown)


rest/src/main/java/org/jboss/pnc/rest/SecurityConstraintFilter.java
```diff
@@ -0,0 +1,80 @@
...
+public class SecurityConstraintFilter implements ContainerRequestFilter {
+
+    private Logger logger = LoggerFactory.getLogger(SecurityConstraintFilter.class);
+    private static final String REQUEST_EXECUTION_START = "request-execution-start";
+
+    @Inject
+    UserService userService;
+
+    @Override
+    public void filter(ContainerRequestContext requestContext) throws IOException {
```
janinko, line 50: Why do you create custom filter instead of the web.xml config?

janinko (changes requested):
Didn't check if it matches the JIRA requirement, but code looks good, just few small things

rest/src/main/java/org/jboss/pnc/rest/SecurityConstraintFilter.java
matejonnet: To be able to catch and log NotAuthorizedException, with web.xml (servlet constraint) the request is rejected before reaching reast-easy.

janinko (approved):

PR #3670, opened 2023-02-06
[NCL-7265] Migrate out of spring-data

janinko:
### Checklist:


michalovjan (changes requested):
Please push important methods from AbstractRepository to Repository and evaluate again which methods should be public and private again.

janinko:
@michalovjan done, publicized just `delete(EntityType entity)`

janinko:
And I've added commit with SortInfo changes

vibe13 (approved):

datastore/src/main/java/org/jboss/pnc/datastore/repositories/internal/AbstractRepository.java
```diff
@@ -18,116 +18,168 @@
...
-        return springRepository.findById(id).orElse(null);
+        Objects.requireNonNull(id, "The given id must not be null!");
+        return this.entityManager.find(entityClass, id);
     }
 
     @Override
     public T queryByPredicates(Predicate<T>... predicates) {
-        return springSpecificationsExecutor.findOne(SpecificationsMapper.map(predicates)).orElse(null);
+        try {
+            return getQuery(null, predicates).setMaxResults(2).getSingleResult();
```
michalovjan, line 137: ?? why is this setMaxResults 2 if you want a single result?? I guess is copy-paste but wtf?

janinko: I'ts copy pasta. How I understand is that having no maxResult, it could query without "LIMIT" in the SQL, potentionaly returning a lot of data from DB, before failing in java with NonUniqueResultException. Having maxResult 1 would always return at most 1 entity and never failing with NonUniqueResultException.

janinko: https://github.com/spring-projects/spring-data-jpa/issues/2594
https://github.com/spring-projects/spring-data-jpa/commit/fa41595818eeb0d1da7f271a7f0554eb544fc6e3

michalovjan: I see thx

michalovjan (approved):

PR #3679, opened 2023-03-06
[NCL-7427] notification endpoint + set record status job

michalovjan:
### Checklist:


rest/src/main/java/org/jboss/pnc/rest/endpoints/internal/BuildTaskEndpointImpl.java
```diff
@@ -113,8 +167,194 @@ public Response buildTaskCompleted(String buildId, BuildResultRest buildResult)
...
+        if (!buildId.equals(notification.getTask().getName())) {
+            return Response.status(Response.Status.BAD_REQUEST)
+                    .entity("Build ID does not correspond to associated Rex Task")
+                    .build();
+        }
+
+        BuildMeta buildMeta = parseBuildMeta(buildId, notification);
+
+        // Task has to be present in Rex, otherwise it is unknown
+        validateBuildIsActive(buildId, notification);
```
matejonnet, line 191: Isn't Rex who is invoking this endpoint? Why asking Rex again ?

matejonnet (changes requested):

michalovjan:
retest this please

rest/src/main/java/org/jboss/pnc/rest/endpoints/internal/BuildTaskEndpointImpl.java
michalovjan: Because, without this check, you would be able to use this endpoint to create or send notification about a build record that may have never been scheduled by Orch. This could be potentially very harmful.

michalovjan:
retest this please

rest/src/main/java/org/jboss/pnc/rest/endpoints/internal/BuildTaskEndpointImpl.java
```diff
@@ -113,8 +167,194 @@ public Response buildTaskCompleted(String buildId, BuildResultRest buildResult)
...
         return buildTaskCompleted(buildId, buildResult);
     }
+
+    @Override
+    @Transactional
+    public Response buildTaskNotification(@NotBlank String buildId, @NotNull NotificationRequest notification)
+            throws InvalidEntityException {
+        logger.debug("Received transition notification for Build '{}'", buildId);
+
+        if (!buildId.equals(notification.getTask().getName())) {
```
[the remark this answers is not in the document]
michalovjan: Rex Task should match buildID. Somewhere along the code I map RexTask to BuildTask stuff and it would be good to have the endpoint buildID match the record that gets saved. I'll comment it up.

rest/src/main/java/org/jboss/pnc/rest/endpoints/internal/BuildTaskEndpointImpl.java
matejonnet: This should be prevented by the authorization. Currently we are using user's token though, as we trust the users that they are not trying to break their own jobs. If required we might consider using Rex service account.

matejonnet (approved):

michalovjan:
tests failures are unrelated

PR #3680, opened 2023-03-16
[NCL-7640] Add support for upgrade paths for deprecated environments

janinko:
Add new endpoint to deprecate environments
Set DEPRECATION_UPGRADE attribute when an environment is being deprecated.

### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/providers/EnvironmentProviderImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/providers/EnvironmentProviderImpl.java
@@ -18,6 +18,7 @@
 package org.jboss.pnc.facade.providers;
 
 import static org.jboss.pnc.facade.providers.api.UserRoles.SYSTEM_USER;
+import static org.jboss.pnc.spi.datastore.predicates.EnvironmentPredicates.replacedBy;
 import static org.jboss.pnc.spi.datastore.predicates.EnvironmentPredicates.withEnvironmentNameAndActive;
 
 import java.util.List;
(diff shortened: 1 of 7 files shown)


janinko:
Requires https://github.com/project-ncl/pnc-api/pull/138 to be merged

facade/src/main/java/org/jboss/pnc/facade/providers/EnvironmentProviderImpl.java
```diff
@@ -48,7 +51,9 @@ public EnvironmentProviderImpl(BuildEnvironmentRepository repository, Environmen
     @Override
     @RolesAllowed(SYSTEM_USER)
     public Environment store(Environment restEntity) throws DTOValidationException {
-        return super.store(restEntity);
+        Environment storedEntity = super.store(restEntity);
+        deprecateActiveEnvironmentsWithSameName(storedEntity);
+        return storedEntity;
```
pkocandr, line 56: I'm missing one thing here. When we deprecate an older version of an env which was an upgrade target for another deprecated env, should we update also the DEPRECATION_UPGRADE value which pointed to the currently deprecated environment to point at the new one? That way we would return the upgrade target right away and save the user from having to walk through the deprecated path iteratively.

pkocandr: Or maybe not here, but in the deprecateActiveEnvironmentsWithSameName().

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/EnvironmentPredicates.java
```diff
@@ -17,10 +17,15 @@
...
 package org.jboss.pnc.spi.datastore.predicates;
 
+import org.jboss.pnc.api.constants.Attributes;
 import org.jboss.pnc.model.BuildEnvironment;
 import org.jboss.pnc.model.BuildEnvironment_;
 import org.jboss.pnc.spi.datastore.repositories.api.Predicate;
 
+import javax.persistence.criteria.Expression;
+import javax.persistence.criteria.MapJoin;
+import java.util.Map;
```
pkocandr, line 27: Another unused import.

pkocandr (changes requested):
I believe the updated envs need to be saved explicitely. The unused imports are just a cherry on top :-)

facade/src/main/java/org/jboss/pnc/facade/providers/EnvironmentProviderImpl.java
```diff
@@ -58,21 +64,54 @@ public void delete(String id) {
...
+        }
+        deprecatedEnvironment.setDeprecated(true);
+        deprecatedEnvironment.putAttribute(Attributes.DEPRECATION_REPLACEMENT, replacementId);
+        shortenTransitivePath(id, replacementId);
+        return mapper.toDTO(deprecatedEnvironment);
+    }
+
+    private void shortenTransitivePath(String deprecatedId, String replacementId) {
+        for (BuildEnvironment environment : repository.queryWithPredicates(replacedBy(deprecatedId))) {
+            environment.putAttribute(Attributes.DEPRECATION_REPLACEMENT, replacementId);
```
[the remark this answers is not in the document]
pkocandr: OK.

pkocandr (approved):

PR #3694, opened 2023-04-19
[NCLSUP-864] Optimize withIdentifierAndSha256 artifact search

janinko:
Current implementation in one case   took over 30 s for 387 checksums and returned 217548 artifacts. This implementation returns only 387 artifacts and seems to be faster (can't tell exaclty how, because the DB was warmed up, butt it's still at least 1 second).

### DIFF
--- a/spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
+++ b/spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
@@ -31,6 +31,8 @@ import org.jboss.pnc.model.ProductMilestone_;
 import org.jboss.pnc.spi.datastore.repositories.api.Predicate;
 
 import javax.persistence.criteria.Join;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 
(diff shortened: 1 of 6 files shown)


thescouser89:
I'll defer the review

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
```diff
@@ -141,6 +143,19 @@ public static Predicate<Artifact> withSha256(Optional<String> sha256) {
...
+    public static Predicate<Artifact> withIdentifierAndSha256(Set<Artifact.IdentifierSha256> identifierSha256Set) {
+        return ((root, query, cb) -> {
+            List<javax.persistence.criteria.Predicate> predicates = new ArrayList<>();
+            for (Artifact.IdentifierSha256 identifierSha256 : identifierSha256Set) {
+                predicates.add(
+                        cb.and(
+                                cb.equal(root.get(Artifact_.identifier), identifierSha256.getIdentifier()),
+                                cb.equal(root.get(Artifact_.sha256), identifierSha256.getSha256())));
+            }
+            return cb.or(predicates.toArray(new javax.persistence.criteria.Predicate[0]));
```
pkocandr, line 155: Why do you use 0 here when the required size of the array is known (`predicates.size()`). This way you allocate an array that's thrown away immediately without usage and a new one is allocated.

pkocandr (approved):
Just a minor optimization could be done. Other than that it seems ok and probably can be merged even without it.

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
janinko: because AI told me to :D
I actually like it, because it's shorter code and the overhead is minimal

janinko: It's even mentioned as an example in javadoc: https://docs.oracle.com/javase/7/docs/api/java/util/Set.html#toArray%28T[]%29 and drecomended when using newer java, acording to https://stackoverflow.com/a/65902425

pkocandr: Yeah, I saw it in the javadoc too, but the implementation below shows that in case the array is shorter than the list it creates a new array right away and the one passed in is thrown away. Also the concern about collection being shrunk during the toArray execution is not valid in this case because the list is not used in multiple threads. But anyway as it claims the performance is the same in newer JDK I'm totally fine with leaving it as it is. Thanks for the reaction :-)

PR #3697, opened 2023-04-25
[Alignment 2.0] implement storing, retrieval and propagation

michalovjan:
### Checklist:

Dependencies:
requires https://github.com/project-ncl/pnc-common/pull/46
requires https://github.com/project-ncl/pnc-api/pull/141

#### New field AlignmentConfig in BuildConfiguration/Audited:
With Alignment 2.0, user will be able to influence alignment process in according to their use case.
AlignmentConfig has 4 fields:
1. _artifactScope_:
   - If it's null, constraints are applied on all GAV in LookupRequests. Otherwise, it can specify G, GA or GAV as scope to which GAVs apply the constraints. More specific scopes override less specific ones. (won't have an effect on NPM configs as project-manipulator does not do dependency manipulation with DA)
2. _ranks_:
   - Defines preferences of which qualities of Artifacts should be prioritised. Consists of ranks of Qualifiers which are PNC-specific data. F.e. Qualifier PRODUCT with value EAP as a first rank will tell later DA in the alignment process to prefer artifacts which are associated with PRODUCT:EAP.  
3. _denyList_
   - Comma-seperated list of Qualifiers which signify Artifacts that should be excluded from analyses. 
4. _allowList_
   - Comma-seperated list of Qualifiers which signify Artifacts that should be exclusively included in analysis. 

Each individual Qualifier is **verified to exist** so user cannot reference a missing entity.

##### From an entity-model standpoint:
- The entity class AlignConfig is an '**embedabble**' class which means that it is **directly bound** to BuildConfiguration similarly as build configuration parameters. This means that updates, deletions, inserts are propagated. 
- The AlignConfig class has a custom Hibernate type on Lists. The type persists the List of Strings as a single String where parts are delimited by ';'.
  - it was almost necessary as persisting Lists is very hard to design in databases, because in general databases don't have a type for arrays (PostgreSQL does but H2) and other solutions would result in a extra join table for every List
  - the drawback is that the Lists have to be loaded/saved/updated from DB in their full whereas with extra tables you could access individual indexes. Fortunately, there is no use case for that, so we save a lot of extra tables.
- For each _field_ in AlignmentConfig there is an _idField_ variant. These are generated in AlignConfigMapper and used both for **optimization** (with ID, you don't have to query for abbreviations and product versions) and **auditability**.  
  - **Auditability** is important as config names, abbreviations, version strings can change even though this does not happen often. We will still see which IDs the alignment was running with and we will have them stored (also in BCA).
- _artifactScope_ from AConfig DTO is mapped to _key_ in alignConfigs' Map
  - in case it is null in DTO, it is mapped to default value which is GLOBAL_SCOPE

##### High-level
This is an optimized endpoint used by DA during alignment to get just enough data about Artifacts without extra baggage like Build data, which you get in ordinary /artifact/getAll endpoint.

I've added an _QueryParam_ so that DA can request **which Qualifiers to attach to ArtifactInfos**. DA specifies not just Qualifiers, but also their Values (QValue.class). The endpoint returns **just enough** data. If DA requested DEPENDENCY:AAAA but the Artifact is also dependency of BBBB. The endpoint will not return BBBB, just AAAA even though the information is available.

The reasoning behind this behaviour is because for multi-value qualifiers like DEPENDENCY, or in the future QUALITY, Orch could possibly return very high amount of unnecessary data. A single Artifact could've been used in hundreds of builds and DA queries for thousands of artifacts. Additionally, DA doesn't require to know that Artifact A is from Quarkus if it will only rank/filter the ones from EAP. 

##### DB-Query
- had to move out of Criteria/JPQL to native SQL
- Native SQL Queries allow joins of arbitrary tables which are not shown or are generated by non-jpa libraries like Hibernate envers
- Additionally, due to BuildRecord BuildConfigurationID field being an actual Integer ID, and not reference to BuildConfiguration table, you are unable to construct TypedQueries because JPA/Hibernate is super-strict on types
- Native query strictly adheres to generic SQL for portability (h2, postgre)
- I've tried to relieve most of the annoyance of making the Query by using a sort of builder (the code is in https://github.com/project-ncl/pnc-api/pull/141)
- the Query joins just enough to get data and for multi-value Qualifiers there are additional filters so that we save join time

#### Passing the AConfig stuff to BPM->Repour->PME/GME->DA
- Currently Orch adds the stuff into default generic parameters so that Repour passes it without much hassle to PME/GME
- In the future we would definitely want a seperate field for this. 

AlignConfigMapper which maps AlignmentConfiguration DTO into entity-model counterpart is currently written by hand mostly without MapStruct's help, as that was faster due to MapStruct's shenanigans. I believe it could be ported to MapStruct definition if there would be a strong desire to do that.


michalovjan:
retest this please

PR #3703, opened 2023-05-11
Nclsup 834

janinko:
### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
@@ -17,20 +17,25 @@
  */
 package org.jboss.pnc.facade.deliverables;
 
+import com.github.packageurl.MalformedPackageURLException;
+import com.github.packageurl.PackageURL;
+import com.github.packageurl.PackageURLBuilder;
 import lombok.extern.slf4j.Slf4j;
 import org.jboss.pnc.api.deliverablesanalyzer.dto.Artifact;
 import org.jboss.pnc.api.deliverablesanalyzer.dto.ArtifactType;
 import org.jboss.pnc.api.deliverablesanalyzer.dto.Build;
 import org.jboss.pnc.api.deliverablesanalyzer.dto.FinderResult;
 import org.jboss.pnc.api.deliverablesanalyzer.dto.MavenArtifact;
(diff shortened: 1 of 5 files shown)


janinko:
retest this please

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
```diff
@@ -241,39 +269,103 @@ private org.jboss.pnc.model.Artifact getPncArtifact(Artifact art) {
...
         }
 
-        return builder.build();
+        return builder;
     }
 
-    private String fill(MavenArtifact mavenArtifact) {
+    private static String createDeployPath(MavenArtifact mavenArt) {
+        String filename = createFileName(mavenArt);
+        String deployPath = mavenArt.getGroupId().replace('.', '/') + "/" + mavenArt.getArtifactId() + "/"
```
pkocandr, line 320: deployPath should start with a slash

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
```diff
@@ -241,39 +269,103 @@ private org.jboss.pnc.model.Artifact getPncArtifact(Artifact art) {
...
+        filename += "." + mavenArt.getType();
+        return filename;
+    }
+
+    private String createBrewOriginURL(MavenArtifact mavenArt, String nvr) {
+        String brewContentUrl = globalConfig.getBrewContentUrl();
+
+        Matcher matcher = NVR_PATTERN.matcher(nvr);
+        if (!matcher.matches()) {
+            throw new IllegalArgumentException("NVR " + nvr + " does not match expected format.");
```
pkocandr, line 339: Wouldn't it be better to log an error and return null instead? Missing the originUrl is not a big deal - we don't really use it - and can be corrected later.

pkocandr (changes requested):
The missing slash at the beginning of the deployPath for Maven artifacts should be corrected. The rest are just details and don't really make much difference IMO, but would be nice to get them too.

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
janinko: this shuldn't really happen at all, unless there is a big bug somewhere

PR #3725, opened 2023-05-31
[NCLSUP-834] Fix missing "/maven" in origin URL

janinko:
### Checklist:


pkocandr (approved):
I was a bit worried if createDeployPath returns string with a leading slash, but it does, so should be ok.

PR #3734, opened 2023-06-08
NCL-7911 - check for null.

matejonnet:
### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationProviderImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationProviderImpl.java
@@ -548,15 +548,15 @@ public class BuildConfigurationProviderImpl extends
                 request.getPreBuildSyncEnabled(),
                 JobNotificationType.BUILD_CONFIG_CREATION,
                 Optional.of(newBuildConfigurationWithId));
-        createBuildConfigurationWithRepository(
-                rcResponse.getTaskId() == null ? null : rcResponse.getTaskId().toString(),
-                Integer.parseInt(rcResponse.getRepository().getId()),
-                newBuildConfigurationWithId);
 
         BuildConfigCreationResponse response;
         if (rcResponse.getTaskId() == null) {
             // scm is internal, not running a RepositoryCreationTask.
-            // onRCCreationSuccess already called with id = rcResponse.getRepository().getId()
+            createBuildConfigurationWithRepository(
+                    null,
+                    Integer.parseInt(rcResponse.getRepository().getId()),
+                    newBuildConfigurationWithId);
+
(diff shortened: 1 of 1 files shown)


facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationProviderImpl.java
```diff
@@ -550,7 +550,7 @@ public BuildConfigCreationResponse createWithScm(BuildConfigWithSCMRequest reque
                 Optional.of(newBuildConfigurationWithId));
         createBuildConfigurationWithRepository(
                 rcResponse.getTaskId() == null ? null : rcResponse.getTaskId().toString(),
-                Integer.parseInt(rcResponse.getRepository().getId()),
+                rcResponse.getRepository() == null ? null : Integer.parseInt(rcResponse.getRepository().getId()),
```
thescouser89, line 553: I'm confused why this code even compiles haa! I had to try it locally and it does indeed.

The second field is an `int`, so returning `null` to it and the compiler allowing it is strange. However at runtime we'll get a `NullPointerException`.

It also looks like `createBuildConfigurationWIthRepository` doesn't wait for the repository to be created by RHPAM before creating the build configuration.

matejonnet: Good catch, that should not compile at all.
Updated, not ideal as it's save to DB and then reading (although it should be cached) but that was the same before.

thescouser89: With this commit, for a build configuration that needs a repository configuration that doesn't exist, what is the workflow?

In the v2.4 branch, we wait for the event that the repository configuration is created by RHPAM before we create the build configuration (via a lambda).

In the v2.5 branch, I don't see any  wait or callback (via a lambda) for  the repository creation workflow to be finished before we create the build configuration. Has there been any other change that have been done to explain the new behaviour?

matejonnet: 2.4 already used the stateless approach [1], but the code was still there to allow switching back via config.
With 2.5 I removed the old methods.

1: https://github.com/project-ncl/pnc/blob/e3b6c87d0c0b003aad64b90070f68a9c9d439002/rest/src/main/java/org/jboss/pnc/rest/endpoints/internal/api/BpmEndpoint.java#L50

thescouser89: I don't understand the workflow to be honest but I'll merge it and run it to see how this works. I'll report in the JIRA if it works.

PR #3739, opened 2023-06-14
[NCL-7933] Use FQN for swagger class names

janinko:
Swagger can't deal with duplicate class names (like Build in 3 different classes), so switiching to FQN should fix the openapi

### Checklist:


janinko:
retest this please

PR #3797, opened 2023-08-23
[NCL-8047] Stop using user token for other requests

thescouser89:
We need to stop using the user token to send other requests to other services.

Instead we'll use our service account.

This is done because we do not know when the user token will expire and should therefore use a service account token that we can control and renew.

Moreover, the user token may not have the appropriate roles to submit requests to other applications.

### Checklist:


thescouser89:
retest this please

PR #3802, opened 2023-08-30
[NCL-8067]: Add scratch flag to DeliverableAnalysisRequest

akridI:
### Checklist:

### DIFF
--- a/integration-test/src/test/java/org/jboss/pnc/integration/endpoints/ProductMilestoneEndpointTest.java
+++ b/integration-test/src/test/java/org/jboss/pnc/integration/endpoints/ProductMilestoneEndpointTest.java
@@ -36,6 +36,7 @@ import org.jboss.pnc.dto.ProductMilestone;
 import org.jboss.pnc.dto.ProductMilestoneCloseResult;
 import org.jboss.pnc.dto.ProductVersion;
 import org.jboss.pnc.dto.ProductVersionRef;
+import org.jboss.pnc.dto.requests.DeliverablesAnalysisRequest;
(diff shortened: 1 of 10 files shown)


akridI:
TODO: Discussion with @janinko about using BPMWireMock together with @Before & @After, rather than @BeforeClass & @AfterClass

integration-test/src/test/java/org/jboss/pnc/integration/endpoints/ProductMilestoneEndpointTest.java
```diff
@@ -87,6 +103,7 @@ public class ProductMilestoneEndpointTest {
     private static ProductMilestone milestone;
     private static String milestoneId;
     private static ProductMilestone milestone2;
+    private static BPMWireMock bpm;
```
akridI, line 106: TODO: Discussion with @janinko about using BPMWireMock together with https://github.com/before & @after, rather than @BeforeClass & @afterclass

janinko: I think we've already talked about this?

janinko (approved):

integration-test/src/test/java/org/jboss/pnc/integration/endpoints/ProductMilestoneEndpointTest.java
akridI: yessir, you're right

akridI:
retest this please

PR #3817, opened 2023-09-12
No authorization for rex requestStart headers

thescouser89:
And same for requestCancel headers. Rex should use its own authorization token from now on.

### Checklist:


thescouser89:
retest this please

thescouser89:
@michalovjan ah I'll look into that, but I just removed the auth token in the headers. I'm scared :fearful:

thescouser89:
So that was me being stupid and removing the auth token from Orch -> Rex also (I was a bit too aggressive with the removals!)

Now I *don't* want to use the user token for Orch -> Rex, and will have to use the keycloak service account for pnc-orchestrator. In the test, it is mocked with a random string.

I'll try to just sneakily inject the user token into the mocked kecyloak service account to workaround it. I couldn't figure out how to get a service account client exported into Keycloak **with the secret set**.

thescouser89:
@michalovjan I think even though I try to inject the user token inside the mocked KeycloakServiceClient, the test still fails because that inject happens inside `BuildTest` which is running as a client. I don't really know how that works :/

Would you happen to know how I could configure the mocked KeycloakServiceClient running inside a container to get either:
- keycloak server information
- token
inside it from the "client" side of tests.

michalovjan:
Unfortunately with the roles, the mocked Wiremock BPM needs to have its own auth for callback to Rex... We can hack it I guess so that it uses the token Rex passes but we need to add the rex-admin role to it in that case. I have a solution for fixed secrets.

thescouser89:
retest this please

PR #3824, opened 2023-09-25
[NCL-8156] achieve determinism for build endpoints

michalovjan:
### Checklist:

@janinko As more than one endpoint uses `getBuildList(...)` method, the default sort change from `ID asc` to `submitTime desc, ID desc` will also affect default sorting on these endpoints:

- **GET /product-milestones/{{id}}/builds**
- **GET /projects/{{id}}/builds**
- **GET /build-configs/{{id}}/builds**
- **GET /users/{{id}}/builds**
- **GET /group-builds/{{id}}/builds**
- **GET /group-configs/{{id}}/builds**

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
@@ -112,6 +112,7 @@ import java.util.Spliterators;
 import java.util.TreeSet;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.ToLongFunction;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
@@ -134,6 +135,7 @@ import static org.jboss.pnc.spi.datastore.predicates.BuildRecordPredicates.withU
 import static org.jboss.pnc.spi.datastore.predicates.BuildRecordPredicates.withoutAttribute;
 import static org.jboss.pnc.spi.datastore.predicates.BuildRecordPredicates.withoutImplicitDependants;
 import static org.jboss.pnc.spi.datastore.predicates.BuildRecordPredicates.withoutLinkedNRRRecordOlderThanTimestamp;
+import static org.jboss.pnc.spi.datastore.repositories.api.OrderInfo.SortingDirection.DESC;
 
 @PermitAll
 @Stateless
@@ -895,7 +897,7 @@ public class BuildProviderImpl extends AbstractUpdatableProvider<Base32LongID, B
         List<Build> runningBuilds = readRunningBuilds(pageInfo, predicate);
 
         List<Build> builds = runningBuilds.stream()
-                .skip(pageInfo.getPageIndex() * pageInfo.getPageSize())
+                .skip((long) pageInfo.getPageIndex() * pageInfo.getPageSize())
                 .limit(pageInfo.getPageSize())
                 .collect(Collectors.toList());
         return new Page<>(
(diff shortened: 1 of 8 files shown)


janinko (changes requested):

integration-test/src/test/java/org/jboss/pnc/integration/endpoints/BuildEndpointTest.java
```diff
@@ -110,7 +113,13 @@ public static EnterpriseArchive deploy() {
     @BeforeClass
     public static void prepareData() throws Exception {
         BuildClient bc = new BuildClient(RestClientConfiguration.asAnonymous());
-        Iterator<Build> it = bc.getAll(null, null).iterator();
+        RemoteCollection<Build> builds = bc.getAll(null, null);
+        // After NCL-8156, the default ordering changed to submitTime
+        Iterator<Build> it = builds.getAll()
+                .stream()
+                .sorted(Comparator.comparingLong(build -> new Base32LongID(build.getId()).getLongId()))
```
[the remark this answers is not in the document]
michalovjan: The comment explains why is the custom sort necessary. I can make it more clear.

facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java
```diff
@@ -940,11 +942,19 @@ private Page<Build> getBuilds(
...
+            sortInfo = rsqlPredicateProducer.getSortInfo(type, pageInfo.getSort());
         }
 
-        SortInfo<BuildRecord> sortInfo = rsqlPredicateProducer.getSortInfo(type, pageInfo.getSort());
+        // NCL-8156 SECONDARY sort by unique column like long ID to achieve determinism
+        ToLongFunction<Build> longId = build -> parseId(build.getId()).getLongId();
+        Comparator<Build> sortByLongId = Comparator.comparingLong(longId).reversed(); // id DESC
+
+        comparing = comparing.thenComparing(sortByLongId);
+        sortInfo = DefaultSortInfo.append(sortInfo, BuildRecord_.id, DESC);
```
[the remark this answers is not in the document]
michalovjan: I wanted to avoid adding a new method to the Interface of SortInfo<T> but I guess why not.

integration-test/src/test/java/org/jboss/pnc/integration/endpoints/BuildEndpointTest.java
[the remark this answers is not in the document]
janinko: but comment says that the default order is by submitTime, but code orders by ID

janinko:
Also, although the commit name sounds cool, maybe somethin more concretre/precise would be better?

spi/src/main/java/org/jboss/pnc/spi/datastore/repositories/api/impl/DefaultSortInfo.java
```diff
@@ -22,36 +22,64 @@
...
+
+    private DefaultSortInfo<T> appendAll(OrderInfo.SortingDirection dir, SingularAttribute<T, ?>... fields) {
+        if (fields != null) {
+            for (SingularAttribute<T, ?> attribute : fields) {
+                order.add(new DefaultOrderInfo<>(dir, root -> root.get(attribute)));
+            }
+        }
+
+        return this;
+    }
```
janinko, line 75: public static <T> SortInfo<T> asc(SingularAttribute<T, ?> ... fields) {
        List<DefaultOrderInfo<T>> collect = Arrays.stream(fields)
                .map(f -> new DefaultOrderInfo<T>(OrderInfo.SortingDirection.ASC, (Root<T> root) -> root.get(f)))
                .collect(Collectors.toList());
        return new DefaultSortInfo(collect);
    }
    
    public static <T> SortInfo<T> desc(SingularAttribute<T, ?> ... fields) {
        List<DefaultOrderInfo<T>> collect = Arrays.stream(fields)
                .map(f -> new DefaultOrderInfo<T>(OrderInfo.SortingDirection.DESC, (Root<T> root) -> root.get(f)))
                .collect(Collectors.toList());
        return new DefaultSortInfo(collect);
    }

michalovjan: This is a matter of personal preference. I don't see any issues with the current way how it is written.

janinko: The problem is the mutable protected method appendAll which makes the object mutable.

michalovjan: It's mutable just internally inside the class, there is no accessor to the list that is mutable from the outside and all of the public methods available produce new objects. The class is effectively immutable. 

What is different is that multiple calls to the appendAll inside the class itself wont create additional objects.

janinko (approved):

PR #3843, opened 2023-10-18
Fix deliverable analysis entities SQL script

akridl:


model/src/main/resources/db-update/20060-from-2.5.0-to-2.6.0.sql
```diff
@@ -70,41 +70,41 @@ BEGIN;
...
 
 --------------------------------------------------------------------------------
 -- DeliverableArtifact
 --------------------------------------------------------------------------------
-CREATE TABLE deliverable_artifact
+CREATE TABLE deliverableartifact
     (
         report_id           BIGINT                  NOT NULL,
         artifact_id         INTEGER                 NOT NULL,
         built_from_source   BOOLEAN                 NOT NULL,
```
janinko, line 91: should be `builtfromsource`. Similarly for other columns. (foreign keys are `*_id`)

model/src/main/resources/db-update/20060-from-2.5.0-to-2.6.0.sql
```diff
@@ -70,61 +70,61 @@ BEGIN;
...
-        reason              TEXT,
-        delan_report_label  TEXT,
-        change              TEXT,
+        id                              INTEGER                 NOT NULL,
+        report_id                       BIGINT,
+        changeorder                     INTEGER,
+        entry_time                      TIMESTAMP,
+        user_id                         INTEGER,
+        reason                          TEXT,
+        deliverableanalyzerreportlabel  TEXT,
```
janinko, line 115: I think the field is named only label

janinko (changes requested):

PR #3850, opened 2023-10-26
[NCLSUP-988] fix creation of extra TRs during del-analysis

michalovjan:
### Checklist:

During sustaining, I've found an issue when Orch was creating extra TargetRepositories without ending slash (apparently all should have an ending slash) after DeliverableAnalyzer finishes. 

I am not familiar with the code but I think this should fix it.

Additionally, I think this should be backported to 2.6.X branch.


michalovjan:
retest this please

janinko (approved):

PR #3857, opened 2023-10-27
[NCL-8234] - Fix the math of the DalAnalyzer report; compute the PURL…

vibe13:
… for the generic not found artifacts

### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
@@ -374,7 +374,6 @@ public class DeliverableAnalyzerManagerImpl implements org.jboss.pnc.facade.Deli
      * @param filename the artifact filename
      * @param sha256 the SHA-256 of the artifact
      * @return the generated purl
-     * @throws MalformedPackageURLException
      */
     private String createGenericPurl(String downloadUrl, String filename, String sha256) {
         try {
--- a/facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
@@ -259,7 +259,7 @@ public class DeliverableAnalyzerManagerImpl implements org.jboss.pnc.facade.Deli
 
         // create
         artifact.setTargetRepository(targetRepo);
-        createGenericPurl(targetRepo.getRepositoryPath(), artifact.getFilename().toString(), artifact.getSha256());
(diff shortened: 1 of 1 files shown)


facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
```diff
@@ -259,6 +259,7 @@ private org.jboss.pnc.model.Artifact findOrCreateArtifact(
 
         // create
         artifact.setTargetRepository(targetRepo);
+        createGenericPurl(targetRepo.getRepositoryPath(), artifact.getFilename().toString(), artifact.getSha256());
```
janinko, line 262: Do you just throw away the generated purl?

vibe13: From my understanding of the code, at this point there is no generated PURL for the not found artifacts yet, so I am not throwing anything away?

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
```diff
@@ -364,6 +366,29 @@ private String createPURL(MavenArtifact mavenArtifact) {
...
 
+    /**
+     * Compute the purl string for a generic download, that does not match package type specific files structure. See
+     * https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#generic
+     *
+     * @param downloadUrl url where the artifact was downloaded from
+     * @param filename the artifact filename
+     * @param sha256 the SHA-256 of the artifact
+     * @return the generated purl
+     * @throws MalformedPackageURLException
```
janinko, line 377: it doesn't throw that

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
janinko: The method returns String and has no sideffect (except for the posible exception) and the return value is not used. So this line basically does nothing?

vibe13: doh of course

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
vibe13: good catch

janinko (approved):

vibe13:
retest this please

PR #3870, opened 2023-11-01
[v2.6.x] Remove older roles when migrating to new SSO server

thescouser89:
We don't use them anymore.

### Checklist:

### DIFF
--- a/rest/src/main/java/org/jboss/pnc/rest/SecurityConstraintFilter.java
+++ b/rest/src/main/java/org/jboss/pnc/rest/SecurityConstraintFilter.java
@@ -52,32 +52,17 @@ public class SecurityConstraintFilter implements ContainerRequestFilter {
         String path = requestContext.getUriInfo().getPath();
 
         if ((path.matches("/builds/ssh-credentials.*") || path.matches("/users/current.*"))
-                && Strings.anyStringEquals(method, HttpMethod.GET)
-                && !isUserInAnyRole(UserService.ROLE_USER, UserService.ROLE_ADMIN, UserService.ROLE_SYSTEM_USER)) {
+                && Strings.anyStringEquals(method, HttpMethod.GET) && !userService.isUserLoggedIn()) {
             throw new NotAuthorizedException("Authorization required to access this resource.");
         }
-        if (path.matches("/.*")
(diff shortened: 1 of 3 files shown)


thescouser89:
retest this please

thescouser89:
retest this please

thescouser89:
retest this please

rest/src/main/java/org/jboss/pnc/rest/SecurityConstraintFilter.java
```diff
@@ -52,32 +52,17 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
...
-                        HttpMethod.DELETE,
-                        HttpMethod.HEAD,
-                        HttpMethod.PATCH)
-                && !isUserInAnyRole(UserService.ROLE_USER, UserService.ROLE_ADMIN, UserService.ROLE_SYSTEM_USER)) {
+        if (path.matches("/.*") && Strings.anyStringEquals(
+                method,
+                HttpMethod.POST,
+                HttpMethod.PUT,
+                HttpMethod.DELETE,
+                HttpMethod.HEAD,
```
michalovjan, line 63: Just for curiosity, why do we require user to be authenticated for HEAD request? I see that it has been here since forever but it strikes like a mistake.

michalovjan (changes requested):

rest/src/main/java/org/jboss/pnc/rest/SecurityConstraintFilter.java
thescouser89: No I don't think it should be there, good catch! I can create another PR to fix that.

thescouser89:
Closing, won't be used for this release.

rest/src/main/java/org/jboss/pnc/rest/SecurityConstraintFilter.java
michalovjan: Nice!

thescouser89: https://github.com/project-ncl/pnc/pull/3877

PR #3881, opened 2023-11-15
[NCL-8276] Do not create extra artifacts when running a Deliverable Analysis

vibe13:
@janinko @michalovjan I made some changes to workaround the mocks and improve the artifact detection and reuse (partially introduced in 2.6.0 already) with a further check. Let's discuss what is needed or what needs to be changed. Thanks!

### DIFF
--- a/datastore/src/main/java/org/jboss/pnc/datastore/repositories/ArtifactRepositoryImpl.java
+++ b/datastore/src/main/java/org/jboss/pnc/datastore/repositories/ArtifactRepositoryImpl.java
@@ -21,6 +21,7 @@ import com.google.common.collect.Lists;
 import lombok.extern.slf4j.Slf4j;
 import org.jboss.pnc.datastore.repositories.internal.AbstractRepository;
 import org.jboss.pnc.model.Artifact;
+import org.jboss.pnc.model.Base32LongID;
 import org.jboss.pnc.spi.datastore.predicates.ArtifactPredicates;
 import org.jboss.pnc.spi.datastore.repositories.ArtifactRepository;
 
@@ -67,6 +68,21 @@ public class ArtifactRepositoryImpl extends AbstractRepository<Artifact, Integer
             artifacts.addAll(artifactsInDb);
         }
         return artifacts;
+    }
+
+    @Override
+    public List<Artifact> withSha256AndDependantBuildRecordIdIn(String sha256, Set<Base32LongID> buildRecordIds) {
+        return queryWithPredicates(ArtifactPredicates.withSha256AndDependantBuildRecordIdIn(sha256, buildRecordIds));
+    }
+
+    @Override
+    public List<Artifact> withIdentifierAndSha256(String identifier, String sha256) {
+        return queryWithPredicates(ArtifactPredicates.withIdentifierAndSha256(identifier, sha256));
+    }
+
+    @Override
+    public List<Artifact> withSha256In(Set<String> sha256) {
+        return queryWithPredicates(ArtifactPredicates.withSha256In(sha256));
     }
 
 }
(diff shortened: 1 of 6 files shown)


datastore/src/main/java/org/jboss/pnc/datastore/repositories/ArtifactRepositoryImpl.java
```diff
@@ -69,4 +70,19 @@ public Set<Artifact> withIdentifierAndSha256(Collection<Artifact.IdentifierSha25
...
+
+    @Override
+    public List<Artifact> withIdentifierAndSha256(String identifier, String sha256) {
+        return queryWithPredicates(ArtifactPredicates.withIdentifierAndSha256(identifier, sha256));
+    }
+
+    @Override
+    public List<Artifact> withSha256In(Set<String> sha256) {
+        return queryWithPredicates(ArtifactPredicates.withSha256In(sha256));
+    }
```
janinko, line 86: Why are you adding these as explicit methods, when ther are Predicates?

facade/src/test/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerTest.java
```diff
@@ -117,6 +128,11 @@ public void initMock() {
                     .orElse(null);
         });
         when(artifactRepository.save(any())).thenAnswer(new RepositorSave(artifacts));
+
+        when(artifactRepository.withSha256AndDependantBuildRecordIdIn(anyString(), anySet()))
+                .thenReturn(Collections.emptyList());
```
janinko, line 133: `By default, for all methods that return a value, a mock will return either null, a primitive/primitive wrapper value, or an empty collection, as appropriate. For example 0 for an int/Integer and false for a boolean/Boolean.`

vibe13: ok thanks so these 3 would not be necessary

datastore/src/main/java/org/jboss/pnc/datastore/repositories/ArtifactRepositoryImpl.java
vibe13: Because it might be me not good enough (possible!) with mocks, but I could not find a way to mock this behavior:
when artifactRepository.queryWithPredicates is called with the Predicate "ArtifactPredicates.withIds" then return artifacts, otherwise if  artifactRepository.queryWithPredicates is called with the Predicates "ArtifactPredicates.withSha256AndDependantBuildRecordIdIn", "ArtifactPredicates.withIdentifierAndSha256", "ArtifactPredicates.withSha256In" then return an empty list.... this is for the test below where you found already an improvement.

janinko: Ah ok. It's fine then I think.

What you describe, i think i implemented something like that by something like this:
```
MyObject specificObejct = new MyObject(); // maybe mock(MyObject.class)
mock(ArtifactPredicates.withSha256AndDependantBuildRecordIdIn()).thenReturn(specificObject);
mock(artifactRepository.queryWithPredicates(specificObject)).thenReturn(whatIwant);
```
However I'm not sure from top of my head how/if mocking static methods is done/possible

vibe13: I will look again but I have tried endless ways, above one probably included (too many to remember!) :-)

vibe13: nope.... as soon as I mock more Predicates I get the `org.mockito.exceptions.misusing.InvalidUseOfMatchersException` because the mock cannot distinguish the different predicates (and their different signature). So explicit methods are the only way I am afraid.

michalovjan (approved):

vibe13:
@janinko @michalovjan Are we good to merge this one? And backport the missing pieces in v2.6.x branch?

michalovjan:
My call is yes. But I'd say @janinko has more valid say in this.

janinko (approved):

vibe13:
retest this please

vibe13:
I will merge this one to unblock further work but then will open more PRs to fix other issues we still have.

PR #3885, opened 2023-11-21
[NCL-8273] Update /product-milestones/{id}/delivered-artifacts

janinko:
Update the endpoint to use data in new format that uses Deliverable Analysis Reports to keep backward compatibility.

### Checklist:

### DIFF
--- a/spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
+++ b/spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
@@ -26,11 +26,17 @@ import org.jboss.pnc.model.Artifact_;
 import org.jboss.pnc.model.Base32LongID;
 import org.jboss.pnc.model.BuildRecord;
 import org.jboss.pnc.model.BuildRecord_;
-import org.jboss.pnc.model.ProductMilestone;
+import org.jboss.pnc.model.DeliverableAnalyzerOperation;
+import org.jboss.pnc.model.DeliverableAnalyzerOperation_;
+import org.jboss.pnc.model.DeliverableAnalyzerReport;
+import org.jboss.pnc.model.DeliverableAnalyzerReport_;
+import org.jboss.pnc.model.DeliverableArtifact;
+import org.jboss.pnc.model.DeliverableArtifact_;
 import org.jboss.pnc.model.ProductMilestone_;
 import org.jboss.pnc.spi.datastore.repositories.api.Predicate;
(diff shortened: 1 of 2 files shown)


spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
```diff
@@ -65,8 +70,16 @@ public static Predicate<Artifact> withDependantBuildRecordId(Base32LongID buildR
...
         return (root, query, cb) -> {
-            Join<Artifact, ProductMilestone> productMilestones = root.join(Artifact_.deliveredInProductMilestones);
-            return cb.equal(productMilestones.get(ProductMilestone_.id), productMilestoneId);
+            Root<DeliverableArtifact> deliverableArtifact = query.from(DeliverableArtifact.class);
+            Join<DeliverableAnalyzerReport, DeliverableAnalyzerOperation> operation = deliverableArtifact
+                    .join(DeliverableArtifact_.report)
+                    .join(DeliverableAnalyzerReport_.operation);
+
+            return cb.and(
+                    cb.equal(operation.get(DeliverableAnalyzerOperation_.productMilestone), productMilestoneId),
```
akridl, line 79: Please, add here also `.get(ProductMilestone_.id)` (to make it consistent with the next line)

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
```diff
@@ -65,8 +70,16 @@ public static Predicate<Artifact> withDependantBuildRecordId(Base32LongID buildR
...
+                    cb.equal(
+                            deliverableArtifact.get(DeliverableArtifact_.artifact).get(Artifact_.id),
+                            root.get(Artifact_.id)));
```
akridl, line 82: Why is this condition needed?

janinko: It's to bind together the two tables. This is not actually a join (because I think we can't make a join in this case) but 2 from tables.

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
janinko: fixed

akridl (approved):
Great job Honzaa! Approved

PR #3926, opened 2024-01-09
[NCL-8060]: Remove filtering and sorting at /artifacts/{id}/revisions

akridl:


rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ArtifactEndpoint.java
```diff
@@ -190,6 +190,8 @@ Page<ArtifactInfo> getAllFiltered(
     @Path("/{id}")
     Artifact getSpecific(@Parameter(description = "ID of the Artifact") @PathParam("id") String id);
 
+    static final String GET_SPECIFIC_BY_PURL_DESC = "Gets a specific artifact by purl of the artifact.";
```
janinko, line 193: Use it also

janinko (changes requested):

rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ArtifactEndpoint.java
akridl: Oh shoot, nice catch

janinko (approved):

rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ArtifactEndpoint.java
```diff
@@ -190,14 +190,16 @@ Page<ArtifactInfo> getAllFiltered(
...
+
     /**
      * {@value GET_SPECIFIC_DESC}
```
janinko, line 196: use it also here

PR #3989, opened 2024-03-26
[NCL-8473] Change bifrost log redirect URL

janinko:
### Checklist:

### DIFF
--- a/rest/src/main/java/org/jboss/pnc/rest/endpoints/BuildEndpointImpl.java
+++ b/rest/src/main/java/org/jboss/pnc/rest/endpoints/BuildEndpointImpl.java
@@ -83,14 +83,8 @@ public class BuildEndpointImpl implements BuildEndpoint {
      * Param 1: build-id
      * <p>
      * Param 2: build-log/alignment-log
-     * <p>
-     * Param 3: lines per batch
-     * <p>
-     * Param 4: rate of batches (in ms)
-     * <p>
-     * Param 5: line format (PLAIN/LEVEL) {@see org.jboss.pnc.api.bifrost.enums.Format} for options
      */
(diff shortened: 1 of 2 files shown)


rest/src/main/java/org/jboss/pnc/rest/endpoints/BuildEndpointImpl.java
```diff
@@ -83,14 +83,8 @@ public class BuildEndpointImpl implements BuildEndpoint {
...
      * Param 2: build-log/alignment-log
-     * <p>
-     * Param 3: lines per batch
-     * <p>
-     * Param 4: rate of batches (in ms)
-     * <p>
-     * Param 5: line format (PLAIN/LEVEL) {@see org.jboss.pnc.api.bifrost.enums.Format} for options
      */
-    private static final String BIFROST_LOGS_ENDPOINT = "/text?matchFilters=mdc.processContext:build-{0},loggerName:org.jboss.pnc._userlog_.{1}&batchSize={2}&batchDelay={3}&format={4}&direction=ASC";
+    private static final String BIFROST_LOGS_ENDPOINT = "/final-log/{0}/{1}";
```
michalovjan, line 87: shouldn't it still be `build-{0}`?

michalovjan (changes requested):

rest/src/main/java/org/jboss/pnc/rest/endpoints/BuildEndpointImpl.java
janinko: not from what I used like: `http://bifrost.example.com/final-log/A66P4Z2Q2NQAA/build-log`
cc @thescouser89 is this correct?

thescouser89: Yup correct! This is what I've been using on devel.

I think if all cases processContext = buildId

PR #4031, opened 2024-06-04
[NCL-8718] - Handle unique constraints error when running multiple identical analysis

vibe13:


facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
```diff
@@ -477,7 +477,25 @@ private org.jboss.pnc.model.Artifact findOrCreateNotFoundArtifact(
...
+            return artifacts.iterator().next();
+        }
+        artifacts = artifacts.stream()
+                .filter(art -> art.getIdentifier().equals(artifact.getFilename()))
+                .collect(Collectors.toList());
+        if (artifacts.size() == 1) {
+            return artifacts.iterator().next();
+        }
+
+        // There was no artifact found with the same SHA-56, filename, target repo and identifier. We can create a new
```
janinko, line 497:
```suggestion
        // There was no artifact found with the same SHA-256, filename, target repo and identifier. We can create a new
```

janinko (changes requested):

PR #4032, opened 2024-06-05
[NCL-8365] delete indy promotion records and continually refresh token

michalovjan:
I've made the Indy class into a bean through a producer method and the auth token wasn't being refreshed so I've fixed that. The mechanism is similar to how we do it in RepositoryDriver with custom IndyClientAuthenticator. 

+ some cleanup of unused dependencies and leftover modules

### Checklist:


remote-build-coordinator/src/main/java/org/jboss/pnc/remotecoordinator/maintenance/DefaultRemoteBuildsCleaner.java
```diff
@@ -173,24 +172,35 @@ private Result deleteBuildsFromIndy(BuildRecord buildRecord, String authToken) {
                     }
                 }
             } catch (IndyClientException e) {
-                String description = MessageFormat
-                        .format("Error in deleting generic http repos for build {0}: {1}", buildContentId, e);
```
pkocandr, line 177: This seems intentional here to have also the exception message in the log in case it is displayed without stacktrace, although not sure if this log is read and displayed anywhere nowadays.

pkocandr (commented):
Looks mostly good apart from some minor things mentioned in the comments.

remote-build-coordinator/src/main/java/org/jboss/pnc/remotecoordinator/maintenance/DefaultRemoteBuildsCleaner.java
michalovjan: Slf4j loggers display stack traces if an Exception is last argument in **error** and **warn** levels. The previous way would make it display twice.

remote-build-coordinator/src/main/java/org/jboss/pnc/remotecoordinator/maintenance/TemporaryBuildsCleaner.java
```diff
@@ -125,12 +125,12 @@ private Result deleteTemporaryBuild(BuildRecord buildRecord, String authToken) t
     }
 
     private void deleteArtifact(Artifact artifact) {
-        if (artifact.getDependantBuildRecords().size() > 0) {
-            log.info("Marking temporary artifact as DELETED: " + artifact.getDescriptiveString());
+        if (!artifact.getDependantBuildRecords().isEmpty()) {
```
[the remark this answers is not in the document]
michalovjan: By this you mean to inverse the condition so that there is no **!** there?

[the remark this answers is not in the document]
pkocandr: Yep.

remote-build-coordinator/src/main/java/org/jboss/pnc/remotecoordinator/maintenance/DefaultRemoteBuildsCleaner.java
pkocandr: Yeah, the SLF4J implementation might have changed since this code was written. Let's leave it like this and in case it causes some unclear logging, we can put it back. But it seems it shouldn't.

PR #4040, opened 2024-06-18
When reporting independent builds, count only self-dependence as inde…

janinko:
…pendent

### Checklist:

### DIFF
--- a/integration-test/src/test/java/org/jboss/pnc/integration/endpoints/BuildEndpointTest.java
+++ b/integration-test/src/test/java/org/jboss/pnc/integration/endpoints/BuildEndpointTest.java
@@ -20,7 +20,6 @@ package org.jboss.pnc.integration.endpoints;
 import io.undertow.Undertow;
 import io.undertow.util.HttpString;
 import org.apache.http.entity.ContentType;
-import org.apache.kafka.common.quota.ClientQuotaAlteration;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.container.test.api.RunAsClient;
 import org.jboss.arquillian.junit.Arquillian;
@@ -72,11 +71,13 @@ import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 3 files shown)


janinko:
retest this please

integration-test/src/test/java/org/jboss/pnc/integration/endpoints/BuildEndpointTest.java
```diff
@@ -731,6 +732,21 @@ public void shouldNotApplyUnknownQualityLevel() throws RemoteResourceException {
...
+        BuildClient client = new BuildClient(RestClientConfiguration.asUser());
+
+        Collection<Build> leafTempBuilds = client.getAllIndependentTempBuildsOlderThanTimestamp(Long.MAX_VALUE)
+                .getAll();
+
+        assertThat(leafTempBuilds).hasSize(1);
+        Map<String, String> attributes = leafTempBuilds.iterator().next().getAttributes();
+        assertThat(attributes)
+                .containsEntry(org.jboss.pnc.api.constants.Attributes.BUILD_BREW_NAME, "org.jboss.pnc:parent");
+        assertThat(attributes).containsEntry(org.jboss.pnc.api.constants.Attributes.BUILD_BREW_VERSION, "1.2.4");
```
pkocandr, line 747: Are we sure we have the right one here? org.jboss.pnc:parent:1.2.4 is twice in the demo data - one persistent and one temporary.

janinko: it should return only temporary, but I'll add chceck that it is the temporary one

pkocandr (approved):

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/BuildRecordPredicates.java
```diff
@@ -382,7 +382,10 @@ public static Predicate<BuildRecord> withoutImplicitDependants() {
                             // All entries in a column have to be NULL which is equivalent to count being 0 (aggregation
                             // functions ignore NULLs)
                             cb.equal(cb.count(buildRecordImplicitDependants.get(BuildRecord_.id)), 0)));
-            return cb.and();
+            return cb.or(
+                    // build can depend on itself
+                    cb.notEqual(root.get(BuildRecord_.id), buildRecordImplicitDependants.get(BuildRecord_.id)),
+                    cb.isNull(buildRecordImplicitDependants.get(BuildRecord_.id)));
```
michalovjan, line 388: what was the reason for adding this isNull condition?

This query needs non-null `buildRecordImplicitDependants.get(BuildRecord_.id)` so that the having clause can exclude builds by built-artifacts that have a dependant. This part effectively deletes all built artifacts that are used as a dependency by other build and makes the count always 0 even for builds that have a dependant artifact.

janinko: because criteria api does not have "is distinct from" (nullsafe `!=`). We need to keep the null because otherwise the lines will completly disappear and it will return nothing. Without the isNull it will keep only lines with non-null `buildRecordImplicitDependants.get(BuildRecord_.id)`, so count would be alwas > 0 and nothing would be ever returned.
Also note that it's or.

michalovjan: gotcha

michalovjan (approved):

PR #4044, opened 2024-06-28
[NCL-8681] Store the license information of the deliverable urls analyzed by the Deliverable Analyzer

vibe13:
### Checklist:


facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
```diff
@@ -330,6 +335,25 @@ private void addDeliveredArtifact(
                 .build();
         report.addDeliverableArtifact(deliverableArtifact);
         // distribution.addDeliverableArtifact(deliverableArtifact);
+
+        Set<DeliverableArtifactLicenseInfo> licenses = Optional.ofNullable(licenseInfo)
+                .orElse(Collections.emptySet())
+                .stream()
+                .map(license -> {
+                    return DeliverableArtifactLicenseInfo.builder()
```
janinko, line 343: As in the comment in deliverable analyzer, I think I would prefer a method for the mapping, however here it's short and simple enough that I'm ok also keeping it as is.

janinko (approved):

PR #4048, opened 2024-07-08
NCL-6481 Add validation for long ALIGNMENT_PARAMETERS

patrikk0123:
### Checklist:

### DIFF
--- a/rest/src/main/java/org/jboss/pnc/rest/endpoints/BuildConfigurationEndpointImpl.java
+++ b/rest/src/main/java/org/jboss/pnc/rest/endpoints/BuildConfigurationEndpointImpl.java
@@ -62,10 +62,7 @@ import javax.ws.rs.NotFoundException;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
-import java.util.List;
-import java.util.Map;
-import java.util.OptionalInt;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


rest/src/main/java/org/jboss/pnc/rest/endpoints/BuildConfigurationEndpointImpl.java
```diff
@@ -110,19 +107,38 @@ public Page<BuildConfiguration> getAll(PageParameters pageParams) {
...
+                List<String> categories = Stream.of(BuildCategory.values())
+                        .map(BuildCategory::name)
+                        .collect(Collectors.toList());
+                throw new InvalidEntityException(
+                        buildCategoryKey + " value '" + buildCategoryStr + "' is invalid. Valid values are: "
+                                + String.join(", ", categories) + '.');
+            }
+        }
+
+        final int MAX_ALIGNMENT_PARAMETERS_LENGTH = 8192;
```
janinko, line 133: This should be at least a constant (`private static final` in the class), if we don't want it configurable

janinko (changes requested):

rest/src/main/java/org/jboss/pnc/rest/endpoints/BuildConfigurationEndpointImpl.java
patrikk0123: Changed, btw I can check all the BC params to be of that maximal length, not just alignment param, if that's desirable.

janinko: I don't think we need it for other params.
This one is limited becaues it's pasted to command line, and command line is limited and users may be incentivized sometimes to make it extra long.
Other params probably not limitied, and users are not likly to make them extra long (meybe except for EXTRA_REPOSITORIES)

PR #4065, opened 2024-08-05
[NCL-7053] Validate RSQL query for comparing entity and not field

patrikk0123:
### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/rsql/converter/CastValueConverter.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/rsql/converter/CastValueConverter.java
@@ -24,6 +24,7 @@ import java.time.Instant;
 import java.time.OffsetDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
+import java.util.Arrays;
 import java.util.Date;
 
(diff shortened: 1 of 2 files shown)


facade/src/main/java/org/jboss/pnc/facade/rsql/converter/CastValueConverter.java
```diff
@@ -55,6 +56,8 @@ public <DB extends GenericEntity<?>, T> Comparable<T> convertComparable(Value<DB
                                 + argument,
                         ex);
             }
+        } else if (Arrays.stream(javaType.getDeclaredFields()).anyMatch(f -> f.getName().equals("id"))) {
```
janinko, line 59: Not all entities have ID as field "id", use something like:
```
javaType.getAnnotation(Entity.class) != null;
```

janinko (changes requested):

facade/src/main/java/org/jboss/pnc/facade/rsql/converter/CastValueConverter.java
janinko: Or maybe
```
GenericEntity.class.isAsignableFrom(javaType)
```

janinko (approved):

PR #4103, opened 2024-10-30
[SBOMER-195] Observe build change notification after transactions

janinko:
### Checklist:

### DIFF
--- a/remote-build-coordinator/src/main/java/org/jboss/pnc/remotecoordinator/notifications/buildTask/BuildStatusMQNotifications.java
+++ b/remote-build-coordinator/src/main/java/org/jboss/pnc/remotecoordinator/notifications/buildTask/BuildStatusMQNotifications.java
@@ -28,6 +28,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.enterprise.context.Dependent;
 import javax.enterprise.event.Observes;
+import javax.enterprise.event.TransactionPhase;
 import javax.inject.Inject;
 
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


janinko:
retest this please

remote-build-coordinator/src/main/java/org/jboss/pnc/remotecoordinator/notifications/buildTask/BuildStatusMQNotifications.java
```diff
@@ -49,7 +50,7 @@ public BuildStatusMQNotifications(MessageSenderProvider messageSenderProvider) {
         this.messageSender = messageSenderProvider.getMessageSender();
     }
 
-    public void observeEvent(@Observes BuildStatusChangedEvent event) {
+    public void observeEvent(@Observes(during = TransactionPhase.AFTER_COMPLETION) BuildStatusChangedEvent event) {
```
michalovjan, line 53: Why AFTER_COMPLETION and not AFTER_SUCCESS? If the transaction rolls back, why do we want to sent the UMB message? This part of the code could be (very likely) retried and _may_ succeed which could send a second UMB message.

michalovjan (changes requested):

remote-build-coordinator/src/main/java/org/jboss/pnc/remotecoordinator/notifications/buildTask/BuildStatusMQNotifications.java
janinko: That would mean:
- [this](https://github.com/project-ncl/pnc/blob/529413ceea6fc741647428c2c26e965153bfbe18/remote-build-coordinator/src/main/java/org/jboss/pnc/remotecoordinator/builder/RemoteBuildCoordinator.java#L629) does not neet to be in finally (since on error it will not be observed)
- [this](https://github.com/project-ncl/pnc/blob/529413ceea6fc741647428c2c26e965153bfbe18/remote-build-coordinator/src/main/java/org/jboss/pnc/remotecoordinator/builder/RemoteBuildCoordinator.java#L742) could be removed (since on error it will not be observed)

michalovjan (approved):

PR #4130, opened 2024-11-20
Delete old and duplicate Observer sending UMB message

janinko:
There is the same class in remote build coordinator and both are observing and sending UMB messages.

### Checklist:


thescouser89 (approved):
Makes sense since this is duplicated in remote-build-coordinator

PR #4137, opened 2024-12-09
Add support for Windows artifacts to Deliverables Analyzer

dwalluck:
### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
@@ -28,6 +28,7 @@ import org.jboss.pnc.api.deliverablesanalyzer.dto.BuildSystemType;
 import org.jboss.pnc.api.deliverablesanalyzer.dto.FinderResult;
 import org.jboss.pnc.api.deliverablesanalyzer.dto.LicenseInfo;
 import org.jboss.pnc.api.deliverablesanalyzer.dto.MavenArtifact;
+import org.jboss.pnc.api.deliverablesanalyzer.dto.WindowsArtifact;
 import org.jboss.pnc.api.dto.Request;
 import org.jboss.pnc.api.enums.DeliverableAnalyzerReportLabel;
 import org.jboss.pnc.api.enums.LabelOperation;
@@ -82,7 +83,6 @@ import java.nio.file.Paths;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
@@ -92,6 +92,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
@@ -542,9 +543,8 @@ public class DeliverableAnalyzerManagerImpl implements org.jboss.pnc.facade.Deli
     private org.jboss.pnc.model.Artifact createArtifact(
             org.jboss.pnc.model.Artifact artifact,
             TargetRepository targetRepo) {
-
         artifact.setTargetRepository(targetRepo);
-        artifact.setPurl(createGenericPurl(artifact.getFilename().toString(), artifact.getSha256()));
+        artifact.setPurl(createGenericPurl(artifact.getFilename(), artifact.getSha256()));
         org.jboss.pnc.model.Artifact savedArtifact = artifactRepository.save(artifact);
         targetRepo.getArtifacts().add(savedArtifact);
         return savedArtifact;
@@ -566,17 +566,12 @@ public class DeliverableAnalyzerManagerImpl implements org.jboss.pnc.facade.Deli
             String nvr,
             TargetRepository targetRepository,
             User user) {
-        if (artifact.getArtifactType() != ArtifactType.MAVEN) {
-            throw new UnsupportedOperationException("Brew artifact " + artifact + " is not Maven!");
-        }
-        MavenArtifact mavenArtifact = (MavenArtifact) artifact;
-
-        org.jboss.pnc.model.Artifact.Builder builder = mapArtifact(mavenArtifact, user);
-        builder.identifier(createIdentifier(mavenArtifact));
-        builder.filename(createFileName(mavenArtifact));
-        builder.deployPath(createDeployPath(mavenArtifact));
-        builder.originUrl(createBrewOriginURL(mavenArtifact, nvr));
-        builder.purl(createPURL(mavenArtifact));
+        org.jboss.pnc.model.Artifact.Builder builder = mapArtifact(artifact, user);
+        builder.identifier(createIdentifier(artifact));
+        builder.filename(createFileName(artifact));
+        builder.deployPath(createDeployPath(artifact));
+        builder.originUrl(createBrewOriginURL(artifact, nvr));
+        builder.purl(createPURL(artifact));
         builder.targetRepository(targetRepository);
 
         return builder.build();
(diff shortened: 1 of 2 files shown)


facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
```diff
@@ -603,49 +597,82 @@ private org.jboss.pnc.model.Artifact.Builder mapArtifact(Artifact artifact, User
...
+                if (!StringUtils.isEmpty(mavenArtifact.getClassifier())) {
+                    purlBuilder.withQualifier("classifier", mavenArtifact.getClassifier());
+                }
+
+            } else if (artifact instanceof WindowsArtifact) {
+                WindowsArtifact windowsArtifact = (WindowsArtifact) artifact;
+                purlBuilder = PackageURLBuilder.aPackageURL()
+                        .withType(PackageURL.StandardTypes.GENERIC)
+                        .withName(windowsArtifact.getName())
+                        .withVersion(windowsArtifact.getVersion());
```
dwalluck, line 673: Namespace?

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
```diff
@@ -673,17 +700,25 @@ private String createGenericPurl(String filename, String sha256) {
...
+                            mavenArtifact.getArtifactId(),
+                            mavenArtifact.getType(),
+                            mavenArtifact.getVersion(),
+                            mavenArtifact.getClassifier())
+                    .stream()
+                    .filter(Objects::nonNull)
+                    .collect(Collectors.joining(":"));
+        } else if (artifact instanceof WindowsArtifact) {
+            WindowsArtifact windowsArtifact = (WindowsArtifact) artifact;
+            return String.join("-", windowsArtifact.getName(), windowsArtifact.getVersion());
```
dwalluck, line 718: Not sure.

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
```diff
@@ -603,49 +597,82 @@ private org.jboss.pnc.model.Artifact.Builder mapArtifact(Artifact artifact, User
...
+    private static String createDeployPath(Artifact artifact) {
+        String filename = createFileName(artifact);
+        String deployPath;
+
+        if (artifact instanceof MavenArtifact) {
+            MavenArtifact mavenArtifact = (MavenArtifact) artifact;
+            deployPath = "/" + mavenArtifact.getGroupId().replace('.', '/') + "/" + mavenArtifact.getArtifactId() + "/"
+                    + mavenArtifact.getVersion() + "/" + filename;
+        } else if (artifact instanceof WindowsArtifact) {
+            deployPath = "/" + filename;
```
janinko, line 609: @pkocandrle I am not sure how the deploy path would be if we got that artifact via brew pull (is it even possible?) and if this matches.

dwalluck: As I understand the code, this isn't the Maven repository path, e.g., /maven2. This is the original Brew path (see `createBrewOriginURL()`).

Maven artifacts start with /maven and Windows artifacts start with /win, so I think it just uses the artifact type to start the path. For Windows, since there is no group identifier, etc., the rest of the path is just the filename itself.

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
dwalluck: @pkocandr Should the namespace be empty? Type is `GENERIC` as there is no `WINDOWS` type.

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
dwalluck: @pkocandr For Windows, I just chose `<name>-<version>` as the identifier. I wonder if it's similar for NPM. It obviously will not match the Maven identifier format.

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
pkocandr: We only have support to pull Maven dependecies from Brew, so it would only be possible if a Windows "artiact" was deployed as Maven artifact in a Maven repo probably as a result of a native build. The deploy path would then normally forllow the groupId artifactId coordinates as any other Maven artifact.

dwalluck: But, the whole point of this code was is add support for Windows artifacts in addition to Maven artifacts. So, when you say that we only support pulling Maven dependencies from Brew I am not sure that I understand.

dwalluck: I am under the impression that this change only applies to Deliverables Analyzer and that it shouldn't matter if other parts of PNC support Maven only.

goldmann:
@dwalluck Can you check why this PR fails?

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
pkocandr: Yeah, empty namespace is probably ok.

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
pkocandr: Yes, NPM uses the same format. However not sure if for a windows artifact is `name` sufficient ID of a package. For NPM there is standardized package naming convention, however for windows...? There are multiple types of artifacts, e.g. executables, dll libraries, ... Will the name capture that?

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
pkocandr: Yeah, my comment was answering Honza's question in parenthesis clarifying that Brew pull in PNC does not support other than Maven artifacts. Composing the deploypath like this is probably ok.

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
dwalluck: I'll try to check in Brew.

I've added `filename` and `sha256` to the Windows PURL, so it's the "generic" PURL plus the `filename`.

If `(name, version)` not unique and we need to make it unique, then we could add one or both of those fields.

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
dwalluck: I at least wanted to make sure that the Brew path is still OK, since I moved it into a method which supports both Maven and Windows. This is the HTTP path to the file in Brew.

For Windows, the filename is just `filename`. Since there's already a `filename`, I am not sure why the Maven `filename` has to be (re)composed from the GAVC in the code above, or is it already in `filename`?

dwalluck: OK, name and version of the build obviously isn't unique to the artifact, the filename of which can contain other parts. For example, java-1.8.0-openjdk-1.8.0.432.b06-1.win.jdk.x86_64.zip.

Name java-1.8.0-openjdk
Version	1.8.0.432.b06
Release	1.win

The component "jdk" seems to only appear in the filename.

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
dwalluck: It could just use the "generic" PURL, but I was thinking to set the PRL name to `name` and set the filename as a separate component instead.

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
dwalluck: What do you think of `<name>:<version>:<filename>` for the id? This will at least make it unique now per artifact.

pkocandr: Yeah, this would make it unique. OTOH there is a lot of duplication - both name and version are in the filename. Although it might not be the case everytime, so maybe this is the best we can do.

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
pkocandr: Fileame will probably be necessary. E.g. for build virtio-win-prewhql-0.1-244 the name is virtio-win-prewhql, version 0.1, release 244 and some of the filenames are viowinarm.cab, viowinx64.cab and viowinx86.cab. So the release number might be needed too to make the identification unique.

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
pkocandr: As explained above, the release is not always part of the filename, so it probably needs to be added too for unique identification.

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
pkocandr: Well, there is no field filename in MavenArtifact class, so that's why we need to recompose both filename and deploypath I assume. I think it is ok to have both Maven and Windows artifacts path handled by single method nd I don't see any issue with suggested implementation.
The deploypath is used in combination with targetRepository fields, but all target repositories are hosted in indy. So we either point target repo to Brew or in the artifact we use only originUrl pointing to Brew and use some special empty target repository indicating, that this artifact is not available in Indy. I like the second option more TBH.

PR #4141, opened 2025-01-13
[NCL-8978] Filter out SCRATCH and DELETED DelAn results on Milestone and Version dashboards

patrikk0123:
### Checklist:

### DIFF
--- a/datastore/src/main/java/org/jboss/pnc/datastore/repositories/DeliverableArtifactRepositoryImpl.java
+++ b/datastore/src/main/java/org/jboss/pnc/datastore/repositories/DeliverableArtifactRepositoryImpl.java
@@ -27,6 +27,7 @@ import org.jboss.pnc.model.BuildRecord;
 import org.jboss.pnc.model.BuildRecord_;
 import org.jboss.pnc.model.DeliverableAnalyzerOperation;
 import org.jboss.pnc.model.DeliverableAnalyzerOperation_;
+import org.jboss.pnc.model.DeliverableAnalyzerReport;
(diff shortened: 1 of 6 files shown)


akridl (approved):
LGTM!
I'd welcome some test for this, but IG it's not something so crucial since it's just an additional statistics for users, and not part of any process workflow (and would require adding a few entities in `DatabaseDataInitializer`, hence resulting in several failures in other tests)

patrikk0123:
@akridl can u take a look again please? i reused `notFromScratchAnalysis` logic and added SCRATCH analysis to test data

patrikk0123:
Also added filter for DELETED analysis labels

akridl:
Cool, test is definitely fine!

I was wondering which deliverable analyses will be included in the (future) dashboard:
a) only those labeled `RELEASED`
b) those labeled `RELEASED`, and the ones having no label at all

If it's a), I'd prefer to flip the condition `not SCRATCH && not DELETED` to `RELEASED`. What's more, based on BL of adding/removing delAn report labels (see https://github.com/project-ncl/pnc/blob/ddb2d8f30f0bd016bd9456859d7ea303120b9c78/facade/src/main/java/org/jboss/pnc/facade/util/labels/DeliverableAnalyzerReportLabelModifierImpl.java#L47-L90), if delAn report has `RELEASED` label, then it **has exactly only** `RELEASED` label (hence the whole logic could be simplified).

If it's b), feel free to merge it.

patrikk0123:
@akridl thanks for the idea, but apparently the safest check is `not SCRATCH && not DELETED`

PR #4170, opened 2025-02-06
[NCL-7765] Product Milestone Interconnection Graph REST API

patrikk0123:
### Checklist:

### DIFF
--- a/common/src/main/java/org/jboss/pnc/common/graph/UndirectedGraphBuilder.java
+++ b/common/src/main/java/org/jboss/pnc/common/graph/UndirectedGraphBuilder.java
@@ -1,0 +1,87 @@
+/**
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2014-2022 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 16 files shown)


common/src/main/java/org/jboss/pnc/common/graph/UndirectedGraphBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+        Vertex<T> newVertex = new NameUniqueVertex<>(nodeId.toString(), node);
+        graph.addVertex(newVertex);
+
+        if (depthLimit <= 0) {
+            return newVertex;
+        }
+
+        for (VertexNeighbor<S> vertexNeighbor : getNeighbors(node)) {
+            S neighborId = vertexNeighbor.getNeighborId();
+            boolean doesNeighborExist = getGraphVertex(graph, neighborId) != null;
```
michalovjan, line 65: with the current implementation it doesn't add any edges between already inserted vertices

michalovjan (changes requested):

common/src/main/java/org/jboss/pnc/common/graph/UndirectedGraphBuilder.java
patrikk0123: should be fixed or so i hope

michalovjan (changes requested):

PR #4187, opened 2025-02-24
[NCL-7350] Cleanup unused Brew Push code

janinko:
### Checklist:


dto/src/main/java/org/jboss/pnc/dto/DeliverableAnalyzerReport.java
```diff
@@ -23,7 +23,7 @@
 import lombok.extern.jackson.Jacksonized;
 import org.jboss.pnc.api.enums.DeliverableAnalyzerReportLabel;
 
-import java.util.Date;
+import java.time.Instant;
```
jomrazek, line 26: This is not removal of unused code :)

PR #4193, opened 2025-02-25
[NCK-7765] Product Milestone Interconnections Graph - shared Delivered Artifacts endpoint

patrikk0123:
### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/providers/ArtifactProviderImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/providers/ArtifactProviderImpl.java
@@ -81,6 +81,7 @@ import static org.jboss.pnc.common.util.StreamHelper.nullableStreamOf;
 import static org.jboss.pnc.facade.providers.api.UserRoles.USERS_ADMIN;
 import static org.jboss.pnc.facade.providers.api.UserRoles.USERS_ARTIFACT_ADMIN;
 import static org.jboss.pnc.facade.providers.api.UserRoles.USERS_BUILD_ADMIN;
+import static org.jboss.pnc.spi.datastore.predicates.ArtifactPredicates.deliveredInMilestones;
 import static org.jboss.pnc.spi.datastore.predicates.ArtifactPredicates.withBuildRecordId;
 import static org.jboss.pnc.spi.datastore.predicates.ArtifactPredicates.withDependantBuildRecordId;
 import static org.jboss.pnc.spi.datastore.predicates.ArtifactPredicates.withDeliveredInProductMilestone;
@@ -382,6 +383,25 @@ public class ArtifactProviderImpl
             throw new NotFoundException();
         }
         return mapper.toDTO(artifact);
+    }
+
+    @Override
+    public Page<org.jboss.pnc.dto.Artifact> getDeliveredArtifactsSharedInMilestones(
+            int pageIndex,
+            int pageSize,
+            String sort,
+            String q,
+            String milestone1Id,
+            String milestone2Id) {
+        Integer milestone1IntId = Integer.valueOf(milestone1Id);
+        Integer milestone2IntId = Integer.valueOf(milestone2Id);
+
+        return queryForCollection(
+                pageIndex,
+                pageSize,
+                sort,
+                q,
+                deliveredInMilestones(milestone1IntId, milestone2IntId));
     }
 
     private CriteriaQuery<Tuple> artifactInfoQuery(
--- a/rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ProductMilestoneEndpoint.java
+++ b/rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ProductMilestoneEndpoint.java
@@ -49,6 +49,7 @@ import org.jboss.resteasy.annotations.jaxrs.QueryParam;
 import javax.validation.Valid;
 import javax.validation.constraints.Max;
 import javax.validation.constraints.Min;
+import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotEmpty;
 import javax.validation.constraints.NotNull;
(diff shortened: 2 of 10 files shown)


datastore/src/main/java/org/jboss/pnc/datastore/repositories/ProductMilestoneRepositoryImpl.java
```diff
@@ -195,4 +196,54 @@ public List<Tuple> getMilestonesSharingDeliveredArtifacts(Integer milestoneId) {
...
+    }
+
+    @Override
+    public int countDeliveredArtifactsSharedInMilestones(Integer milestone1Id, Integer milestone2Id) {
+        TypedQuery<Artifact> typedQuery = getDeliveredArtifactsSharedInMilestonesQuery(milestone1Id, milestone2Id);
+
+        return typedQuery.getResultList().size();
+    }
+
+    private TypedQuery<Artifact> getDeliveredArtifactsSharedInMilestonesQuery(
```
michalovjan, line 221: Please, redo this query through queryWithPredicates + predicates. You're returning a regular db model, there's no need for custom queries. You'll get both count() and query with that at the same time; + support for filters/sorting. For reference, joins+group+having with predicates were done f.e. here [1].

[1] https://github.com/project-ncl/pnc/blob/1721880b2fef262c83e3fccd2c285432cd30a2b0/facade/src/main/java/org/jboss/pnc/facade/providers/BuildProviderImpl.java#L247

Custom queries are suited for corner cases when you're returning Tuple or when the query is incompatible with predicates.

datastore/src/main/java/org/jboss/pnc/datastore/repositories/ProductMilestoneRepositoryImpl.java
```diff
@@ -195,4 +196,54 @@ public List<Tuple> getMilestonesSharingDeliveredArtifacts(Integer milestoneId) {
 
         return entityManager.createQuery(query).getResultList();
     }
+
+    @Override
+    public List<Artifact> getDeliveredArtifactsSharedInMilestones(
```
michalovjan, line 201: Query returning Artifacts belongs to Artifact repository.

michalovjan (changes requested):

datastore/src/main/java/org/jboss/pnc/datastore/repositories/ProductMilestoneRepositoryImpl.java
patrikk0123: true

facade/src/main/java/org/jboss/pnc/facade/providers/ArtifactProviderImpl.java
```diff
@@ -384,6 +385,25 @@ public org.jboss.pnc.dto.Artifact getSpecificFromPurl(String purl) {
         return mapper.toDTO(artifact);
     }
 
+    @Override
+    public Page<org.jboss.pnc.dto.Artifact> getDeliveredArtifactsSharedInMilestones(
```
michalovjan, line 389: You can leave this method in PMProvider. On facade layer, you can have whatever model's Repository or even use multiple repositories (see BuildProvider). Just the Repository layer should be compliant by the return type, the facade(Providers) layer is more... contextual??

rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ProductMilestoneEndpoint.java
```diff
@@ -518,4 +520,36 @@ List<DeliveredArtifactInMilestones> compareArtifactVersionsDeliveredInMilestones
...
+                            content = @Content(schema = @Schema(implementation = ErrorResponse.class))),
+                    @ApiResponse(
+                            responseCode = SERVER_ERROR_CODE,
+                            description = SERVER_ERROR_DESCRIPTION,
+                            content = @Content(schema = @Schema(implementation = ErrorResponse.class))) })
+    @GET
+    @Path("/delivered-artifacts/shared")
+    Page<Artifact> getDeliveredArtifactsSharedInMilestones(
+            @Valid @BeanParam PageParameters pageParameters,
+            @Parameter(description = PM_ID) @QueryParam("milestone1") String milestone1Id,
```
michalovjan, line 553: I'd validation with something akin to `@NotBlank` as it's a required parameter right? Same for the second milestone ID.

michalovjan (changes requested):
Looks Good otherwise! You've got some conflicts though.

facade/src/main/java/org/jboss/pnc/facade/providers/ArtifactProviderImpl.java
patrikk0123: this might be problem bcs `queryForCollection` is inherited from AbstractProvider
so in `milestoneProvider`, I would need to add dependency to `artifactProvider`
(milestone's `queryForCollection` wouldn't do for page of Artifacts)

rest-api/src/main/java/org/jboss/pnc/rest/api/endpoints/ProductMilestoneEndpoint.java
patrikk0123: `@NotBlank` looks about right

facade/src/main/java/org/jboss/pnc/facade/providers/ArtifactProviderImpl.java
patrikk0123: in the end i added the artifactProvider dependency, both providers are @Stateless so no circular dependency should happen

michalovjan: no sorry, you've had it correct the first time... I got confused, been a long time since adding things in facade.

PR #4210, opened 2025-03-07
NCL-9004 Update Orch to support PNC builds via Dingrogu

tecarter94:
Initial implementation of Dingrogu build process using PNC build backend

https://issues.redhat.com/browse/NCL-9004


thescouser89 (approved):
LGTM!

PR #4214, opened 2025-03-13
[NCLSUP-1175] Revert temporary build cleanup commit

thescouser89:
Commit: https://github.com/project-ncl/pnc/commit/581505d822483a4d1ec1dc89375bcb53d36c2a5b

We had to previously revert the change in remote builds cleaner to use a
shared Indy object for deletion of temporary builds.

On paper, the shared Indy object makes a lot of sense since we can
re-use the underlying threadpool for http requests. In practice, the
resource in the thread pool wasn't cleaned up properly and stayed in the
pool. This eventually caused all the threads in the threadpool to be
unavailable and cause further deletions to hang.

We workaround it by switching back to creating an Indy object for every
request.

This is all fixed now in Indy client 3.4.1, so this commit removes the
workaround.

### Checklist:

### DIFF
--- a/build-coordinator/src/main/java/org/jboss/pnc/coordinator/maintenance/DefaultRemoteBuildsCleaner.java
+++ b/build-coordinator/src/main/java/org/jboss/pnc/coordinator/maintenance/DefaultRemoteBuildsCleaner.java
@@ -65,7 +65,7 @@ public class DefaultRemoteBuildsCleaner implements RemoteBuildsCleaner {
 
     private Logger logger = LoggerFactory.getLogger(DefaultRemoteBuildsCleaner.class);
 
-    private IndyFactory indyFactory;
+    private final Indy indy;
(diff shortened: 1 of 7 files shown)


janinko:
Just ignore the `OpeanAPI diff` fail, it should be hopefuly fixed for future PRs.

michalovjan (changes requested):
Please do this mainly for DefaultRemoteBuildsCleaner in remote-build-coordinator module. That's the one actually used.

michalovjan (changes requested):
Yo-yo-yo. Reiterating. There are 2 DefaultRemoteBuildsCleaners. Please do this mainly for DefaultRemoteBuildsCleaner in **remote-build-coordinator** module. That's the one actually used.

michalovjan (approved):
Great job!

PR #4232, opened 2025-04-16
[NCL-9083] Fix withDeliveredInProductMilestone predicate not to affect …

patrikk0123:
…original query

### Checklist:

### DIFF
--- a/spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
+++ b/spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
@@ -18,7 +18,6 @@
 package org.jboss.pnc.spi.datastore.predicates;
 
 import org.apache.commons.collections.CollectionUtils;
-import org.jboss.pnc.api.enums.DeliverableAnalyzerReportLabel;
 import org.jboss.pnc.enums.ArtifactQuality;
 import org.jboss.pnc.enums.BuildCategory;
 import org.jboss.pnc.enums.RepositoryType;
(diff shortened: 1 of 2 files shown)


patrikk0123:
added test.
before applying the fix, it fails; now, it succeeds

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
```diff
@@ -79,21 +78,27 @@ public static Predicate<Artifact> withDependantBuildRecordId(Base32LongID buildR
...
+            Join<DeliverableArtifact, Artifact> artifacts = deliverableAnalyzerReports
+                    .join(DeliverableAnalyzerReport_.artifacts)
+                    .join(DeliverableArtifact_.artifact);
+
+            subquery.select(artifacts);
+            subquery.where(
+                    cb.and(
+                            DeliverableAnalyzerReportPredicates.notFromDeletedAnalysis(cb, deliverableAnalyzerReports),
+                            DeliverableAnalyzerReportPredicates.notFromScratchAnalysis(cb, deliverableAnalyzerReports),
+                            cb.equal(productMilestones.get(ProductMilestone_.id), productMilestoneId)));
```
michalovjan, line 98: You can save a join by removing 
```            
Join<DeliverableAnalyzerOperation, ProductMilestone> productMilestones = deliverableAnalyzerOperations
                    .join(DeliverableAnalyzerOperation_.productMilestone);
```
and putting
```
cb.equal(deliverableAnalyzerOperations.get(DeliverableAnalyzerOperation_.productMilestone).get(ProductMilestone_.id), productMilestoneId)));
```

Since the operations table has IDs directly, Hibernate doesn't need to join the PM table.

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
```diff
@@ -79,21 +78,27 @@ public static Predicate<Artifact> withDependantBuildRecordId(Base32LongID buildR
...
+            subquery.groupBy(artifacts);
```
michalovjan, line 99: Do you need `groupBy()` if you're using `root.in()`? Having duplicate date with `in()` shouldn't matter right?

michalovjan (changes requested):

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
patrikk0123: no, it's a left-over; removed

spi/src/main/java/org/jboss/pnc/spi/datastore/predicates/ArtifactPredicates.java
patrikk0123: true

michalovjan (approved):

PR #4251, opened 2025-05-14
Rename brewBuildId -> brewId in delAn RSQL Mapper

akridl:
This change was made to make the name consistent with DTO field, not model field

### Checklist:


facade/src/main/java/org/jboss/pnc/facade/rsql/mapper/DeliverableArtifactRSQLMapper.java
```diff
@@ -51,7 +51,7 @@ public DeliverableArtifactRSQLMapper() {
     @Override
     protected SingularAttribute<? super DeliverableArtifact, ?> toAttribute(String name) {
         switch (name) {
-            case "brewBuildId":
+            case "brewId": // match field name in DTO, not in model
```
janinko, line 54: why is there a comment here?

janinko (changes requested):

janinko (approved):

PR #4287, opened 2025-06-23
Fix build push migration SQL script

janinko:
### Checklist:

### DIFF
--- a/model/src/main/resources/db-update/30200-build-push.sql
+++ b/model/src/main/resources/db-update/30200-build-push.sql
@@ -31,6 +31,12 @@
 --  productmilestonerelease_id | bigint                 |           |          |
 --  userinitiator              | character varying(255) |           |          |
 
+--                    Table "operation_parameters"
+--     Column    |          Type           | Collation | Nullable | Default
+-- --------------+-------------------------+-----------+----------+---------
+--  operation_id | bigint                  |           | not null |
+--  value        | character varying(8192) |           | not null |
+--  key          | character varying(50)   |           | not null |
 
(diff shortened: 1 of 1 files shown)


model/src/main/resources/db-update/30200-build-push.sql
```diff
@@ -147,7 +147,22 @@ BEGIN; -- Data migration
         LEFT JOIN productmilestonerelease mr ON bpr.productmilestonerelease_id = mr.id
         LEFT JOIN productmilestone m ON mr.milestone_id = m.id
         LEFT JOIN product_version_attributes pva ON m.productversion_id = pva.product_version_id AND pva.key = 'BREW_TAG_PREFIX'
-    WHERE (bpr.tagprefix IS NOT NULL AND bpr.tagprefix != '') OR pva.value IS NOT NULL;
+    WHERE ((bpr.tagprefix IS NOT NULL AND bpr.tagprefix != '') OR pva.value IS NOT NULL)
+    AND bpr.id >= 10000000000000000;
+
+    INSERT INTO operation_parameters
+    SELECT
+        (((bpr.id % 100000000) + 529820) << 22)::bigint AS operation_id,
```
akridl, line 155: I believe the same should be done for `buildpushreport` table as well, since there can be an ID >= 10000000000000000, which won't be found in the `buildrecordpushresult` table, because there is added that number -- `(((bpr.id % 100000000) + 529820) << 22)`

akridl (changes requested):

thescouser89:
retest this please

PR #4335, opened 2025-08-11
[NCLSUP-1264] Improve Indy temporary deletion builds

thescouser89:
Before running the deletion, the tracking service verifies it by querying promote service with the trackingID.  If it failed to get the promotion record, in this case 404, it considers the deletion is invalid and fails the deletion. I believe it is part of the new deletion guard check recently pushed to Indy.

We handle this by first taking care of builds with status NO_REBUILD_REQUIRED, which is considered successful, but doesn't produce any artifacts. We can skip the Indy (and Causeway) deletion step there.

We handle missing tracking reports in Indy more gracefully by first checking if they are even present in Indy. Only if present do we attempt to delete the temporary artifacts.

### Checklist:


pkocandr (approved):
Looks good. Thank you!

PR #4501, opened 2026-04-20
NCL-9481 - Adapt brew artifact assertion for delan result

polacektomas:
### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
@@ -911,8 +911,8 @@ public class DeliverableAnalyzerManagerImpl implements org.jboss.pnc.facade.Deli
             this.user = user;
             prefetchPNCArtifacts(builds);
             prefetchTargetRepos(builds);
-            prefetchBrewArtifacts(builds, false);
-            prefetchBrewArtifacts(builds, true);
+            prefetchBrewArtifacts(builds);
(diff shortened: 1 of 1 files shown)


facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
```diff
@@ -919,8 +911,8 @@ public ArtifactCache(Collection<Build> builds, User user) {
             this.user = user;
             prefetchPNCArtifacts(builds);
             prefetchTargetRepos(builds);
-            prefetchBrewArtifacts(builds);
-            prefetchBrewImportedArtifacts(builds);
+            prefetchBrewArtifacts(builds, false);
+            prefetchBrewArtifacts(builds, true);
```
michalovjan: I like the way how it looked before.

What about just creating a method `prefetchBrewImportedArtifacts(builds)` which would just call `prefetchBrewArtifacts(builds,true)`.

It's just personal preference.

michalovjan (changes requested):

facade/src/main/java/org/jboss/pnc/facade/deliverables/DeliverableAnalyzerManagerImpl.java
polacektomas: Sure why not, a bit more readability.

michalovjan:
retest this please

michalovjan (approved):

polacektomas:
retest this please

PR #4502, opened 2026-04-21
[NCL-9557] update error log checksums

michalovjan:
### Checklist:


thescouser89:
At first I was a bit confused but now I get it. If there are error log sent to bifrost, the "final build log" in the attachment will obviously have an updated sha, so we need to update the attachment with the updated sha.

thescouser89 (approved):

PR #4507, opened 2026-04-23
[NCL-9665] Provide values for BUILD_CATEGORY in build config parameters

akridl:
### Checklist:

### DIFF
--- a/facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationSupportedGenericParametersProviderImpl.java
+++ b/facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationSupportedGenericParametersProviderImpl.java
@@ -23,7 +23,6 @@ import org.jboss.pnc.facade.providers.api.BuildConfigurationSupportedGenericPara
 
 import javax.annotation.security.PermitAll;
 import javax.enterprise.context.ApplicationScoped;
-import java.io.IOException;
 import java.util.HashSet;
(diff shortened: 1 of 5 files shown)


facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationSupportedGenericParametersProviderImpl.java
```diff
@@ -18,14 +18,19 @@
 package org.jboss.pnc.facade.providers;
 
 import org.jboss.pnc.api.constants.BuildConfigurationParameterKeys;
+import org.jboss.pnc.api.enums.BuildCategory;
```
akridl: shouldn't we use `BuildCategory` from orch in here?

akridl:
temporarily changing to draft: will be changed to use values from `BuildConfigurationParameterKeys` (which will be introduced soon)

depends on: https://github.com/project-ncl/pnc-api/pull/353

facade/src/main/java/org/jboss/pnc/facade/providers/BuildConfigurationSupportedGenericParametersProviderImpl.java
akridl: it was correct

anyways, the values are now provided as part of `BuildConfigurationParameterKeys`, so no need to do this part

michalovjan (approved):

PR #4525, opened 2026-05-25
[NCL-9487] Add roles to /v2/users/current

thescouser89:
### Checklist:

### DIFF
--- a/rest/src/main/java/org/jboss/pnc/rest/endpoints/UserEndpointImpl.java
+++ b/rest/src/main/java/org/jboss/pnc/rest/endpoints/UserEndpointImpl.java
@@ -53,19 +53,40 @@ public class UserEndpointImpl implements UserEndpoint {
 
     @Override
     public Response loginAndRedirect(String redirectPath) {
+
         // Ensure path starts with / to make it an absolute path
         if (redirectPath == null || redirectPath.trim().isEmpty()) {
-            redirectPath = "/";
-        } else if (!redirectPath.startsWith("/")) {
-            redirectPath = "/" + redirectPath;
+            return redirectToHomePage();
+        } else if (redirectPath.startsWith("/")) {
+            // remove the leading '/' if present
+            redirectPath = redirectPath.substring(1);
         }
 
+        String[] urlToRedirect = redirectPath.split("/", 3);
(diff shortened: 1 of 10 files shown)


thescouser89:
retest this please

rest/src/main/webconfig/auth/web.xml
```diff
@@ -54,17 +54,6 @@
 
     <!-- make sure interceptor sees this and redirect user to sso server -->
     <!-- the rest is handled inside the SecurityConstraintFilter -->
-    <security-constraint>
```
patrikk0123, line 57: why was this here?

rest/src/main/java/org/jboss/pnc/rest/endpoints/UserEndpointImpl.java
```diff
@@ -53,19 +53,40 @@ public User getCurrentUser() {
...
+        }
+
+        String[] urlToRedirect = redirectPath.split("/", 3);
+
+        if (urlToRedirect.length >= 2) {
+            String scheme = urlToRedirect[0];
+            String serverName = urlToRedirect[1];
+            String endPart = urlToRedirect.length == 3 ? urlToRedirect[2] : "/";
+            String absoluteUrl = scheme + "://" + serverName + "/" + endPart;
+            return Response.status(Response.Status.FOUND).location(URI.create(absoluteUrl)).build();
```
patrikk0123, line 72: I meant it should to return 401

thescouser89: This is for the login endpoint, not for the current/users endpoint. If it returns 401, there's no redirect to the SSO server.

rest/src/main/webconfig/auth/web.xml
thescouser89: I initially added it to test if it would return a 302 instead of a 401. But I forgot to remove it afterwards.

thescouser89:
@patrikk0123 Let me know if I resolved the questions you asked! :)

patrikk0123:
@thescouser89 it looks good, but there are probably more things to be added

PR #4562, opened 2026-06-30
[v3.5] Identify cached artifacts by including targetrepository

thescouser89:
This prevents Lightwell type builds from re-using community artifacts already imported in PNC DB, into the built-artifacts section with the wrong target repository.

### Checklist:

* [ ] Have you added unit tests for your change?

PR #4566, opened 2026-07-07
[v3.5.x] Identify cached artifacts by including targetrepository

thescouser89:
This prevents Lightwell type builds from re-using community artifacts already imported in PNC DB, into the built-artifacts section with the wrong target repository.

### Checklist:


thescouser89:
Opening this to target the right branch. Closing #4562

PR #4580, opened 2026-07-22
[NCL-9871] Start classifying conflicting artifacts as FAILED build

akridl:
### Checklist:


akridl:
Converting to a draft, reason: most likely, this won't be needed because once the promotion validation rules (at Indy) are adjusted, it should never happen that the same GAV will be uploaded twice. Implying that an exception from Indy's promotion validation service will be propagated into PNC's repository driver, resulting in a build failure (FAILED build).

Meaning, that this whole PR shouldn't be needed anymore.

akridl:
this solution is not needed anymore

see https://redhat.atlassian.net/browse/NCL-9871 for more details

PR #4589, opened 2026-07-28
[NCL-9843] Explicit Build Category

patrikk0123:
### Checklist:


dto/src/main/java/org/jboss/pnc/dto/Build.java
```diff
@@ -24,6 +24,7 @@
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
 import org.jboss.pnc.api.enums.AlignmentPreference;
+import org.jboss.pnc.api.enums.BuildCategory;
```
patrikk0123: theres conflicting pnc.enums `BuildCategory` which also contains `LEGACY_REDHAT` but one from pnc-api is not deprecated and was upgraded most recently

PR #4596, opened 2026-08-04
feat(NCL-9906): detect artifact conflict if deployed on the same target repository type and path

vibe13:
If an artifact with the same identifier is built and is going to be deployed to different repositories (e.g. type=`MAVEN`, path=`upstream-temporary-builds` and type=`MAVEN`, path=`upstream-builds`) , do not raise a conflict because those are allowed in different destination repositories.

### DIFF
--- a/datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
+++ b/datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
@@ -54,6 +54,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


datastore/src/test/java/org/jboss/pnc/datastore/DatastoreTest.java
```diff
@@ -261,6 +262,37 @@ public void initBuildRecordData() throws Exception {
 
     }
 
+    @Test
+    @InSequence(3)
+    public void shouldOnlyReportBuiltArtifactConflictForSameRepositoryPath() {
```
michalovjan, line 267: this scenario can never happen since a build cannot produce both temporary and persistent artifacts, that'd be an extreme bug

datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
```diff
@@ -111,26 +112,31 @@ public DefaultDatastore(
 
     @Override
     public Map<Artifact, String> checkForBuiltArtifacts(Collection<Artifact> artifacts) {
-        Map<RepositoryType, Map<String, Artifact>> repoTypes = new HashMap<>();
+        Map<RepositoryType, Map<String, List<Artifact>>> repoTypes = new HashMap<>();
```
michalovjan: I am confused a bit. This method's input are **only** `builtArtifacts` (not `dependencies`) of a build record that you're trying to save.

Out of those `buildArtifacts`, `repoTypes` is constructed. RepoTypes Map **key** almost always has **only one** RepositoryType in it because maven builds produce only maven artifacts; npm produce only npm artifacts; etc. Each of these Artifacts has a **unique identifier** and should be in the list **only once**. Therefore, I don't understand the change in repoTypes from `Artifact` to `List<Artifact>`. You should never have 2 Artifacts with the same identifier in the list of `builtArtifacts`. So you should end up with the List of always one Artifact.

I think what you're trying to achieve is to **only** change this condition that determines '_what is a conflict_' on line (130 -> 129)
```
if (conflict.getTargetRepository().getRepositoryType() == repoType) {
```
to something like either (comparing Target repo IDs instead of only RepositoryTypes)
```
if (conflict.getTargetRepository().getId() == repoId) {
```
or (add condition that the repository must be the same temporary quality for it to constitute a conflict)
```
if (conflict.getTargetRepository().getRepositoryType() == repoType && conflict.getTargetRepository().getTemporaryRepo().equals(isTemporaryRepo)) {
```
maybe easier would be to have the Map key be the entire TargetRepository so that the condition is easier to change 
```
Map<TargetRepository, Map<String, List<Artifact>>> repoTypes
```

michalovjan (changes requested):

datastore/src/test/java/org/jboss/pnc/datastore/DatastoreTest.java
vibe13: Ok updated the test

datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
vibe13: You are right, I did an overkill change.

> Comparing Target repo IDs instead of only RepositoryTypes
The repository associated with the incoming artifact has not necessarily been persisted yet, so this might not be reliable

> Add condition that the repository must be the same temporary quality for it to constitute a conflict
It would still report a conflict between two different temporary Maven repositories

So I added the checks that a conflict happen if the same `identifier` is already inside the same repository path.
This means that the same identifier can be both in `lw-upstream` and `lw-temporary-upstream`, but conflicts if they are already in the same repo path.

datastore/src/test/java/org/jboss/pnc/datastore/DatastoreTest.java
```diff
@@ -261,6 +262,40 @@ public void initBuildRecordData() throws Exception {
 
     }
 
+    @Test
+    @InSequence(3)
```
michalovjan: isn't this going to conflict with the other `InSequence(3)` test here? that one persists some data, maybe give it a `(4)` or last `(7)` to be sure this is deterministic

datastore/src/test/java/org/jboss/pnc/datastore/DatastoreTest.java
michalovjan: test looks good!

michalovjan (commented):
Thanks Andrea for fixing this mess! :)

datastore/src/main/java/org/jboss/pnc/datastore/DefaultDatastore.java
michalovjan: 👍! thank you!!

datastore/src/test/java/org/jboss/pnc/datastore/DatastoreTest.java
vibe13: LOL, done :-)

michalovjan (approved):

vibe13:
Thanks @michalovjan !!

PR #4619, opened 2026-08-21
Update to pnc-api 3.5.2

akridl:
### Checklist:


akridl:
thanks to this, `LIGHTWELL_NOVEL` build category is going to be accessible at devel as well

PR #4621, opened 2026-08-21
Re-add retry loop with backoff to generated REST clients

thescouser89:
Replace the single-retry 401 handler with a configurable retry loop that supports both 401 token refresh and retryable exceptions with backoff.

When we switched from PNC 3.4.x to PNC 3.5.x, we removed the Apache Retry Handler from the REST client generation codebase.

This commit attempts to re-add retries to the client, with a default of 5 retries with a retry delay of 500 ms by default.

### Checklist:


thescouser89:
Retry present in 3.4.x removed in 3.5.x: https://github.com/project-ncl/pnc/blob/v3.4.x/rest-api/src/main/java/org/jboss/pnc/client/ApacheHttpClient43EngineWithRetry.java

